PR #161, opened 2023-02-12
d4xx: Enable single DFU device node for each camera.

dmipx:
The naming will be /dev/d4xx-dfu-30-0010 to match media-ctl output.

Signed-off-by: Dmitry Perchanov <dmitry.perchanov@intel.com>


dmipx:
already merged change

PR #165, opened 2023-06-20
firmware log parser added

Tamir91:
Tracked on [LRS-789]


scripts/firmware_log_parser.py
```diff
@@ -0,0 +1,513 @@
+#!/usr/bin/env python3
+"""
+This script parse firmware log
```
Nir-Az, line 3: Please add a short readme

Nir-Az: inside a new folder called `fw_log_parser` next to the parser

Tamir91: May I add the `HWLoggerEventsDS5.xml` file to the folder because I wrote about him in a readme?

Nir-Az: No need, it's from the FW team

dmipx (approved):
Good job!

PR #179, opened 2023-12-10
Update build-jp5.yml

dmipx:
Full build complete

### DIFF
--- a/.github/workflows/build-jp5.yml
+++ b/.github/workflows/build-jp5.yml
@@ -14,8 +14,8 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: setup workspace
-      run: ./setup_workspace.sh 5.0.2
+      run: yes | ./setup_workspace.sh 5.0.2
     - name: apply patches
-      run: ./apply_patches.sh apply 5.0.2
+      run: git config --global user.email "builder@example.com" && git config --global user.name "builder" && ./apply_patches.sh apply 5.0.2
     - name: build
       run: ./build_all.sh 5.0.2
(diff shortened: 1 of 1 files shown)


.github/workflows/build-jp5.yml
```diff
@@ -14,8 +14,8 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: setup workspace
-      run: ./setup_workspace.sh 5.0.2
+      run: yes | ./setup_workspace.sh 5.0.2
```
Nir-Az, line 17: What does `yes` means? :)

.github/workflows/build-jp5.yml
```diff
@@ -14,8 +14,8 @@ jobs:
...
     - name: apply patches
-      run: ./apply_patches.sh apply 5.0.2
+      run: git config --global user.email "builder@example.com" && git config --global user.name "builder" && ./apply_patches.sh apply 5.0.2
     - name: build
       run: ./build_all.sh 5.0.2
```
Nir-Az, line 21: Did you see it fail if the build fail?
I am asking because we had GHA jobs that passed even if the bash return != 0.

Overall it looks great

.github/workflows/build-jp5.yml
Nir-Az: I suggest to run it on any branch.
```
on:
  push:
    branches: ['**']
  pull_request:
    branches: ['**']
```

dmipx: > What does `yes` means? :)

answer "yes" for all questions. In that case it accepts nvidia license

dmipx:
done

Nir-Az (approved):
LGTM

PR #181, opened 2023-12-18
d4xx: dfu release verify comm.

dmipx:
Verify camera communication after dfu procedure, update state firmware version.

Tracked by [LRS-936]


Nir-Az (approved):
LGTM

PR #185, opened 2024-03-05
d4xx: fix dfu release warning

dmipx:
Excessive Warning message of "no communication with d4xx" and App startup delay

https://elixir.bootlin.com/linux/latest/source/drivers/base/regmap/regmap.c#L2850
```
/**
 * regmap_raw_read() - Read raw data from the device
 *
 * @map: Register map to read from
 * @reg: First register to be read from
 * @val: Pointer to store read value
 * @val_len: Size of data to read
 *
 * A value of zero will be returned on success, a negative errno will
 * be returned in error cases.
 */
```
Tracked-by: [RSDSO-19607]

Reported-by: Yusheng Hsu <yusheng.hsu@intel.com>

### DIFF
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -5049,10 +5049,11 @@ static int ds5_dfu_device_release(struct inode *inode, struct file *file)
 #endif
 	/* Verify communication */
 	do {
-		msleep_range(100);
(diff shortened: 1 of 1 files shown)


Nir-Az (approved):
LGTM

hsuys (approved):
Confirm fix for the excessive warning and application delay on AGX Xavier JP5.0.2.

PR #189, opened 2024-03-20
d4xx: hwmc return success on ioctl

dmipx:
LRS expects success on hwmc ioctl call.

[Error] /tmp/sandbox/src/librealsense-exception.h:52 - xioctl(VIDIOC_G_EXT_CTRLS) failed Last Error: Bad message


kernel/realsense/d4xx.c
```diff
@@ -2514,8 +2514,10 @@ static int ds5_g_volatile_ctrl(struct v4l2_ctrl *ctrl)
 			u16 dataLen = 0;
 			u16 bufLen = ctrl->dims[0];
 			ret = ds5_get_hwmc(state, data,	bufLen, &dataLen);
-			if (ret)
+			if (ret) {
```
Nir-Az, line 2517: Let's speak about it,
We should get real return value from FW

Nir-Az (approved):

PR #191, opened 2024-03-31
JP 5.1.2 vi s/g parm callbacks

dmipx:
LRS cannot detect cameras with JP5.1.2 as set/get fps fails with indirect call.

### DIFF
--- a/kernel/nvidia/5.1.2/0003-vi-channel-verify-s-g-callbacks.patch
+++ b/kernel/nvidia/5.1.2/0003-vi-channel-verify-s-g-callbacks.patch
@@ -1,0 +1,44 @@
+From d8e057a16a3a10300633ee9fad32131ee3189cc1 Mon Sep 17 00:00:00 2001
+From: Dmitry Perchanov <dmitry.perchanov@intel.com>
+Date: Sun, 31 Mar 2024 15:05:18 +0200
+Subject: [PATCH] vi: channel: verify s/g callbacks
+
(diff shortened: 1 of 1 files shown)


kernel/nvidia/5.1.2/0003-vi-channel-verify-s-g-callbacks.patch
```diff
@@ -0,0 +1,44 @@
...
++++ b/drivers/media/platform/tegra/camera/vi/channel.c
+@@ -2261,9 +2261,11 @@ __tegra_channel_get_parm(struct tegra_channel *chan,
+ 	struct v4l2_subdev_frame_interval interval;
+ 
+ 	/* dmipx: fixing G_PARM EINVAL error */
+-//	ret = v4l2_subdev_call(sd, video, g_frame_interval, &interval);
+-	ret = sd->ops->video->g_frame_interval(sd, &interval);
+-
++	ret = v4l2_subdev_call(sd, video, g_frame_interval, &interval);
++	if (ret) {
```
Nir-Az, line 23: we want to call `g_frame_interval` when `v4l2_subdev_call` fails? (`ret < 0`)
Don't we want some error log if something is not valid?

dmipx: No, we try again another method. This issue with indirect calls is investigated by v4l community.

Nir-Az (approved):

PR #192, opened 2024-04-01
d4xx: return hwmc error code to lrs sdk.

dmipx:
Align MIPI driver NOT_READY error to UVC flow

Tracked-by: [RSDSO-19622]

### DIFF
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -1776,6 +1776,8 @@ enum DS5_HWMC_ERR {
 	DS5_HWMC_ERR_CMD     = -1,
 	DS5_HWMC_ERR_PARAM   = -6,
 	DS5_HWMC_ERR_NODATA  = -21,
+	DS5_HWMC_ERR_UNKNOWN = -64,
+	DS5_HWMC_ERR_LAST,
 };
 
 static int ds5_get_hwmc_status(struct ds5 *state)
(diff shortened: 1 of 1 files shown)


kernel/realsense/d4xx.c
```diff
@@ -5639,4 +5628,4 @@ MODULE_AUTHOR("Guennadi Liakhovetski <guennadi.liakhovetski@intel.com>,\n\
 				Shikun Ding <shikun.ding@intel.com>");
 MODULE_AUTHOR("Dmitry Perchanov <dmitry.perchanov@intel.com>");
 MODULE_LICENSE("GPL v2");
-MODULE_VERSION("1.0.1.20");
+MODULE_VERSION("1.0.1.21");
```
Nir-Az, line 5631: Please bump to 22 as 5.16 RC is 21
https://github.com/IntelRealSense/realsense_mipi_platform_driver/blob/db6472e0590596a33c37bca9834ca37833aceefb/kernel/realsense/d4xx.c#L5642

Nir-Az:
We must add live tests gating to the driver, hard to tell if this change doesn't break anything else..

kernel/realsense/d4xx.c
Nir-Az: Please remove this change as the version was already updated

dmipx: squash and drop unnecessary commits.

dmipx:
> We must add live tests gating to the driver, hard to tell if this change doesn't break anything else..

Good idea. I will add unit-test

PR #193, opened 2024-04-01
Add  GHA JP5.12 & JP6 builds

Nir-Az:

PR #194, opened 2024-04-03
Jp512 sync nodtc

dmipx:
Fixes: Try adding more GHA builds #193


Nir-Az:
Not needed as was pushed to the original PR
Thanks!

PR #213, opened 2024-07-02
Updated AE_SETPOINT_GET control's range

Arun-Prasad-V:
[Tracked by LRS-1164]

`v4l2-compliance` tool is failing while testing this control as there is no range set.


dmipx (approved):
LGTM
According to 
```
static const struct v4l2_ctrl_config ds5_ctrl_ae_setpoint_set = {
	.ops = &ds5_ctrl_ops,
	.id = DS5_CAMERA_CID_AE_SETPOINT_SET,
	.name = "ae setpoint set",
	.type = V4L2_CTRL_TYPE_INTEGER,
	.min = 0,
	.max = 4095,
	.step = 1,
	.def = 0,
};
```

PR #218, opened 2024-07-19
try to fix the update FW issue using dfu.

mengyui:
try to fix the update FW issue using dfu.
    when update the one's FW, but the updated device is not the given one.


kernel/realsense/d4xx.c
```diff
@@ -5414,6 +5421,7 @@ static int ds5_probe(struct i2c_client *c, const struct i2c_device_id *id)
 #endif
 	if (!state)
 		return -ENOMEM;
+	memset(state, 0, sizeof(*state));
```
Nir-Az, line 5424: Why do we wish to zero this struct here?

mengyui: The memory block has been zero-initialized by devm_kzalloc(). I will remove this line.

Nir-Az:
Merged on PR #232

PR #222, opened 2024-08-03
Add missing git checkout step to readme

Seudonym:
The instructions for JP6 in the dev branch omit a git checkout step.


Nir-Az:
Hi @Seudonym I know it's a bit confusing, but we normally edit the development branch docs and when we release a major release it gets pushed to master branch as a stable validate release.
We ask the users to clone the main branch so it will be relevant when we release it.

Sorry for the confusion but I prefer we keep it as it..

PR #226, opened 2024-08-13
d4xx: Locking on open dfu device

dmipx:
Fix bus clock restore on dfu open/close.

sometimes we have race condition while running tests
```
[109371.797365] d4xx 30-001a: ds5_dfu_device_release(): i2c-2 bus_clk 400000, restore to 100000
[109371.807802] d4xx 30-001a: ds5_dfu_device_open(): i2c-2 bus_clk = 100000, set 400000
[109371.814781] d4xx 30-001a: ds5_dfu_device_release(): i2c-2 bus_clk 400000, restore to 100000
[109371.823742] d4xx 30-001a: ds5_dfu_device_open(): i2c-2 bus_clk = 100000, set 400000
[109371.830182] d4xx 30-001a: ds5_dfu_device_open(): i2c-2 bus_clk = 400000, set 400000
[109371.838771] d4xx 30-001a: ds5_dfu_device_release(): i2c-2 bus_clk 400000, restore to 400000
[109371.847535] d4xx 30-001a: ds5_dfu_device_release(): i2c-2 bus_clk 400000, restore to 400000
[109371.855556] d4xx 30-001a: ds5_dfu_device_open(): i2c-2 bus_clk = 400000, set 400000
[109371.862556] d4xx 30-001a: ds5_dfu_device_release(): i2c-2 bus_clk 400000, restore to 400000
```

### DIFF
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -4967,16 +4967,20 @@ static int ds5_dfu_device_open(struct inode *inode, struct file *file)
 			state->client->adapter);
 #endif
 #endif
-	if (state->dfu_dev.device_open_count)
+	mutex_lock(&state->lock);
+	if (state->dfu_dev.device_open_count) {
+		mutex_unlock(&state->lock);
 		return -EBUSY;
+	}
 	state->dfu_dev.device_open_count++;
 	if (state->dfu_dev.dfu_state_flag != DS5_DFU_RECOVERY)
 		state->dfu_dev.dfu_state_flag = DS5_DFU_OPEN;
 	state->dfu_dev.dfu_msg = devm_kzalloc(&state->client->dev,
 			DFU_BLOCK_SIZE, GFP_KERNEL);
-	if (!state->dfu_dev.dfu_msg)
+	if (!state->dfu_dev.dfu_msg) {
+		mutex_unlock(&state->lock);
 		return -ENOMEM;
-
+	}
 	file->private_data = state;
 #ifdef CONFIG_TEGRA_CAMERA_PLATFORM
 #if LINUX_VERSION_CODE < KERNEL_VERSION(5, 15, 10)
@@ -4984,9 +4988,10 @@ static int ds5_dfu_device_open(struct inode *inode, struct file *file)
 	while (parent && i2c_parent_is_i2c_adapter(parent))
 		parent = i2c_parent_is_i2c_adapter(state->client->adapter);
 
-	if (!parent)
+	if (!parent) {
+		mutex_unlock(&state->lock);
 		return 0;
-
+	}
 	dev_dbg(&state->client->dev, "%s(): i2c-%d bus_clk = %d, set %d\n",
 			__func__,
 			i2c_adapter_id(parent),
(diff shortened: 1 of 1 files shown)


kernel/realsense/d4xx.c
```diff
@@ -4967,26 +4967,29 @@ static int ds5_dfu_device_open(struct inode *inode, struct file *file)
 			state->client->adapter);
 #endif
 #endif
+	mutex_lock(&state->lock);
 	if (state->dfu_dev.device_open_count)
 		return -EBUSY;
```
Nir-Az, line 4972: We need `mutex_unlock(&state->lock);` here right?

kernel/realsense/d4xx.c
```diff
@@ -4997,6 +5000,7 @@ static int ds5_dfu_device_open(struct inode *inode, struct file *file)
 	i2c_set_adapter_bus_clk_rate(parent, DFU_I2C_BUS_CLK_RATE);
 #endif
 #endif
+	mutex_unlock(&state->lock);
```
Nir-Az, line 5003: Who reads `state->dfu_dev.bus_clk_rate`?

We set using the i2c API right?
Why do we need to keep the last value? can't we hard coded set 400 on entrance and set 100 on exit?

Nir-Az:
Nice catch Dima.
Can you write how did you loop enter and exit to DFU mode?
Perhaps this can turn into a unit test for this fix

Nir-Az:
@mengyui On your PR you also added lock and said it caused some crash.
Can you specify on which use case so we can double check here?

Nir-Az:
Another interesting question is how do we see 2 opens w/o a close in between.
How can this happen? Was it teated on a single camera?

mengyui:
The issue which i'm trying to fix is related to "device_open_count"
when try to update FW on the system with multi-9296/D457, the issue may be found.

BTW, should the lock be unlocked before returning from the function?
https://github.com/dmipx/realsense_mipi_platform_driver/blame/busclk-fix/kernel/realsense/d4xx.c#L4972

kernel/realsense/d4xx.c
dmipx: we can hard code values

dmipx:
> Another interesting question is how do we see 2 opens w/o a close in between. How can this happen? Was it teated on a single camera?

It was tested with unit-tests on single camera.

dmipx:
> The issue which i'm trying to fix is related to "device_open_count" when try to update FW on the system with multi-9296/D457, the issue may be found.
> 
> BTW, should the lock be unlocked before returning from the function? https://github.com/dmipx/realsense_mipi_platform_driver/blame/busclk-fix/kernel/realsense/d4xx.c#L4972

Thanks.
Fixed.

dmipx:
> Nice catch Dima. Can you write how did you loop enter and exit to DFU mode? Perhaps this can turn into a unit test for this fix

This is not "enters" dfu mode, it's open dfu device and close dfu device. For Jetson we switch I2C bus speed clock.
The test can be as follows (for jp4/5 only):
1. verify i2c bus speed 100k
2. open dfu device file
3. verify i2c bus speed 400k
4. close dfu device file
5. verify i2c bus speed 100k

kernel/realsense/d4xx.c
dmipx: right, this was another adjustment of where mutex should be hold and we missed that.

Arun-Prasad-V (approved):
LGTM.

PR #242, opened 2024-11-04
Support for dynamic change of i2c bus_clk_rate

Arun-Prasad-V:
Tracked by RSDSO-19802

### DIFF
--- a/kernel/kernel-jammy-src/6.0/0005-support-for-dynamic-change-of-i2c-bus-clk-rate.patch
+++ b/kernel/kernel-jammy-src/6.0/0005-support-for-dynamic-change-of-i2c-bus-clk-rate.patch
@@ -1,0 +1,172 @@
+From ce6faaedef68204e8b132bf44a0e4d2432dfb7ad Mon Sep 17 00:00:00 2001
+From: Arun-Prasad-V <arun.prasad.v@intel.com>
+Date: Mon, 4 Nov 2024 12:52:39 +0530
+Subject: [PATCH] Support for dynamic change of i2c bus-clk-rate
+
+Signed-off-by: Arun Prasad V <arun.prasad.v@intel.com>
+---
+ drivers/i2c/busses/i2c-tegra.c | 42 +++++++++++++++++++++++++++++++
+ drivers/i2c/i2c-core-base.c    | 45 ++++++++++++++++++++++++++++++++++
+ include/linux/i2c.h            |  5 ++++
+ 3 files changed, 92 insertions(+)
+
+diff --git a/drivers/i2c/busses/i2c-tegra.c b/drivers/i2c/busses/i2c-tegra.c
+index cd50436e5734..4269343266c2 100644
+--- a/drivers/i2c/busses/i2c-tegra.c
++++ b/drivers/i2c/busses/i2c-tegra.c
+@@ -1484,6 +1484,37 @@ static int tegra_i2c_xfer_msg(struct tegra_i2c_dev *i2c_dev,
+ 	return 0;
+ }
+ 
++static int tegra_i2c_change_clock_rate(struct tegra_i2c_dev *i2c_dev)
++{
++	u32 val;
++	int err;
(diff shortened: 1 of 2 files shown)


kernel/kernel-jammy-src/6.0/0005-support-for-dynamic-change-of-i2c-bus-clk-rate.patch
```diff
@@ -0,0 +1,172 @@
...
++	val = I2C_CNFG_NEW_MASTER_FSM | I2C_CNFG_PACKET_MODE_EN |
++	      FIELD_PREP(I2C_CNFG_DEBOUNCE_CNT, 2);
++
++	if (i2c_dev->hw->has_multi_master_mode)
++		val |= I2C_CNFG_MULTI_MASTER_MODE;
++
++	i2c_writel(i2c_dev, val, I2C_CNFG);
++
++	if (i2c_dev->prod_list)
++		tegra_i2c_config_prod_settings(i2c_dev);
```
ev-mp, line 35: If the function is not "void" then you need to check return value here and in 37

kernel/kernel-jammy-src/6.0/0005-support-for-dynamic-change-of-i2c-bus-clk-rate.patch
```diff
@@ -0,0 +1,172 @@
...
+ static int tegra_i2c_xfer(struct i2c_adapter *adap, struct i2c_msg msgs[],
+ 			  int num)
+ {
+@@ -1497,6 +1528,16 @@ static int tegra_i2c_xfer(struct i2c_adapter *adap, struct i2c_msg msgs[],
+ 		return ret;
+ 	}
+ 
++	if (adap->bus_clk_rate != i2c_dev->bus_clk_rate) {
++		i2c_dev->bus_clk_rate = adap->bus_clk_rate;
++		 ret = tegra_i2c_change_clock_rate(i2c_dev);
```
ev-mp, line 61: The function does not receive the target I2C frequency, so how do you synchronize selection and activation of the proper I2C clock

Arun-Prasad-V: `adap->bus_clk_rate` will be set by calling `i2c_set_adapter_bus_clk_rate()` API from `d4xx.c` file. Then, in this function, during the next i2c transfer, it will check the clock rate and configure it if changed.

kernel/kernel-jammy-src/6.0/0005-support-for-dynamic-change-of-i2c-bus-clk-rate.patch
Arun-Prasad-V: these are "void" functions only.

kernel/kernel-jammy-src/6.0/0005-support-for-dynamic-change-of-i2c-bus-clk-rate.patch
ev-mp: Is there a limitation to attach the 2nd call to the first one and make the operation atomic? Otherwise there is no guarantee that the transaction will use proper clock setting

Arun-Prasad-V: The functions in this file `i2c-tegra.c` are not exposed to others (there is no corresponding `.h` file which can be imported by others). That's why a new data member `bus_clk_rate` is added to `i2c_adapter` struct (which is exposed in `i2c.h` file) to get the value from user and assign it to `i2c_dev->bus_clk_rate` here in this function.

**Note**: This same logic exist in JP5 source code and didn't get ported to JP6 for some reason. But anyway, it works well for us in JP5.

ev-mp (approved):

PR #254, opened 2025-02-11
changes for jetpack 6.2 compatibility

rickwalder:
Updated and tested on jetpack 6.2. I didn't test on 6.0 since (if I did it right) nothing changed that should touch 6.0

I did NOT update setup_workspace.sh, it would be awesome if someone familiar with that script could contribute an update!

ETA: perhaps more formally, this adds support for Jetson Linux 36.4.3, running 5.15.148-tegra kernel


Nir-Az:
Thanks @rickwalder , much better.
I do see a problematic change you did.
We wish to maintain a single sensor driver, this change should be reverted.
image

rickwalder:
Hey Nir,

There was a change needed in d4xx.c. I suppose the way I made the change this isn't apparent in the repo, but the change is required due to a double definition of this msleep_range function:

```
user@ubuntu:~/realsense_mipi_platform_driver$ diff kernel/realsense/6.0/d4xx.c kernel/realsense/6.2/d4xx.c
498,503c498
< #if LINUX_VERSION_CODE > KERNEL_VERSION(5, 15, 136)
< static inline void msleep_range(unsigned int delay_base)
< {
< 	usleep_range(delay_base * 1000, delay_base * 1000 + 500);
< }
< #endif
```
that definition is removed in the 6.2 version.

At some point between 5.15.136 and 5.15.148 nvidia must have added this definition back elsewhere. 

Instead of how I did it, the check should probably include whatever upper bound that is (it must be <= .148). What we could do for now is check that the version is between .136 and .148, but I'm not sure how this would affect jetpack 6.1.

Thoughts?

Nir-Az:
What you suggested may fix it.
Need to handle different code for different kernel versions if needed.
Still a single D4xx file is a must from our side.

ymodlin:
Hi Rick,
Do you plan to fix this PR?

rickwalder:
Not currently, I’m actually not using it anymore. IIRC it needs someone to find the difference between the kernel versions and change the directives in d4xx appropriately. Otherwise it’s ready to go, I used this change without issue for a few months. On May 15, 2025, at 5:41 AM, ymodlin ***@***.***> wrote:﻿
ymodlin left a comment (IntelRealSense/realsense_mipi_platform_driver#254)
Hi Rick,
Do you plan to fix this PR?

ymodlin:
Before the change what issue did you experience with jetpack 6.2?

rickwalder:
IIRC the patches weren't applied correctly. Things had changed locations in the files to be patched so line numbers needed to be updated

PR #260, opened 2025-04-10
Fix IMU for JP6

Nir-Az:
Continue of #257 - fix by @jzhao80


Nir-Az:
Validated on JP5.0.2 and JP6.0 on FW version 5.16.8.0
Applied by @jzhao80

PR #270, opened 2025-05-29
https instead of git for downloading sources

remibettan:


Nir-Az:
@remibettan can you rebase to see if the GHA now pass?

Nir-Az:
Looks like GHA doesn't like this change, closing for now

PR #300, opened 2025-09-26
RSDSO-20168

ttkhuong:
Add DTB and boot option to enable calibration streaming.


ttkhuong:
Will open PR to dev branch instead

PR #301, opened 2025-09-26
Add DTB and boot option to enable calibration streaming

ttkhuong:


ymodlin:
please fix all copilot comments and reply solved.

hsuys (changes requested):
Suggest to use different name to avoid causing conflict with the regular (non-calibration mode)
ie: overlay-name = "Jetson RealSense Camera D457 Calibration";

ymodlin:
Tri you need to update the version to be 1.0.1.32

ymodlin (approved):

PR #331, opened 2025-12-29
CI: testing frame order and fps

Kontra2B:
Functional V4L2 camera FPS test verifies:

- frame order for repeated and dropped frames (no frame number repetition and 1 drop allowed)
- each frame FPS calculated from time delta falls within KPI
- requeseted frame count arrive


test/groovy/LRS_libci_pipeline.groovy
```diff
@@ -0,0 +1,76 @@
+import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException
```
Nir-Az, line 1: I suggest renaming the file as libci is not related

PR #337, opened 2026-01-21
Fix readme file to remove Intel

sareluzi:


ymodlin (dismissed):
Removing "Intel" is a necessary change thanks for the contribution

PR #347, opened 2026-02-05
Tracked-on RSDSO 21089: Device did not reset after executing the hardware reset script.

ymodlin:


sareluzi:
@ymodlin - please mention which JIRA is it related to

PR #355, opened 2026-02-10
Fix security vulnerabilities across codebase

sareluzi:
## Summary
- Add bounds checking for buffer access in ds5_fw_logger.cpp (HIGH severity)
- Fix buffer overflow in hwmc.cpp - increased buffer size + validation (HIGH severity)
- Replace VLA with heap allocation in StreamView.cpp to prevent stack overflow (HIGH severity)
- Quote shell variables and add validation in build_all.sh (HIGH severity)
- Add malloc null-check in StreamView.cpp (MEDIUM severity)
- Replace unsafe strcpy with strncpy in framesextract.c (MEDIUM severity)
- Add input validation for device paths in test_fw_version.py (MEDIUM severity)
- Add security warning for NOPASSWD sudoers in install.tegra.artifacts.sh (LOW severity)
- Use GitHub actor variable instead of hardcoded git identity in workflows (LOW severity)

## Test plan
- [ ] Compile utilities/streamApp on target platform to verify C++ changes
- [ ] Run `./build_all.sh --clean 6.2` to verify shell script changes
- [ ] Run pytest tests to verify Python validation changes
- [ ] Verify GitHub Actions workflows pass on this PR

### DIFF
--- a/.github/workflows/build-jp502.yml
+++ b/.github/workflows/build-jp502.yml
@@ -20,6 +20,9 @@ jobs:
     - name: setup workspace
       run: yes | ./setup_workspace.sh 5.0.2
     - name: apply patches
-      run: git config --global user.email "builder@example.com" && git config --global user.name "builder" && ./apply_patches.sh 5.0.2
+      run: |
+        git config --global user.email "${{ github.actor }}@users.noreply.github.com"
+        git config --global user.name "${{ github.actor }}"
+        ./apply_patches.sh 5.0.2
     - name: build
       run: ./build_all.sh 5.0.2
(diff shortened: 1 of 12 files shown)


sareluzi:
Thanks for the thorough review! I've addressed all 6 comments in commit 1595c7a:

1. **hwmc.cpp - uninitialized params** - Moved `memset(params, 0, sizeof(params))` before the empty check so params are always initialized even on early return.

2. **StreamView.cpp - memory leak on malloc failure** - Added cleanup loop to free previously allocated buffers before returning -1.

3. **StreamView.cpp - per-frame allocation overhead** - Changed to static buffer with resize-on-demand pattern to avoid allocation on every frame.

4. **test_fw_version.py - multi-entry ls output** - Now using `splitlines()` to parse ls output and pattern matching to find valid DFU device entry.

5. **build_all.sh - unquoted variable** - Quoted all remaining `$TEGRA_KERNEL_OUT` occurrences including the `mkdir -p` calls.

PR #393, opened 2026-03-10
[RSDEV-6494] - Fix rmmod/modprobe after camera disconnect

ejgoldik:
- Removed "ser_prim" from xavier device tree - thus removing the ser address change to 0x42
- Added slight cleanup when probe fails, so next probe attempt will start cleanly


sareluzi (approved):
LGTM

PR #396, opened 2026-03-12
Jp7.x

Kontra2B:
JetPack 7.x build implemented


hardware/nvidia/platform/t19x/galen/kernel-dts/5.1.2/0001-Porting-dts-to-jetpack-5.1.2.patch
```diff
@@ -1,27 +0,0 @@
...
-diff --git a/tegra194-p2888-0001-p2822-0000.dts b/tegra194-p2888-0001-p2822-0000.dts
-index a46befc..7ae66c1 100644
---- a/tegra194-p2888-0001-p2822-0000.dts
-+++ b/tegra194-p2888-0001-p2822-0000.dts
-@@ -18,7 +18,8 @@
-  * Top level DTS file for CVM:P2888-0001 and CVB:P2822-0000.
-  */
- #include "common/tegra194-p2888-0001-p2822-0000-common.dtsi"
--#include "common/tegra194-p2822-camera-modules.dtsi"
-+#include "common/tegra194-camera-d4xx.dtsi"
```
ejgoldik, line 20: Why is this necessary?

hardware/nvidia/platform/t19x/galen/kernel-dts/5.x/0001-Porting-dts-to-jetpack-5.0.2.patch
ejgoldik, line 1: Why is this part of this PR?

kernel/realsense/d4xx.c
```diff
@@ -1442,11 +1442,11 @@ static int ds5_sensor_enum_frame_size(struct v4l2_subdev *sd,
 
 static int ds5_sensor_enum_frame_interval(struct v4l2_subdev *sd,
 #if LINUX_VERSION_CODE < KERNEL_VERSION(5, 15, 10)
-				     struct v4l2_subdev_pad_config *cfg,
```
ejgoldik, line 1566: Please remove the blank space changes (Probably done by editor automation)

kernel/realsense/d4xx.c
```diff
@@ -6099,4 +6176,4 @@ MODULE_AUTHOR("Guennadi Liakhovetski <guennadi.liakhovetski@intel.com>,\n\
 				Shikun Ding <shikun.ding@intel.com>,\n\
 				Dmitry Perchanov <dmitry.perchanov@intel.com>");
 MODULE_LICENSE("GPL v2");
-MODULE_VERSION("1.0.2.17");
+MODULE_VERSION("1.0.2.18");
```
ejgoldik: Align with Uzi regarding the correct version

hardware/nvidia/platform/t19x/galen/kernel-dts/5.x/0001-Porting-dts-to-jetpack-5.0.2.patch
Kontra2B: Reverting to 5.x instead of 5.1.2

kernel/realsense/d4xx.c
Kontra2B: It is my change to align indentation.

setup_workspace.sh
```diff
@@ -52,74 +46,71 @@ if [[ $(uname -m) == aarch64 ]]; then
...
 fi
 
-echo "In a case you have local changes you may reset them with ./apply_patches.sh $1 reset"
-echo
+[[ -d sources_$JETPACK_VERSION ]] && echo -e "\e[33mIn a case you have local changes you may reset them with ./apply_patches.sh reset\e[0m\n"
 # Clone L4T kernel source repo
 cd $DEVDIR
 
-# Check if local tar ball exists in /home/nvidia_sources_cache
+# Check if local tar ball exists in ~/nvidia_sources_cache
```
ejgoldik: This does not work on rslnx as the path is "/home/nvidia_sources_cache" and not "~/nvidia_sources_cache" which translates to "/home/<user name>/nvidia_sources_cache"

Revert this

hardware/nvidia/platform/t19x/galen/kernel-dts/5.1.2/0001-Porting-dts-to-jetpack-5.1.2.patch
Kontra2B: This is not my change. I was here in the patch. I only restored 5.x folder and the file has been moved there. No change here.

setup_workspace.sh
```diff
@@ -52,74 +46,71 @@ if [[ $(uname -m) == aarch64 ]]; then
...
+
     # Check what directory was extracted and rename if necessary
     # The tar ball might contain sources_6.x instead of sources_6.0
     EXTRACTED_DIR=$(tar -tzf "$TARBALL_PATH" | head -1 | cut -d'/' -f1)
-    if [[ "$EXTRACTED_DIR" != "sources_$1" ]]; then
-        echo "Renaming extracted directory from $EXTRACTED_DIR to sources_$1..."
-        mv "$EXTRACTED_DIR" "sources_$1"
+    if [[ "$EXTRACTED_DIR" != "sources_$JETPACK_VERSION" ]]; then
+        echo "Renaming extracted directory from $EXTRACTED_DIR to sources_$JETPACK_VERSION..."
+        mv "$EXTRACTED_DIR" "sources_$JETPACK_VERSION"
```
ejgoldik: Output should remain sources_6.0/sources_5.0.2 etc explicitly and not 6.x 5.x etc. Unless you want to discuss with Yogev and the team about changing it. In any case such a change should not be a part of this PR

setup_workspace.sh
Kontra2B: It is only  a comment. Reverted in next commit.

hardware/realsense/tegra264-camera-d4xx-overlay-advantech.dtso
ejgoldik, line 1: Better change the name to tegra264-camera-d4xx-**overlay**-advantech.dtso to align with the current convention which currently holds for all the other overlays (even if it is redundant). In a later PR we can rename all of them.

ejgoldik (changes requested):

setup_workspace.sh
Kontra2B: I have discussed with Yogev. It is much more convinient if you want to upgrade JP version. Both checking out and building run faster. I am not sure I understood his decision. I will revert when we make a final choice.

hardware/realsense/tegra264-camera-d4xx-overlay-advantech.dtso
Kontra2B: The convetion changes to use dtso extension to align with kernel makefile rule. I will add overlay word with next commit.

README_JP7.md
```diff
@@ -0,0 +1,198 @@
+# RealSense™ camera driver for GMSL* interface
+
+# D457 MIPI on NVIDIA® Jetson AGX Orin™ JetPack 6.x
```
sareluzi: **[BUG]** Title says **"JetPack 6.x"** — should be **"JetPack 7.x"**.

```suggestion
# D457 MIPI on NVIDIA® Jetson AGX Orin™ JetPack 7.x 
```

kernel/realsense/d4xx.c
```diff
@@ -1957,6 +1985,10 @@ static const struct v4l2_subdev_ops ds5_depth_subdev_ops = {
 
 /* FIXME: identical to ds5_depth_pad_ops, use one for both */
 static const struct v4l2_subdev_pad_ops ds5_ir_pad_ops = {
+#if LINUX_VERSION_CODE > KERNEL_VERSION(6, 0, 0)
+	.get_frame_interval	= ds5_sensor_g_frame_interval,
```
sareluzi, line 2069: **[BUG]** `ds5_ir_pad_ops` got `.get_frame_interval` / `.set_frame_interval` for kernel >6.0, but **`ds5_depth_pad_ops`**, **`ds5_rgb_pad_ops`**, and **`ds5_imu_pad_ops`** did not get them. This means depth, RGB, and IMU sensor subdevs won't support frame interval via pad ops on kernel 6.x+, while IR does. Looks like an oversight — all four pad_ops structs should be consistent.

kernel/realsense/d4xx.c
```diff
@@ -4434,15 +4480,18 @@ static int ds5_mux_get_fmt(struct v4l2_subdev *sd,
 
 /* Video ops */
 static int ds5_mux_g_frame_interval(struct v4l2_subdev *sd,
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 10, 0) || LINUX_VERSION_CODE > KERNEL_VERSION(6, 8, 0)
```
sareluzi: **[BUG]** Version guard mismatch: `ds5_mux_g_frame_interval` and `ds5_mux_s_frame_interval` use `KERNEL_VERSION(5, 10, 0) || > KERNEL_VERSION(6, 8, 0)` for the state parameter, but the sensor-level functions (`ds5_sensor_g_frame_interval`/`s_frame_interval`) and the pad_ops both use `KERNEL_VERSION(6, 0, 0)`. These boundaries should be consistent to avoid compilation failures on kernels between 6.0 and 6.8.

kernel/realsense/d4xx.c
```diff
@@ -4327,11 +4359,17 @@ static int ds5_mux_set_fmt(struct v4l2_subdev *sd,
...
 	struct v4l2_subdev *remote_sd;
-	u32 pad = fmt->pad;
+	u32 pad;
 	int ret = 0;
+
+	if (!state) return -EINVAL;
+	if (!fmt) return -EINVAL;
+
 	pad = ds5_state_to_pad(state);
+	pad = fmt->pad;
```
sareluzi: **[BUG]** Dead code: `pad = ds5_state_to_pad(state)` on line above is immediately overwritten by `pad = fmt->pad` here. The `ds5_state_to_pad()` call and its result are unused. Same issue exists in `ds5_mux_get_fmt` below (~line 4437-4438).

kernel/realsense/d4xx.c
```diff
@@ -4378,19 +4416,27 @@ static int ds5_mux_set_fmt(struct v4l2_subdev *sd,
...
+
+	if (!state) return -EINVAL;
+	if (!fmt) return -EINVAL;
+
+	sensor = state->mux.last_set;
+	tmp = *fmt;
 	pad = ds5_state_to_pad(state);
+	pad = fmt->pad;
+
 	sensor = state->mux.last_set;
```
sareluzi: **[BUG]** `sensor = state->mux.last_set` is assigned here, but it was already assigned 5 lines above (~line 4435). This second assignment is redundant.

kernel/realsense/d4xx.c
```diff
@@ -2527,7 +2559,7 @@ static int ds5_s_ctrl(struct v4l2_ctrl *ctrl)
 				memcpy(calib_cmd, &set_calib_data, sizeof(set_calib_data));
 				calib_cmd->header = 276;
 				calib_cmd->param1 = DEPTH_CALIBRATION_ID;
-				memcpy(calib_cmd->Data, (u8 *)ctrl->p_new.p, 256);
+				__memcpy(calib_cmd->Data, (u8 *)ctrl->p_new.p, 256);
```
sareluzi: **[CONCERN]** Change from `memcpy()` to `__memcpy()` bypasses compile-time fortify-source size checking. Was this done to work around a specific compiler warning? If so, the underlying issue should be documented in a comment. `__memcpy` is generally discouraged in kernel code. Same applies to the two other occurrences below (~lines 2591, 2627).

build_all.sh
```diff
@@ -69,52 +73,55 @@ export KERNEL_MODULES_OUT=$TEGRA_KERNEL_OUT/modules
...
         # Remove unwanted
         rm defconfig .config
-        make ARCH=arm64 mrproper
+        make mrproper
         cd $SRCS
         # Building the Image with custom_defconfig
-        make ARCH=arm64 KERNEL_DEF_CONFIG=custom_defconfig -C kernel
+        make KERNEL_DEF_CONFIG=custom_defconfig -C kernel
     else
         # Building the Image with default defconfig
```
sareluzi, line 109: **[STYLE]** This line uses a **tab** for indentation while the rest of the file uses **spaces** (compare with the `make KERNEL_DEF_CONFIG=...` line just above). The JP7 dtb copy block below (~lines 117-122) also uses mixed tabs. Please use consistent spacing throughout.

apply_patches.sh
```diff
@@ -29,138 +23,163 @@ elif [[ "$1" == "--fg12-16ch-dual" ]]; then
...
 }
 
-apply_external_patches "$ACTION" "$1" "$D4XX_SRC_DST" "$L4T_VERSION"
+if [[ ! -d "sources_${JETPACK_VERSION}" ]]; then
+    echo "Sources folder not found. Run ./setup_workspace.sh first"
+    exit 2
+fi
 
-[[ -d "sources_$SOURCES_VERSION/$KERNEL_DIR" ]] && apply_external_patches "$ACTION" "$1" "$KERNEL_DIR" "$L4T_VERSION"
+apply_external_patches "$version" "$D4XX_SRC_DST"
```
sareluzi: **[BUG/RISK]** `apply_external_patches "$version" "$D4XX_SRC_DST"` passes `$version` (e.g., `7.1` or `4.6.1`) as the first arg, which becomes the patch subdirectory name inside the function (`ls -Ld ${PWD}/$2/$1`). For JP4, `JETPACK_VERSION` becomes `4.x` while `$version` stays `4.6.1`. The function will look for patch dirs like `nvidia-oot/4.6.1/` or `kernel/nvidia/4.6.1/`. The directory rename in this PR changed `4.6.1/` to `4.x/` for some paths. Please verify that patch sub-directories consistently match the `$version` value for **all** JetPack versions.

sareluzi (commented):
## Code Review — JetPack 7.x Support (PR #396)

Overall a solid effort to unify the build infrastructure and add JP7 support. Found several issues ranging from bugs to style concerns. See inline comments plus these additional notes:

### `apply_patches.sh` — Removed `.realsense_patch_base` tracking
The previous code stored the original commit hash before applying patches in `.realsense_patch_base` and used it for reset. Now reset always does `git reset --hard $L4T_VERSION`. If the source repo was on a different commit than the tag, the reset behavior changes. Likely fine, but worth noting as a behavioral change.

### `build_all.sh` — `make install -C kernel` was removed for JP6+
The old code had `make ARCH=arm64 install -C kernel` which copies the kernel Image to the output. That line is now gone. Verify that the kernel Image is still being installed to `$TEGRA_KERNEL_OUT/rootfs/boot/` for JP6. For JP7, `make -C kernel` builds it, but the `install` target also copies it to the output directory.

scripts/setup-common
```diff
@@ -1,40 +1,81 @@
...
+	fi
+fi
+
+export JETPACK_VERSION="$version"
 export REVISION="35.1"
 export L4T_VERSION=jetson_"${REVISION}"
 export KERNEL_DIR="kernel/kernel-5.10"
 export LICENSE=release
+export ARCH=arm64
+export version="$version"
```
sareluzi: **[STYLE]** `export version` — `version` is a very generic variable name for an exported env var. It could clash with other scripts or tools in the environment. Consider a more specific name (e.g., `JETPACK_INPUT_VERSION`) or keep it unexported since `setup-common` is sourced.

scripts/setup-common
```diff
@@ -1,40 +1,81 @@
...
+if [[ $1 == "-h" ]]; then
+    echo "$0 JetPack_version"
+    echo "$0 -h"
+    echo "JetPack_version can be $supported"
+    exit 1
+fi
+
+version="$1"
+if [[ -z "$version" ]]; then
+	version="${supported%%,*}"
```
sareluzi: **[MINOR]** When no argument is passed and `jetpack_version` file doesn't exist, `version` silently defaults to `${supported%%,*}` = `7.1`. This changes previous behavior which required an explicit version argument. Users with older Jetson boards running `./build_all.sh` without arguments will silently target 7.1 instead of getting an error.

test/test_metadata/test_metadata.c
```diff
@@ -116,6 +116,7 @@ static int streamFrames(int videoNode_fd, int md_fd, void* metaDataBuffers[SIZE_
         mdV4l2Buffer.type = V4L2_BUF_TYPE_META_CAPTURE;
         mdV4l2Buffer.memory = V4L2_MEMORY_MMAP;
 
+		fprintf(stderr, "Dqueue buffers...\n");
```
sareluzi: **[STYLE]** These `fprintf(stderr, ...)` debug statements (here plus ~8 more throughout the file) add noise to stderr during normal operation. They also use **tab** indentation while surrounding code uses **spaces**. Consider removing them before merge, or guarding behind a `--verbose` flag.

kernel/realsense/d4xx.c
```diff
@@ -6079,6 +6155,7 @@ MODULE_DEVICE_TABLE(of, d4xx_of_match);
 static struct i2c_driver ds5_i2c_driver = {
 	.driver = {
 		.owner = THIS_MODULE,
+		.of_match_table = of_match_ptr(d4xx_of_match),
```
sareluzi, line 7034: **[MINOR]** Adding `.of_match_table` here — verify this doesn't conflict with how the match table was previously associated with the driver (some older kernel setups may have used `MODULE_DEVICE_TABLE` alone or a separate mechanism). Also confirm `d4xx_of_match` is defined and visible at this point for all kernel version paths.

kernel/realsense/d4xx.c
Kontra2B: This is old file. In the new one the spaces are removed.

kernel/realsense/d4xx.c
Kontra2B: @sareluzi ?

sareluzi: Let's keep the version as is and @Nikolai-L will bump it when he merges his code

kernel/realsense/d4xx.c
sareluzi: Still unresolved after latest push. The mux functions use `KERNEL_VERSION(5, 10, 0) ; > KERNEL_VERSION(6, 8, 0)` while sensor-level functions and `ds5_pad_ops` use `KERNEL_VERSION(6, 0, 0)`. These need to be aligned.

kernel/realsense/d4xx.c
sareluzi: Still present. `pad = ds5_state_to_pad(state)` is immediately overwritten by `pad = fmt->pad`. Same in `ds5_mux_get_fmt`.

kernel/realsense/d4xx.c
sareluzi: Still present. `sensor = state->mux.last_set` appears twice  at the top of the function and again a few lines later.

kernel/realsense/d4xx.c
sareluzi: Still present. Three occurrences of `__memcpy` remain (~lines 2518, 2547, 2583). Please use `memcpy`  the double-underscore variant bypasses compiler instrumentation (KASAN, KCSAN).

apply_patches.sh
sareluzi: Still relevant. The patch subdirectory is selected via `` (e.g. '7.1'), but the actual directories under `kernel/kernel-noble-src/` are named '7.0'. If someone creates a release like '7.0.1' or if the version string diverges from the directory name, patches silently won't be found.

scripts/setup-common
sareluzi: Still present. Exporting a variable named just `version` pollutes the environment and risks collisions with other scripts or tools that source setup-common. Consider renaming to `JETPACK_VERSION` or `JP_VERSION`.

scripts/setup-common
sareluzi: Still present. When no argument is given and no `jetpack_version` file exists, the script silently defaults to the first entry in the supported list (currently '7.1'). This could be surprising  a warning message would help.

test/test_metadata/test_metadata.c
sareluzi: Still present. The `fprintf(stderr, ...)` debug prints look like leftover development aids. Consider removing them or gating behind a verbose/debug flag.

build_all.sh
sareluzi: Partially fixed  the `make install` was restored, thanks. However the JP7 dtb-copy block (lines ~85-91 in the new file) still uses mixed tabs and spaces for indentation.

apply_patches.sh
```diff
@@ -29,138 +29,168 @@ elif [[ "$1" == "--fg12-16ch-dual" ]]; then
...
+	echo Usage:
+	echo $0 [reset]
 fi
 
-ACTION="$2"
-[[ -z "$ACTION" ]] && ACTION="apply"
+[[ $# -gt 0 ]] && [[ $1 != "-h" ]] && echo -e too many arguments && exit 1
+. scripts/setup-common
+
+if [[ "$2" == reset ]]; then
```
sareluzi: **Dead code.** By this point all positional arguments have been consumed by shift operations above (lines 7-8, 26, 29, 34), and line 42 exits if any extra args remain. So `dollar-2` is always empty here. The reset handling already happens at lines 5-8 and 32-35. This block can be removed.

kernel/kernel-noble-src/7.0/0006-fix-null-pointer-exception.patch
```diff
@@ -0,0 +1,13 @@
...
++++ b/drivers/media/v4l2-core/v4l2-subdev.c
+@@ -1568,7 +1568,7 @@ __v4l2_subdev_state_get_format(struct v4l2_subdev_state *state,
+ 		if (stream)
+ 			return NULL;
+ 
+-		if (pad >= state->sd->entity.num_pads)
++		if (state->sd && pad >= state->sd->entity.num_pads)
+ 			return NULL;
+ 
+ 		return &state->pads[pad].format;
```
sareluzi, line 13: This patches the kernel framework (v4l2-subdev.c) to add a NULL guard on `state->sd`. While it prevents the crash, it masks the actual bug - `state->sd` should never be NULL when __v4l2_subdev_state_get_format is called. The root cause is likely that D4XX creates/uses a v4l2_subdev_state without properly associating it with a subdev.

Patching the kernel framework is fragile (breaks on kernel upgrades) and hides the real issue. Consider investigating *why* `state->sd` is NULL and fixing that in the D4XX driver instead.

sareluzi (commented):
## Follow-up Review (v2)

### Previously resolved (thank you):
- README_JP7.md title: **Fixed**
- Consolidated pad_ops: **Fixed**
- `make install` restored in build_all.sh: **Fixed**

### Still unresolved (replies posted on original threads):
- d4xx.c: Version guard mismatch
- d4xx.c: Dead code pad assignment
- d4xx.c: Duplicate sensor assignment
- d4xx.c: __memcpy should be memcpy
- apply_patches.sh: version used as patch subdir name
- setup-common: export version (generic name)
- setup-common: Silent default to first supported version
- test_metadata.c: Debug fprintf calls
- build_all.sh: Mixed tabs/spaces in JP7 dtb block

### New findings in this round:
See inline comments below.

apply_patches.sh
Kontra2B: V.x folders are for sources only. Patches for kernel and nvidia-oot use version. For hardware V.x is used (line 109)

scripts/setup-common
Kontra2B: There will be no V.x folder and an error triggered.

Example:JetPack version 4.6.1...

Native build

./build_all.sh: line 81: cd: /home/wojtek/realsense_mipi_platform_driver/sources_4.x/kernel/kernel-4.9: No such file or directory

kernel/realsense/d4xx.c
Kontra2B: Before the module was configured to be loaded in /etc/. This change enables device tree to load it. No previous association. It is visible since all kernel builds pass.

kernel/realsense/d4xx.c
sareluzi: @Kontra2B - bump to 24

apply_patches.sh
```diff
@@ -2,10 +2,10 @@
...
 
-if [[ $# < 1 ]]; then
-    echo "apply_patches.sh [--one-cam | --dual-cam] JetPack_version [apply]"
-    echo "apply_patches.sh JetPack_version reset"
-    exit 1
+ACTION="apply"
+if [[ "$1" == reset ]]; then
+	ACTION="reset"
+	shift
 fi
```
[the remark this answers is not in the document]
Kontra2B: This is on purpose to handle arbitrary argument order.

kernel/realsense/d4xx.c
Kontra2B: It is 24 already

kernel/realsense/d4xx.c
```diff
@@ -3203,7 +3195,7 @@ static int ds5_s_ctrl(struct v4l2_ctrl *ctrl)
 				memcpy(calib_cmd, &set_calib_data, sizeof(set_calib_data));
 				calib_cmd->header = 276;
 				calib_cmd->param1 = DEPTH_CALIBRATION_ID;
-				memcpy(calib_cmd->Data, (u8 *)ctrl->p_new.p, 256);
+				__memcpy(calib_cmd->Data, (u8 *)ctrl->p_new.p, 256);
```
sareluzi: Still present from previous rounds. Three occurrences of `__memcpy` (lines 3198, 3227, 3263). The double-underscore variant bypasses compiler sanitizer instrumentation (KASAN, KCSAN). Please use `memcpy`.

kernel/realsense/d4xx.c
```diff
@@ -5186,15 +5189,18 @@ static int ds5_mux_get_fmt(struct v4l2_subdev *sd,
 
 /* Video ops */
 static int ds5_mux_g_frame_interval(struct v4l2_subdev *sd,
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 10, 0) || LINUX_VERSION_CODE > KERNEL_VERSION(6, 8, 0)
```
sareluzi: **Bug (High): Version guard mismatch — will break compilation on kernel 6.8.0.**

The mux frame interval functions use:
```c
#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 10, 0) || LINUX_VERSION_CODE > KERNEL_VERSION(6, 8, 0)
```

But `ds5_pad_ops` and `ds5_sensor_video_ops` use:
```c
#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 8, 0)
```

Two problems:
1. `> KERNEL_VERSION(6, 8, 0)` excludes 6.8.0 itself, while `>= KERNEL_VERSION(6, 8, 0)` includes it. On kernel 6.8.0, pad_ops will reference the state parameter but the function signature won't have it → compilation error.
2. The `< KERNEL_VERSION(5, 10, 0)` branch is wrong — `struct v4l2_subdev_state` didn't exist before 5.15, so this would add a parameter of a type that doesn't exist on kernel 4.9.

All guards should consistently use `>= KERNEL_VERSION(6, 8, 0)`.

kernel/realsense/d4xx.c
```diff
@@ -1616,8 +1617,17 @@ static int ds5_sensor_get_fmt(struct v4l2_subdev *sd,
 	if (fmt->which == V4L2_SUBDEV_FORMAT_TRY)
 #if LINUX_VERSION_CODE < KERNEL_VERSION(5, 15, 10)
 		fmt->format = *v4l2_subdev_get_try_format(sd, cfg, fmt->pad);
-#else
+#elif LINUX_VERSION_CODE < KERNEL_VERSION(6, 8, 0)
 		fmt->format = *v4l2_subdev_get_try_format(sd, v4l2_state, fmt->pad);
+#else
+	{
+		struct v4l2_mbus_framefmt * framefmt;
```
sareluzi: Kernel coding style: `struct v4l2_mbus_framefmt *framefmt` (no space before `*`). Same issue on line 1743.

apply_patches.sh
```diff
@@ -29,138 +29,166 @@ elif [[ "$1" == "--fg12-16ch-dual" ]]; then
...
-elif [[ -d "sources_$JETPACK_VERSION" ]]; then
-    SOURCES_VERSION="$JETPACK_VERSION"
-else
-    SOURCES_VERSION="$1"  # Default to original input if neither exists yet
+if [[ $1 == "-h" ]]; then
+	echo Usage:
+    echo "$0 [--one-cam | --dual-cam | --max96712-EVB | --fg12-16ch | --fg12-16ch-dual ] [reset] [-h]"
+	echo -e 'reset\t: hard reset (git) to version from jetpack_version file'
+	echo -e '-h\t: show this help'
 fi
```
sareluzi: **Bug: `-h` missing `exit 1`.** This prints the help text but falls through to `setup-common` which prints *its* help and exits. The user sees two help messages — confusing.

Same issue in `setup_workspace.sh` line 38.

build_all.sh
```diff
@@ -34,7 +28,14 @@ NPROC=$(nproc)
 
 . $DEVDIR/scripts/setup-common "$1"
 
-SRCS="$DEVDIR/sources_$1"
+if [[ "$1" == "-h" ]]; then
```
sareluzi, line 29: **Dead code: this `-h` handler can never execute.** Line 29 sources `setup-common "$1"`, which sees `$1 == "-h"` and calls `exit 1` before reaching this block.

Either move this handler before the `setup-common` source, or pass `-h` handling entirely to `setup-common`.

build_all.sh
```diff
@@ -69,53 +73,56 @@ export KERNEL_MODULES_OUT=$TEGRA_KERNEL_OUT/modules
...
-    make ARCH=arm64 dtbs
     mkdir -p $TEGRA_KERNEL_OUT/rootfs/boot/dtb
-    cp $SRCS/nvidia-oot/device-tree/platform/generic-dts/dtbs/tegra234-p3737-0000+p3701-0000-nv.dtb $TEGRA_KERNEL_OUT/rootfs/boot/dtb/
-    cp $SRCS/nvidia-oot/device-tree/platform/generic-dts/dtbs/tegra234-p3737-0000+p3701-0005-nv.dtb $TEGRA_KERNEL_OUT/rootfs/boot/dtb/
-    cp $SRCS/nvidia-oot/device-tree/platform/generic-dts/dtbs/tegra234-camera-d4xx-overlay*.dtbo $TEGRA_KERNEL_OUT/rootfs/boot/
+    if version_lt "$JETPACK_VERSION" "7.0"; then
+        make dtbs
+        cp $SRCS/nvidia-oot/device-tree/platform/generic-dts/dtbs/tegra234-camera-d4xx-overlay*.dtbo $TEGRA_KERNEL_OUT/rootfs/boot/
+    else
+        cp $SRCS/$KERNEL_DIR/arch/arm64/boot/dts/nvidia/tegra2[36]4-camera-d4xx-overlay*.dtbo $TEGRA_KERNEL_OUT/rootfs/boot/
```
sareluzi: **Potential issue: JP7 DTBOs not built before copy.**

For `version_lt < 7.0` (JP6), `make dtbs` is called explicitly. For JP7, DTBOs are copied from `$KERNEL_DIR/arch/arm64/boot/dts/nvidia/` without a preceding `make dtbs`. If `make -C kernel` (line ~108) doesn't implicitly build device tree overlays for JP7, this `cp` will fail with "no such file".

Please verify that `make -C kernel` builds DTBOs for the JP7 kernel tree, or add an explicit `make dtbs` for this path too.

kernel/kernel-noble-src/7.0/0006-fix-null-pointer-exception.patch
sareluzi, line 13: This patches the kernel framework (`v4l2-subdev.c`) to add a NULL guard on `state->sd`. While it prevents the crash, it masks the real bug — `state->sd` should never be NULL when `__v4l2_subdev_state_get_format` is called.

Patching the kernel framework is fragile (breaks on kernel upgrades) and hides the root cause. Consider investigating *why* `state->sd` is NULL in the D4XX driver path and fixing it there instead.

sareluzi (commented):
## Review Round 3 (after force-push to 89de0e1b)

Major rewrite of shell scripts — great improvement in structure and consistency with `version_lt`, `JP_INPUT_VERSION`, unified sources dir naming. Many of the issues from the previous review rounds have been addressed.

### Fixed since last review (thank you):
- README_JP7.md title
- Consolidated pad_ops into single `ds5_pad_ops` / `ds5_subdev_ops`
- `make install` restored in build_all.sh
- `export version` renamed to `JP_INPUT_VERSION`
- Duplicate `sensor = state->mux.last_set` removed
- Dead `$2 == reset` block removed (rewritten)
- `$version` as patch subdir replaced by `JP_INPUT_VERSION`

### Still present + New findings:
See inline comments below.

apply_patches.sh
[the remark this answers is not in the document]
Kontra2B: Updated argument handling

kernel/kernel-noble-src/7.0/0006-fix-null-pointer-exception.patch
Kontra2B: We talked about this today on daily. This is an ad hoc solution that works. I will address the root cause trying to bring advantech board.

sareluzi: how is it related to Advantech board?

apply_patches.sh
Kontra2B: Yes. Intended. setup-common prints supported JP version that are common for each script invoked. Here is example output:
Usage:
./apply_patches.sh [--one-cam | --dual-cam | --max96712-EVB | --fg12-16ch | --fg12-16ch-dual ] [reset] [-h]
reset   : hard reset (git) to version from jetpack_version file
-h      : show this help
supported JetPack versions: 7.1, 7.0, 6.2.1, 6.2, 6.1, 6.0, 5.1.2, 5.0.2, 4.6.1
But here I can add exit as the script does not take JP version as param.

kernel/realsense/d4xx.c
Kontra2B: does not build with memcpy

kernel/kernel-noble-src/7.0/0006-fix-null-pointer-exception.patch
Kontra2B: It is related to any overlay IMO. Advantech overlay is not working so when I assume to rework it I will address the issue.

sareluzi: if not Advantech board related- let's address it now and not add another patch if we can avoid it

build_all.sh
Kontra2B: make kernel does build overlays

kernel/kernel-noble-src/7.0/0006-fix-null-pointer-exception.patch
Kontra2B: Too complicated. Should go under a new task. Requires redesign of d4xx.c: removing of media cotroller muxer as I explained yesterday on daily. Definitely not to be included in this task IMO.

kernel/realsense/d4xx.c
sareluzi: Thanks for the explanation. What build error do you get with `memcpy`? This is likely a `CONFIG_FORTIFY_SOURCE` size check failure because the destination (`calib_cmd->Data`) is a flexible array member or incomplete type. If so, the proper fix is to use `unsafe_memcpy()` (available since 5.18) or add `__HAVE_ARCH_MEMCPY`-guarded wrappers — not `__memcpy`, which also skips KASAN instrumentation. Could you share the error message?

sareluzi: Thanks for the explanation. What build error do you get with memcpy? This is likely a CONFIG_FORTIFY_SOURCE size check failure because the destination (calib_cmd->Data) is a flexible array member or has incomplete size info. If so, the proper fix is to use unsafe_memcpy() (available since kernel 5.18) or annotate the struct member with __counted_by -- not __memcpy, which also skips KASAN instrumentation. Could you share the error message?

kernel/realsense/d4xx.c
sareluzi: Partially fixed - thank you for changing > to >=, that resolves the 6.8.0 gap. However the other branch < KERNEL_VERSION(5, 10, 0) is still problematic: it adds a struct v4l2_subdev_state parameter on kernels where that type does not exist (introduced in 5.14). On kernel 4.9 this would fail to compile. Since this driver targets 4.9+ via existing ifdefs, the mux frame interval guards should match the sensor ones: just use >= KERNEL_VERSION(6, 8, 0).

build_all.sh
sareluzi: Partially fixed - the -h check is now before setup-common (line 29), so it can execute. However it still has no exit -- after printing the help, it falls through to setup-common which will print its own help and exit. Add exit 0 after the echo lines (like apply_patches.sh does).

sareluzi: Partially fixed - the -h check is now before setup-common (line 29), so it can execute. However it still has no exit -- after printing the help text it falls through to setup-common which prints its own help and exits. Add exit 0 after the echo lines (like apply_patches.sh now does).

ejgoldik (dismissed):

scripts/setup-common
```diff
@@ -1,40 +1,81 @@
...
+export version="$version"
+
+function version_lt {
+    IFS='.' read -r -a v1 <<< "$1"
+    IFS='.' read -r -a v2 <<< "$2"
+    for i in 0 1 2; do
+        [[ v1[i] -lt v2[i] ]] && return 0
+        [[ v1[i] -gt v2[i] ]] && return 1
+    done
+    return 1
```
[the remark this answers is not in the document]
Kontra2B: Works as expected.

build_all.sh
```diff
@@ -69,52 +73,56 @@ export KERNEL_MODULES_OUT=$TEGRA_KERNEL_OUT/modules
...
+    make kernel_name=$UBUNTU_CODENAME modules
     mkdir -p $TEGRA_KERNEL_OUT/rootfs/boot/dtb
-    cp $SRCS/nvidia-oot/device-tree/platform/generic-dts/dtbs/tegra234-p3737-0000+p3701-0000-nv.dtb $TEGRA_KERNEL_OUT/rootfs/boot/dtb/
-    cp $SRCS/nvidia-oot/device-tree/platform/generic-dts/dtbs/tegra234-p3737-0000+p3701-0005-nv.dtb $TEGRA_KERNEL_OUT/rootfs/boot/dtb/
-    cp $SRCS/nvidia-oot/device-tree/platform/generic-dts/dtbs/tegra234-camera-d4xx-overlay*.dtbo $TEGRA_KERNEL_OUT/rootfs/boot/
+    if version_lt "$JETPACK_VERSION" "7.0"; then
+        make dtbs
+        cp $SRCS/nvidia-oot/device-tree/platform/generic-dts/dtbs/tegra234-camera-d4xx-overlay*.dtbo $TEGRA_KERNEL_OUT/rootfs/boot/
+    else
+        cp $SRCS/$KERNEL_DIR/arch/arm64/boot/dts/nvidia/tegra234-camera-d4xx-overlay*.dtbo $TEGRA_KERNEL_OUT/rootfs/boot/ 2>/dev/null || true
```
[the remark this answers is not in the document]
Kontra2B: Done

kernel/realsense/d4xx.c
Kontra2B: CC [M]  /home/wojtek/realsense_mipi_platform_driver/sources_7.1/nvidia-oot/drivers/media/i2c/d4xx.o
In file included from ./include/linux/string.h:292,
                 from ./arch/arm64/include/asm/processor.h:32,
                 from ./include/linux/sched.h:13,
                 from ./include/linux/delay.h:23,
                 from /home/wojtek/realsense_mipi_platform_driver/sources_7.1/nvidia-oot/drivers/media/i2c/d4xx.c:20:
In function ‘fortify_memcpy_chk’,
    inlined from ‘ds5_s_ctrl’ at /home/wojtek/realsense_mipi_platform_driver/sources_7.1/nvidia-oot/drivers/media/i2c/d4xx.c:3199:5:
./include/linux/fortify-string.h:542:25: error: call to ‘__write_overflow_field’ declared with attribute warning: detected write beyond size of field (1st parameter); maybe use struct_group()? [-Werror=attribute-warning]
  542 |                         __write_overflow_field(p_size_field, size);
      |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
make[6]: *** [scripts/Makefile.build:243: /home/wojtek/realsense_mipi_platform_driver/sources_7.1/nvidia-oot/drivers/media/i2c/d4xx.o] Error 1
make[5]: *** [scripts/Makefile.build:481: /home/wojtek/realsense_mipi_platform_driver/sources_7.1/nvidia-oot/drivers/media/i2c] Error 2
make[4]: *** [scripts/Makefile.build:481: /home/wojtek/realsense_mipi_platform_driver/sources_7.1/nvidia-oot/drivers/media] Error 2
make[4]: *** Waiting for unfinished jobs....
make[3]: *** [scripts/Makefile.build:481: /home/wojtek/realsense_mipi_platform_driver/sources_7.1/nvidia-oot/drivers] Error 2
make[2]: *** [/home/wojtek/realsense_mipi_platform_driver/sources_7.1/kernel/kernel-noble-src/Makefile:1925: /home/wojtek/realsense_mipi_platform_driver/sources_7.1/nvidia-oot] Error 2

Kontra2B: I do not think we will ever use KASAN. It is known to work best with clang.

apply_patches.sh
```diff
@@ -2,165 +2,182 @@
...
-            rm -f "sources_$SOURCES_VERSION/$3/.realsense_patch_base"
-        else
-            git -C "sources_$SOURCES_VERSION/$3" reset --hard $4
+        echo -e "\e[33m$(ls -Ld ${PWD}/$2/$1)\e[0m"
+        ls -Lw1 "${PWD}/$2/$1"
+        git -C "${source}" apply "${PWD}/$2/$1"/*
+    elif [[ "$ACTION" = "reset" ]]; then
+        if ! git -C "${source}" diff --quiet || ! git -C "${source}" diff --cached --quiet; then
+            read -p "Repo ${source} has changes that will be hard reset. Continue (Y/n)? " confirm
+            [[ -n "$confirm" && "$confirm" != "y" && "$confirm" != "Y" ]] && exit 1
```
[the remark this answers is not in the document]
Kontra2B: No need. To get here one explicitly issues ./apply_patches.sh w/wo reset. The confirmation with Y is annoying for developer. Still one needs to confirm hitting enter.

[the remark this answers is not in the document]
Kontra2B: No need. To get here explicit ./apply_patches.sh reset must be issued. Confirmation with Y is annoying while developing. Still one needs to hit enter to proceed.

apply_patches.sh
```diff
@@ -2,165 +2,182 @@
...
+            git config user.email "$GIT_AUTHOR_EMAIL"
+    fi
+
     # Update local git identity for subrepos
-    git -C "sources_$SOURCES_VERSION/$D4XX_SRC_DST" config user.name "$GIT_AUTHOR_NAME"
-    git -C "sources_$SOURCES_VERSION/$D4XX_SRC_DST" config user.email "$GIT_AUTHOR_EMAIL"
-    if [[ -d "sources_$SOURCES_VERSION/$KERNEL_DIR" ]]; then
-        git -C "sources_$SOURCES_VERSION/$KERNEL_DIR" config user.name "$GIT_AUTHOR_NAME"
-        git -C "sources_$SOURCES_VERSION/$KERNEL_DIR" config user.email "$GIT_AUTHOR_EMAIL"
+    git -C "${SRCS}/$D4XX_SRC_DST" config user.name "$GIT_AUTHOR_NAME"
```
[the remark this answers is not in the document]
Kontra2B: Good point. Not my change. Will address with another task.

build_all.sh
Kontra2B: It is on purpose to list available JP versions that are hardcoded only in setup-common. Example:
wojtek@advantech ~/realsense_mipi_platform_driver (JP7.x) [SIGINT]> ./setup_workspace.sh -h
Usage:
./setup_workspace.sh [JetPack]
supported JetPack versions: 7.1, 7.0, 6.2.1, 6.2, 6.1, 6.0, 5.1.2, 5.0.2, 4.6.1
wojtek@advantech ~/realsense_mipi_platform_driver (JP7.x)>

kernel/realsense/d4xx.c
sareluzi: Thanks for the log. As expected, this is a `CONFIG_FORTIFY_SOURCE` issue with the zero-length array `Data[0]` in `struct hwm_cmd` (line 255):

```c
struct hwm_cmd {
    ...
    unsigned char Data[0];  // <- old-style GNU zero-length array
};
```

FORTIFY_SOURCE can't determine the size of a `[0]` field, so it flags `memcpy` into it as a potential overflow.

**The proper fix is to change `Data[0]` to `Data[]`** (C99 flexible array member):

```c
unsigned char Data[];  // flexible array member - FORTIFY_SOURCE understands this
```

This tells the compiler the field is variable-length, and `memcpy` will work without warnings. No need for `__memcpy` at all.

Flexible array members have been the standard kernel practice since the `[0]` → `[]` migration (see Gustavo A. R. Silva's tree-wide cleanup, commit `c47a6ceb`). The `[0]` form is technically a GNU extension and is flagged by `-Wgnu-zero-length-array` in newer compilers.

kernel/realsense/d4xx.c
Kontra2B: done

kernel/realsense/d4xx.c
Kontra2B: done

kernel/realsense/d4xx.c
sareluzi: Thanks for fixing `g_frame_interval` (line 5192). However `s_frame_interval` at line 5229 still has the same stale guard:
```c
#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 10, 0) || LINUX_VERSION_CODE >= KERNEL_VERSION(6, 8, 0)
```
Please apply the same fix there: change it to just `>= KERNEL_VERSION(6, 8, 0)`.

sareluzi: Thanks for fixing `g_frame_interval` (line 5192). However `s_frame_interval` at line 5229 still has the same stale guard:
```c
#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 10, 0) || LINUX_VERSION_CODE >= KERNEL_VERSION(6, 8, 0)
```
Please apply the same fix there: change it to just `>= KERNEL_VERSION(6, 8, 0)`.

kernel/realsense/d4xx.c
sareluzi: Verified in d49a716: `Data[0]` changed to `Data[]` and all three `__memcpy` reverted to `memcpy`. Looks good, resolving.

sareluzi: Verified in d49a716: `Data[0]` changed to `Data[]` and all three `__memcpy` reverted to `memcpy`. Looks good, resolving.

build_all.sh
sareluzi: Fair enough — the intentional fall-through to `setup-common` shows the supported JP version list, which is useful. Resolving.

sareluzi (approved):

ejgoldik (approved):

PR #403, opened 2026-03-25
Restore max9296 GPIO tunneling for external sync

ejgoldik:
Restore GPIO tunneling for external sync that was mistakenly removed by PR 398


ipilchin1 (approved):
Good job, thanks

PR #404, opened 2026-03-25
Max96712 multicamera

ejgoldik:
This enables up to four cameras connected at once to the same max96712 deserializer.
Note that as max96712 has 8 video pipes, the user can stream up to 8 streams at once from all cameras connected to the same deserializer (ir left + right count as 1)

### DIFF
--- a/hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-1-2-3.dts
+++ b/hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-1-2-3.dts
@@ -20,7 +20,7 @@
 
 /* camera control gpio definitions */
 / {
-	overlay-name = "Jetson dual RealSense Camera D4xx with a single max96712 and tca9546 (fg12-16ch)";
+	overlay-name = "Jetson four RealSense Cameras D4xx with a single max96712 and tca9546 (fg12-16ch)";
 	jetson-header-name = "Jetson AGX CSI Connector";
 	compatible = JETSON_COMPATIBLE;
 
@@ -41,7 +41,7 @@
 		target-path = "/";
 		__overlay__ {
 			tegra-capture-vi {
-				num-channels = <8>;
+				num-channels = <16>;
 				ports {
 					status = "okay";
 					port@0 {
@@ -132,13 +132,101 @@
 							remote-endpoint = <&d4xx_csi_out7>;
 						};
 					};
+					port@8 {
+						status = "okay";
+						reg = <8>;
+						d4xx_vi_in8: endpoint {
+							status = "okay";
+							vc-id = <8>;
+							port-index = <0>;
+							bus-width = <2>;
+							remote-endpoint = <&d4xx_csi_out8>;
+						};
+					};
+					port@9 {
+						status = "okay";
+						reg = <9>;
+						d4xx_vi_in9: endpoint {
+							status = "okay";
+							vc-id = <9>;
+							port-index = <0>;
+							bus-width = <2>;
+							remote-endpoint = <&d4xx_csi_out9>;
+						};
+					};
+					port@10 {
+						status = "okay";
+						reg = <10>;
+						d4xx_vi_in10: endpoint {
+							status = "okay";
+							vc-id = <10>;
+							port-index = <0>;
+							bus-width = <2>;
+							remote-endpoint = <&d4xx_csi_out10>;
+						};
+					};
+					port@11 {
+						status = "okay";
+						reg = <11>;
+						d4xx_vi_in11: endpoint {
+							status = "okay";
+							vc-id = <11>;
+							port-index = <0>;
+							bus-width = <2>;
+							remote-endpoint = <&d4xx_csi_out11>;
+						};
+					};
+					port@12 {
+						status = "okay";
+						reg = <12>;
+						d4xx_vi_in12: endpoint {
+							status = "okay";
+							vc-id = <12>;
+							port-index = <0>;
+							bus-width = <2>;
+							remote-endpoint = <&d4xx_csi_out12>;
+						};
+					};
+					port@13 {
+						status = "okay";
+						reg = <13>;
+						d4xx_vi_in13: endpoint {
+							status = "okay";
+							vc-id = <13>;
+							port-index = <0>;
+							bus-width = <2>;
+							remote-endpoint = <&d4xx_csi_out13>;
+						};
+					};
+					port@14 {
+						status = "okay";
+						reg = <14>;
+						d4xx_vi_in14: endpoint {
+							status = "okay";
+							vc-id = <14>;
+							port-index = <0>;
+							bus-width = <2>;
+							remote-endpoint = <&d4xx_csi_out14>;
+						};
+					};
+					port@15 {
+						status = "okay";
+						reg = <15>;
+						d4xx_vi_in15: endpoint {
+							status = "okay";
+							vc-id = <15>;
+							port-index = <0>;
+							bus-width = <2>;
+							remote-endpoint = <&d4xx_csi_out15>;
+						};
+					};
 				};
 			};
 
 			bus@0 {
 				host1x@13e00000 {
 					nvcsi@15a00000 {
-						num-channels = <8>;
+						num-channels = <16>;
 						channel@0 {
 							status = "okay";
 							reg = <0>;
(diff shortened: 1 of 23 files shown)


kernel/realsense/d4xx.c
```diff
@@ -63,7 +66,7 @@ struct dser_interface {
 	int (*power_on)(struct device *dev);
 	void (*power_off)(struct device *dev);
```
sareluzi, line 67: Double semicolon at end of line:
```c
int ser_pipe_id = state->dser_ops->get_ser_pipe_id(state->dser_dev, pipe_id, ser_vc_id);;
```
Harmless but should be cleaned up.

sareluzi (commented):
## PR Review: Max96712 multicamera

This PR adds support for 4 cameras on a single MAX96712 deserializer. Key changes:
- New DT overlay for 4-camera config
- Dynamic pipe allocation in MAX96712 (replaces hardcoded vc_id == pipe_id)
- New `bind_ser_to_dser_pipe` / `get_ser_pipe_id` APIs
- `external_addr_reassign` DT property for MAX9295
- MAX96712 reset via register when GPIO not available
- `MAX_DS5_NUM` increased from `MAX_DSER_NUM * 2` to `* 4`
- `-ENOSR` early-return to avoid pointless retries

14 comments below.

kernel/realsense/d4xx.c
```diff
@@ -1813,12 +1820,17 @@ static int ds5_setup_pipeline(struct ds5 *state, u8 data_type1, u8 data_type2,
 	int ret = 0;
 	/* While some deserializers can support up to 8 pipes, the serializer only supports
 	 * four pipes and four vc_ids (0 - 3).
-	 * In this case, a second camera connected to a deserializer, will have its pipes 0 - 3
-	 * with vc_id 0 - 3 connected to the deserializer's pipes 4 - 7 and vc_ids 4 - 7
+	 * To use multiple cameras under this restriction, a second camera connected
+	 * to a deserializer will have its vc_id 0 - 3 mapped to outside vc_id 4 - 7 etc.
+	 * The ser_pipe to dser_pipe mapping depends on the deserializer.
 	 */
```
sareluzi, line 1826: **1/14 - Double semicolon**

```c
int ser_pipe_id = state->dser_ops->get_ser_pipe_id(state->dser_dev, pipe_id, ser_vc_id);;
```

Extra semicolon at end of line.

kernel/realsense/d4xx.c
```diff
@@ -1813,12 +1820,17 @@ static int ds5_setup_pipeline(struct ds5 *state, u8 data_type1, u8 data_type2,
...
-	int ser_pipe_id = pipe_id % DS5_MAX_STREAMS;
 	int ser_vc_id = vc_id % DS5_MAX_STREAMS;
+	int ser_pipe_id = state->dser_ops->get_ser_pipe_id(state->dser_dev, pipe_id, ser_vc_id);;
```
sareluzi: **3/14 - Missing NULL check for `get_ser_pipe_id`**

`get_ser_pipe_id` is called unconditionally (line 1826), but `bind_ser_to_dser_pipe` gets a NULL check (line 1828). Both are new interface additions.

Since both are populated for max9296 and max96712, the NULL check on `bind_ser_to_dser_pipe` is inconsistent — either remove it (both are always set) or add one for `get_ser_pipe_id` too.

kernel/realsense/d4xx.c
```diff
@@ -566,7 +569,7 @@ struct dser_control {
 };
 static struct dser_control dser_inited[MAX_DSER_NUM];
 
-#define MAX_DS5_NUM (MAX_DSER_NUM * 2) /* assuming max 2 DS5 cameras per deserializer */
+#define MAX_DS5_NUM (MAX_DSER_NUM * 4) /* assuming max 2 DS5 cameras per deserializer */
```
sareluzi: **4/14 - Comment is now wrong**

```c
#define MAX_DS5_NUM (MAX_DSER_NUM * 4) /* assuming max 2 DS5 cameras per deserializer */
```

The comment says "max 2" but the multiplier is now 4. Should be "max 4 DS5 cameras per deserializer".

kernel/realsense/d4xx.c
```diff
@@ -5183,6 +5195,10 @@ static int ds5_mux_s_stream(struct v4l2_subdev *sd, int on)
 		if (!ds5_config_done) {
 			ret = ds5_configure(state);
 			if (ret < 0) {
+				if (ret == -ENOSR) {
```
sareluzi, line 5195: **14/14 - `-ENOSR` check will never trigger**

```c
if (ret == -ENOSR) {
```

Good idea, but `max96712_get_available_pipe_id` returns `-ENOMEM` (not `-ENOSR`) when all pipes are allocated. This early exit will never fire for the MAX96712 path. Either change `max96712_get_available_pipe_id` to return `-ENOSR`, or check for `-ENOMEM` here too.

nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch
```diff
@@ -322,20 +329,51 @@ index 8e237eef..33646ace 100644
...
++	unsigned int pipe_en_val = 0;
++
++	if (pipe_id < 0 || pipe_id >= MAX96712_MAX_PIPES)
++		return -EINVAL;
++
++	mutex_lock(&priv->lock);
++	priv->video_pipe_alloc[pipe_id] = false;
++
++	max96712_read_reg(priv, MAX96712_VIDEO_PIPE_EN_ADDR, &pipe_en_val);
++	max96712_write_reg(priv, MAX96712_VIDEO_PIPE_EN_ADDR, pipe_en_val &= ~(1 << pipe_id));
```
sareluzi: **5/14 - Compound assignment inside function call**

```c
max96712_write_reg(priv, MAX96712_VIDEO_PIPE_EN_ADDR, pipe_en_val &= ~(1 << pipe_id));
```

Same pattern in `max96712_set_pipe` (line ~431 with `|=`). Using `&=` / `|=` as a function argument is confusing. Split into two statements:
```c
pipe_en_val &= ~(1 << pipe_id);
max96712_write_reg(priv, MAX96712_VIDEO_PIPE_EN_ADDR, pipe_en_val);
```

nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch
```diff
@@ -322,20 +329,51 @@ index 8e237eef..33646ace 100644
...
++	struct max96712 *priv = dev_get_drvdata(dev);
++	unsigned int pipe_en_val = 0;
++
++	if (pipe_id < 0 || pipe_id >= MAX96712_MAX_PIPES)
++		return -EINVAL;
++
++	mutex_lock(&priv->lock);
++	priv->video_pipe_alloc[pipe_id] = false;
++
++	max96712_read_reg(priv, MAX96712_VIDEO_PIPE_EN_ADDR, &pipe_en_val);
```
sareluzi: **6/14 - Error from `max96712_read_reg` not checked**

`max96712_read_reg()` return value is ignored before using `pipe_en_val` to write back. If the read fails, you'll write garbage. Check `err` and bail if non-zero — same as `max96712_bind_ser_to_dser_pipe` already does.

nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch
```diff
@@ -322,20 +329,51 @@ index 8e237eef..33646ace 100644
...
++	max96712_write_reg(priv, MAX96712_VIDEO_PIPE_EN_ADDR, pipe_en_val &= ~(1 << pipe_id));
++	mutex_unlock(&priv->lock);
++
++	max96712_bind_ser_to_dser_pipe(dev, pipe_id, 0, 0);
```
sareluzi: **7/14 - `bind_ser_to_dser_pipe` called after releasing lock — not atomic**

```c
mutex_unlock(&priv->lock);
max96712_bind_ser_to_dser_pipe(dev, pipe_id, 0, 0);
```

`bind_ser_to_dser_pipe` takes the same lock internally so no deadlock, but release_pipe + unbind is not atomic. Another thread could allocate the pipe between the unlock and the bind call. Consider calling the register writes directly under the same lock instead of going through the exported function.

nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch
```diff
@@ -322,20 +329,51 @@ index 8e237eef..33646ace 100644
 +
 +int max96712_get_available_pipe_id(struct device *dev, int vc_id)
 +{
-+	/* TODO EHUD: Currently vc_id == pipe id but that's definitely
-+	 * not the end goal if we plan to support more than two cameras
-+	 */
-+	return vc_id;
++	int i;
++	int pipe_id = -ENOMEM;
```
sareluzi: **8/14 - `-ENOMEM` is wrong error code for pipe exhaustion**

```c
int pipe_id = -ENOMEM;
```

This isn't a memory error — it's resource exhaustion. `-ENOSR` (no stream resources) or `-ENOSPC` would be more accurate. Also note: in d4xx.c the new early-exit checks for `-ENOSR`, not `-ENOMEM`, so these need to be aligned or the early-exit will never trigger.

nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch
```diff
@@ -566,6 +643,9 @@ index 8e237eef..33646ace 100644
 +        gpio_direction_output(priv->pwdn_gpio, 1);
 +        gpio_set_value(priv->pwdn_gpio, 0);
 +        msleep(100);
++    } else {
++        max96712_write_reg(priv, MAX96712_PWR1_ADDR, 0x40);
```
sareluzi: **9/14 - Magic register value needs a comment**

```c
max96712_write_reg(priv, MAX96712_PWR1_ADDR, 0x40);
msleep(200);
```

Writing `0x40` to PWR1 (`0x13`) — what bit does this set? Presumably device reset? A brief comment or a `#define` would help.

Also: the commit message says "Allow for max96712 reset even without FG" — should "FG" be "GPIO"?

nvidia-oot/6.0/0007-Allow-external-address-assignment-on-max9295.patch
```diff
@@ -1,59 +1,162 @@
...
+-	};
+ 	struct reg_pair map_bpp8dbl[] = {
+ 		{0x0312, 0x0F},
+ 	};
+@@ -561,21 +560,17 @@ static int __max9295_set_pipe(struct device *dev, int pipe_id, u8 data_type1,
+ 
+ 	map_pipe_control[0].val = 0x40 | data_type1;
+ 	map_pipe_control[1].val = 0x40 | data_type2;
++	if (pipe_id == 0) {
++		map_pipe_control[1].val |= 0x80;
```
sareluzi, line 127: **10/14 - `pipe_x_val` static removed (good) but new logic needs verification**

The old code tracked `pipe_x_val` across calls via a `static` variable (which was itself problematic for multi-device). The replacement:
```c
if (pipe_id == 0) {
    map_pipe_control[1].val |= 0x80;
}
```
This unconditionally enables the multi-pipe bit only for pipe 0. In the old code, this bit was carried forward to all subsequent pipes via register `0x0315`. Is the intent that only pipe 0 gets multi-pipe enabled on the serializer? Worth verifying against the MAX9295 datasheet — register `0x0315` bit 7 (multi-pipe enable) may need to be set independently.

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-1-2-3.dts
```diff
@@ -0,0 +1,1467 @@
...
+					};
+					port@1 {
+						status = "okay";
+						reg = <1>;
+						d4xx_vi_in1: endpoint {
+							status = "okay";
+							vc-id = <1>;
+							port-index = <0>;
+							bus-width = <2>;
+							remote-endpoint = <&d4xx_csi_out1>;
```
sareluzi, line 66: **11/14 - 1467-line DT file is mostly copy-paste**

This is a very large hand-written DT file. Consider whether it can be generated from a template or share more structure with the existing `cams-0-1.dts` variant to reduce maintenance burden and avoid divergence bugs.

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-1-2-3.dts
```diff
@@ -0,0 +1,1467 @@
...
+							remote-endpoint = <&d4xx_csi_out3>;
+						};
+					};
+					port@4 {
+						status = "okay";
+						reg = <4>;
+						d4xx_vi_in4: endpoint {
+							status = "okay";
+							vc-id = <4>;
+							port-index = <0>;
```
sareluzi, line 97: **12/14 - All 16 channels route through `port-index = <0>` — confirm this is correct**

All VI input endpoints specify `port-index = <0>`, meaning all 16 virtual channels go through CSI port 0 on a single 2-lane (`bus-width = <2>`) MIPI link. With 4 cameras behind a single MAX96712, this is likely correct (single CSI-2 link with VCs), but worth confirming the hardware supports 16 VCs on a single 2-lane CSI port with sufficient bandwidth for 4 cameras.

sareluzi:
**13/14 - JP5.1.2 patch symlinks — verify they apply cleanly**

`kernel/nvidia/5.1.2/0008-Allow-external-address-assignment-on-max9295.patch` and `0009-Align-max9296-api-to-max96712.patch` are symlinks to the 5.0.2 versions. That's fine for identical patches, but the 5.0.2 and 5.1.2 max96712 patches (0017 vs 0007) are NOT symlinks and have different content. Verify the max9295 and max9296 patches truly apply cleanly to both 5.0.2 and 5.1.2 kernel source trees.

kernel/realsense/d4xx.c
ejgoldik: get_available_pipe_id returns ENOMEM, but in ds5_configure (which calls get_available_pipe_id):
if (sensor->pipe_id < 0) {
			dev_err(&state->client->dev, "No free pipe in %s\n",state->dser_ops->name);
			return -ENOSR;
}
I've checked, and this is does happen upon the ninth pipe attempted allocation

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-1-2-3.dts
ejgoldik: We have no other option. This is HW wiring, not a software choice. Also, while there are 16 VC's here, only 8 can be used at a time.

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-1-2-3.dts
ejgoldik: I've started looking into it in the past (using MACRO-like functions to simplify it) but it proved to be a big task. I'm definitely not saying it cannot be done, but it's a task on its own.

nvidia-oot/6.0/0007-Allow-external-address-assignment-on-max9295.patch
ejgoldik: The old code was basically chaos. Register 0x315 belongs to pipe x, and its MSB was always set "if pipe x was ever configured" but upon init we always configure all pipes in a loop so that was always true, so whenever a pipe which was not x was used, we also issued another write to an address in pipe x.

The new logic simply means "if you configure X, configue X, otherwise don't"

nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch
ejgoldik: ENOMEM was used here to be aligned with max9296, in any case as stated above, ds5_configure treats all errors an ENOSR anyway. Will change it for clarity.

nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch
ejgoldik: FG for fangzhu because only there the GPIO exists. will add a comment

nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch
ejgoldik: Will create an auxiliary function `__max96712_bind_ser_to_dser_pipe` which does not lock and use it within the locked scope

ejgoldik:
> **13/14 - JP5.1.2 patch symlinks — verify they apply cleanly**
> 
> `kernel/nvidia/5.1.2/0008-Allow-external-address-assignment-on-max9295.patch` and `0009-Align-max9296-api-to-max96712.patch` are symlinks to the 5.0.2 versions. That's fine for identical patches, but the 5.0.2 and 5.1.2 max96712 patches (0017 vs 0007) are NOT symlinks and have different content. Verify the max9295 and max9296 patches truly apply cleanly to both 5.0.2 and 5.1.2 kernel source trees.

I had to make it not a link as the base file max96712 differs by a tiny bit between 5.0.1 and 5.1.2 so you can't use a link. Verified it applies correctly.

sareluzi (approved):

PR #405, opened 2026-03-31
Update max9295 and max9296 driver configurations and patches

Nikolai-L:
Adjust the register addresses for the max9295 to support module removal without camera, change the regmap cache type to REGCACHE_NONE for both max9295 and max9296, and remove obsolete patches while updating versioning for nvidia-oot. This streamlines the driver and removes unnecessary complexity.


ejgoldik (approved):

sareluzi (commented):
**Bug: Dual-camera DT overlay not updated**

The single-camera overlay removes `ser_prim: max9295_prim@40` and moves `ser_a` from `@42` to `@40`, but `tegra234-camera-d4xx-overlay-dual.dts` still has the old two-node layout with `ser_prim@40` + `ser_a@42`.

If the register address change is needed for rmmod without camera, the dual-camera configuration will still have the old serializer address layout and may fail on rmmod without camera.

Should the same consolidation be applied to the dual-camera overlay?

Kontra2B (approved):

PR #406, opened 2026-03-31
Advantech board orin support

ejgoldik:
Added support for Advantech.
According to the vendor, this is for camera 0, but according to the device tree, this seems to be camera 4.

In a followup PR we will create another device tree for camera 0 and this device tree will either be removed or renamed to _4.dts

### DIFF
--- a/kernel/nvidia/5.1.2/0007-Adding-max96712-support-for-D4xx.patch
+++ b/kernel/nvidia/5.1.2/0007-Adding-max96712-support-for-D4xx.patch
@@ -529,7 +529,7 @@ index 329a10dfd..ebc6a1dfc 100644
 +	struct reg_pair map_phy_opt[] =
 +	{
 +		/* MIPI D-PHY Config */
-+		{MAX96712_MIPI_PHY0_ADDR, 0x04}, //0x8A0
++		{MAX96712_MIPI_PHY0_ADDR, 0x24}, //0x8A0
 +		{MAX96712_MIPI_TX10_ADDR, 0x50}, //0x94a - 2 MIPI lanes (Master controller for PHYA), and enable extended VC
 +		{MAX96712_MIPI_PHY3_ADDR, 0xe4}, //0x8A3, Lower phys to lower lanes
 +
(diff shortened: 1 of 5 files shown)


sareluzi:
@ejgoldik - did you verify it's working also on FG?

sareluzi (commented):
### `0x04 → 0x24` PHY register change affects ALL max96712 builds, not just Advantech

The three patch file changes (`kernel/nvidia/5.0.2/0017-...`, `kernel/nvidia/5.1.2/0007-...`, `nvidia-oot/6.0/0006-...`) modify `MAX96712_MIPI_PHY0` (reg 0x8A0) from `0x04` to `0x24`, enabling bit 5 (MIPI clock output 0). This is **not scoped to Advantech** — it changes the max96712 driver for all configurations on JP 5.0.2, 5.1.2, and 6.0. Could this break existing MAX96712 EVB setups?

Additionally, each JP version has **two** patches with the same name ("Adding-max96712-support-for-D4xx") that both define `map_phy_opt[]` with this register. Only the later-numbered patch is changed to `0x24`; the earlier one (0016 for JP5.0.2, 0006 for JP5.1.2, 0005 for JP6.0) still has `0x04`. If both patches contribute code to max96712.c, the inconsistency may matter.

---

### Makefile patch adds Advantech entry but confirm the DTS is copied during `apply_patches`

The overlay Makefile patch will reference `tegra234-camera-d4xx-overlay-advantech.dtbo`, but confirm the new `.dts` file is copied into the NVIDIA build tree during `apply_patches.sh`. Other DTS files have explicit copy steps — if this one is missing, the build will fail with "No rule to make target."

sareluzi:
@ejgoldik - what about JP7/Thor support?

kernel/nvidia/5.1.2/0007-Adding-max96712-support-for-D4xx.patch
```diff
@@ -529,7 +529,7 @@ index 329a10dfd..ebc6a1dfc 100644
 +	struct reg_pair map_phy_opt[] =
 +	{
 +		/* MIPI D-PHY Config */
-+		{MAX96712_MIPI_PHY0_ADDR, 0x04}, //0x8A0
++		{MAX96712_MIPI_PHY0_ADDR, 0x24}, //0x8A0
```
ejgoldik: Need to make this conditional (Will make a "force_clk_0" flag in the dts)

ejgoldik (commented):
Need to clk_0 force conditional (Will make a "force_clk_0" flag in the dts)

ejgoldik:
> @ejgoldik - did you verify it's working also on FG?

Verified now, works.

kernel/nvidia/5.1.2/0007-Adding-max96712-support-for-D4xx.patch
ejgoldik: Fixed

ejgoldik:
> @ejgoldik - what about JP7/Thor support?

Will be in follow-up

sareluzi (approved):

PR #407, opened 2026-04-01
tracked on RSDSO-21339: Unify Output Directory Structure and Jenkins Builds For JP4/5 and JP6/7

Kontra2B:
Generic output folder structure under images for all jetpack versions:
rootfs: boot, lib
boot: vmlinuz-x.y.z-tegra, dtb
lib: modules/x.y.z-tegra
Image name changes from Image to vmlinuz-x.y.z-tegra


build_all.sh
```diff
@@ -65,8 +68,8 @@ if [[ $CLEAN == 1 ]]; then
     rm -rf $BUILD_SRCS/out
 fi
 
-mkdir -p $TEGRA_KERNEL_OUT
-export KERNEL_MODULES_OUT=$TEGRA_KERNEL_OUT/modules
+ln -sfn $TEGRA_KERNEL_OUT $BUILD_SRCS/out
+ln -sfn $INSTALL_MOD_PATH $INSTALL_MOD_PATH/usr # some depmod require usr in the path
```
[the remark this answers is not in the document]
Kontra2B: required by some depmod for example on Suse

build_all.sh
```diff
@@ -77,17 +80,20 @@ export KERNEL_MODULES_OUT=$TEGRA_KERNEL_OUT/modules
 if version_lt "$JETPACK_VERSION" "6.0"; then
     #JP4/5
     cd $BUILD_SRCS/$KERNEL_DIR
-    make O=$TEGRA_KERNEL_OUT tegra_defconfig
+    make tegra_defconfig
     if [[ "$DEVDBG" == "1" ]]; then
         scripts/config --file $TEGRA_KERNEL_OUT/.config --enable DYNAMIC_DEBUG
     fi
```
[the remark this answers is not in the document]
Kontra2B: fixed with next commit

ejgoldik (changes requested):
Need to fix all the README files for the installation instruction as well as the deploy_kernel script for the path and file names change

scripts/setup-common
```diff
@@ -24,13 +24,13 @@ export ARCH=arm64
...
-    for i in 0 1 2; do
-        [[ v1[i] -lt v2[i] ]] && return 0
-        [[ v1[i] -gt v2[i] ]] && return 1
-    done
-    return 1
+	IFS='.' read -r -a v1 <<< "$1"
+	IFS='.' read -r -a v2 <<< "$2"
+	for i in 0 1 2; do
+		[[ v1[i] -lt v2[i] ]] && return 0
+		[[ v1[i] -gt v2[i] ]] && return 1
```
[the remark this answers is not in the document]
Kontra2B: Works fine. Not POSIX compliant. Let me know if we want POSIX compatybility. Then all [[ ]] entries should be replaced.

build_all.sh
```diff
@@ -56,7 +55,11 @@ else
 fi
 
 export LOCALVERSION=-tegra
-export TEGRA_KERNEL_OUT="$DEVDIR/images/${JP_INPUT_VERSION}"
+TEGRA_KERNEL_OUT="$DEVDIR/images/${JP_INPUT_VERSION}"	# ./images/5.1.6
+export INSTALL_MOD_PATH=${TEGRA_KERNEL_OUT}/rootfs/
+export INSTALL_PATH=${INSTALL_MOD_PATH}/boot/
```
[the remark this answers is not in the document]
Kontra2B: KERNELRELEASE is not ment to be assigned during regular build. Can be set by user as env variable though.

build_all.sh
```diff
@@ -77,17 +80,20 @@ export KERNEL_MODULES_OUT=$TEGRA_KERNEL_OUT/modules
...
-    make O=$TEGRA_KERNEL_OUT tegra_defconfig
+    make tegra_defconfig
     if [[ "$DEVDBG" == "1" ]]; then
         scripts/config --file $TEGRA_KERNEL_OUT/.config --enable DYNAMIC_DEBUG
     fi
-    make O=$TEGRA_KERNEL_OUT -j${NPROC}
-    make O=$TEGRA_KERNEL_OUT modules_install INSTALL_MOD_PATH=$KERNEL_MODULES_OUT
+    make -j$(nproc)
+    make -j$(nproc) modules
+    make dtbs
```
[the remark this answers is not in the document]
Kontra2B: fixed

apply_patches.sh
```diff
@@ -102,34 +103,34 @@ fi
...
+            ln -sfr hardware/realsense/tegra264-camera-d4xx-overlay*.dtso "${BUILD_SRCS}/$KERNEL_DIR/arch/arm64/boot/dts/nvidia/"
         fi
     fi
 
     # Stage all modified files after patching
     git -C "${BUILD_SRCS}/$D4XX_SRC_DST" add drivers/media/i2c/d4xx.c
     git -C "${BUILD_SRCS}/$D4XX_SRC_DST" add -u
-    [[ -d "${BUILD_SRCS}/$KERNEL_DIR" ]] && git -C "${BUILD_SRCS}/$KERNEL_DIR" add -A
+    [[ -d "${BUILD_SRCS}/$KERNEL_DIR" ]] && git -C "${BUILD_SRCS}/$KERNEL_DIR" add -u
     if [[ -d "${BUILD_SRCS}/hardware/nvidia/t23x/nv-public" ]]; then
```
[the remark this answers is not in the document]
Kontra2B: Symlinked files are managed by other repo (mipi driver). We do not want one file to exist in different repos. We do not want symlinks in the repos. Then we may have intermidate build artifacts from previous builds like o, cmd, ko files we do not want to add to sub-repos.

sareluzi:
@Kontra2B  - what is the motivation for the rename to vmlinuz-x.y.z-tegra?

sareluzi:
@Kontra2B  - I see you updated JP7 readme. What about all the rest?
Also - did you update install/deploy scripts accordingly?

Kontra2B:
> @Kontra2B - what is the motivation for the rename to vmlinuz-x.y.z-tegra?

To use linux kernel native makefile install target instead of bash install command. Native image names are hardcoded to vmlinux/vmlinuz-\<ver\>-\<localbuild\>. Localbuild is tegra in this case. There can be many image files with different names under /boot folder. The installed vmlinuz image bears kernel version in its name that facilitates handling of images with many boot options in extlinux. No need to manyally rename Image to Image.org, Image.back or Image.old which is errorprone. It is also easier to link initrd.img-\<ver\> file by explicit kernel \<ver\>

Kontra2B:
> @Kontra2B - I see you updated JP7 readme. What about all the rest? Also - did you update install/deploy scripts accordingly?

That was just a quick fix for Inessa to have it handy. Still working on readme files.

sareluzi:
## Comprehensive review

### `apply_patches_ext.sh` deleted but still referenced in multiple READMEs

`apply_patches_ext.sh` is deleted in this PR, but it's still referenced in:
- `README_JP4.md:57`
- `README_JP5.md:71`
- `README_JP6.0.md:86`
- `README_JP6.2.md:88`
- `README_tools.md:55, 60, 64`

These references will become broken instructions for users who build using the "external sources" workflow.

### README and deploy script gaps — comprehensive stale-path list

Building on @ejgoldik's earlier review — here is the complete list of files with stale paths that need updating for the new `rootfs/boot/vmlinuz-*-tegra` and `rootfs/boot/dtb/` structure:

**README_JP6.0.md** (6 stale refs):
- Lines 101, 137, 139: `Image` → `vmlinuz-*-tegra`
- Lines 103-104, 137, 283: DTBOs should be under `boot/dtb/`

**README_JP6.2.md** (8 stale refs):
- Lines 103, 131, 135: `Image` → `vmlinuz-*-tegra`
- Lines 105-108, 131, 292: DTBOs should be under `boot/dtb/`

**README_JP4.md** (line 111):
- `images/4.6.1/arch/arm64/boot/Image` → new unified rootfs path
- Lines 112-116: all artifact paths use old structure

**README_JP5.md** (lines 124-138):
- `images/5.1.2/arch/arm64/boot/Image` → new unified rootfs path
- scp commands on lines 133-138 use old paths

**README.md** (line 54):
- JP4 dtb path reference still uses old structure

**scripts/install_to_kernel.sh**:
- Line 11: help text says "copy Image"
- Lines 71-72: `boot/tegra234-camera-d4xx-overlay*.dtbo` → `boot/dtb/tegra234-...`
- Lines 77-84: `boot/Image` → `boot/vmlinuz-*-tegra`

**scripts/deploy_kernel.sh**:
- Line 54: `arch/arm64/boot/Image` for JP 5.0.2 uses old structure

### `vmlinux` vs `vmlinuz` naming inconsistency with Jenkins PR

The README_JP7.md in this PR uses `vmlinux-*-tegra` but the corresponding Jenkins PR (librealsense-deploy#82) uses `vmlinuz-*-tegra`. Which is the actual name produced by `make install`? The two PRs should be consistent.

apply_patches.sh
```diff
@@ -8,174 +8,145 @@ ACTION="apply"
...
+		JP5_D4XX_DTSI="tegra194-camera-d4xx-fg12-16ch.dtsi"
+	elif [[ "$1" == "--fg12-16ch-dual" ]]; then
+		JP5_D4XX_DTSI="tegra194-camera-d4xx-fg12-16ch-dual.dtsi"
+	elif [[ "$1" == reset ]]; then
+		ACTION="reset"
+	elif [[ $1 == "-h" ]]; then
+		echo Usage:
+		echo "$0 [--one-cam | --dual-cam | --max96712-EVB | --fg12-16ch | --fg12-16ch-dual ] [reset] [-h]"
+		echo -e 'reset\t: hard reset (git) to version from jetpack_version file'
+		echo -e '-h\t: show this help'
```
sareluzi, line 27: **Bug: `-h` flag leaks into version variable and `exit 0` is unreachable**

When `-h` is passed, the code sets `version=$1` (i.e. `version="-h"`), then `break`s out of the while loop. The `exit 0` two lines above the `break` is never reached because the execution order is:
```
echo Usage: ...
version=$1    # version="-h"
break         # exits the while loop — skips exit 0!
```
After the loop, `. scripts/setup-common ${version}` is called with `-h`, which will fail with "Wrong JetPack version -h!".

Fix: use `exit 0` instead of `break`, or move `exit 0` after the echo/help lines and before `version=$1`.

apply_patches.sh
```diff
@@ -8,174 +8,145 @@ ACTION="apply"
...
-        rm -rf "${BUILD_SRCS}/nvidia-oot/drivers/net/ethernet/nvidia/nvethernet/nvethernetrm"
-        ln -sf ../../../../../../nvethernetrm "${BUILD_SRCS}/nvidia-oot/drivers/net/ethernet/nvidia/nvethernet/nvethernetrm"
-    fi
+	if [[ "$ACTION" == reset ]] || [[ "$ACTION" == apply ]]; then
+		rm -rf "${BUILD_SRCS}/nvidia-oot/drivers/net/ethernet/nvidia/nvethernet/nvethernetrm"
+		ln -sf ../../../../../../nvethernetrm "${BUILD_SRCS}/nvidia-oot/drivers/net/ethernet/nvidia/nvethernet/nvethernetrm"
+	fi
 fi
 
 apply_external_patches() {
```
sareluzi, line 68: **`git add -u` won't stage new files added by patches**

`git add -u` only stages modifications to already-tracked files. If any patch adds a new file (e.g., a new DTS include), it will be left untracked and silently excluded from the commit.

I saw your response to the Copilot comment about not wanting symlinks committed — that makes sense. But this `git add -u` is inside `apply_external_patches()` which runs *before* the symlink creation, so it's capturing the `git apply` results. If any of those patches ever add new files (not just modify existing ones), they'll be missed. Consider `git add -A` with a pathspec if you want to be safe, or document the assumption that patches only modify existing files.

build_all.sh
sareluzi: **`KERNELRELEASE` is unset — DTBs path has empty trailing component**

`INSTALL_DTBS_PATH=${INSTALL_PATH}/dtb/${KERNELRELEASE}` uses `KERNELRELEASE` which is never assigned in this script, resulting in `images/<ver>/rootfs/boot/dtb/` (trailing slash, empty version component). Was this intentional (flat `dtb/` directory), or should `KERNELRELEASE` be populated after configuration, e.g.:
```bash
KERNELRELEASE=$(make -s -C $BUILD_SRCS/$KERNEL_DIR kernelrelease 2>/dev/null)
```

sareluzi (commented):
File-specific review comments below.

build_all.sh
sareluzi, line 72: **Circular symlink `rootfs/usr -> rootfs/`**

This creates `rootfs/usr` as a symlink pointing back to `rootfs/`, forming a directory cycle (`rootfs/usr/usr/usr/...`). Tools like `tar` (without `--no-recursion`), `rsync`, and recursive `find` may traverse this infinitely or fail.

The current deploy script's `tar czf ... -C rootfs boot lib` happens to avoid this by only archiving `boot` and `lib`, but a generic `tar` of `rootfs/` would loop. I see your note that this is required by some depmod (e.g. on SUSE) — if so, worth adding a comment explaining the reason, and ensuring packaging steps exclude the `usr` symlink.

kernel/kernel-jammy-src/Makefile
```diff
@@ -62,7 +53,7 @@ install:
 	@echo   "================================================================================"
 	@echo   "Installing $(KERNEL_SRC_DIR) sources"
 	@echo   "================================================================================"
-	install $(kernel_image) $(INSTALL_MOD_PATH)/boot/
+	$(MAKE) -C $(kernel_source_dir) install
```
sareluzi: **`make install` missing `$(O_OPT)` and `LOCALVERSION`**

The `install` target calls `$(MAKE) -C $(kernel_source_dir) install` but doesn't pass `$(O_OPT)` or `LOCALVERSION=$(version)`, while the `modules_install` invocation immediately below passes both. If the build used an out-of-tree output directory or relies on `LOCALVERSION` for the `-tegra` suffix, the kernel `install` step could install wrong artifacts or name them without the suffix.

Consider:
```makefile
$(MAKE) -C $(kernel_source_dir) $(O_OPT) LOCALVERSION=$(version) install
```

Same issue exists in `kernel/kernel-noble-src/Makefile`.

build_all.sh
Kontra2B: For now it is flat, but there is this option to keep dtb binaries under particular kernel version folder in the future.

apply_patches.sh
Kontra2B: Intended. -h is handled as version in setup-common that echoes additional common help for each scripts and then exits.

apply_patches.sh
Kontra2B: This would hold true only for a fresh build. Any subsequent build would commit many of existing compile time artifacts and symlinks.

build_all.sh
Kontra2B: Required by some depmod to build on Suse for example.

kernel/kernel-jammy-src/Makefile
Kontra2B: Fixed in next commit.

sareluzi: Verified in the latest commit — `$(O_OPT)` and `LOCALVERSION` now passed through to the kernel `install` target in both Makefiles. Thanks, resolving.

apply_patches.sh
sareluzi: Confirmed by the merged `-h`/empty-version branch in `setup-common` — the help path now exits cleanly. Resolving.

build_all.sh
sareluzi: OK — intentional flat layout for now. Minor suggestion: drop `${KERNELRELEASE}` from the path until it's actually populated, since the unused variable expansion produces a stray trailing slash and reads like a bug. Resolving either way.

apply_patches.sh
sareluzi: Understood — if the commit step's only purpose is to keep `reset` quiet, then yes the new-files concern is moot. Two options worth considering for cleanup: (a) drop the add/commit entirely and pass `--hard` with no confirmation prompt on `reset`; or (b) keep the commit but use `git add -A` so it's at least correct for the rare new-file case. Either way the current state is acceptable — flagging as a future cleanup.

build_all.sh
sareluzi: OK on the requirement. Could you add an inline comment in `build_all.sh` next to this line explaining the SUSE depmod dependency? Otherwise the next person looking at this will assume it's a bug and try to remove it. Once commented, this can be resolved.

sareluzi:
## Follow-up on stale paths after latest commits

Nice fix on `vmlinu?-*-tegra` to bridge the `vmlinux`/`vmlinuz` ambiguity (good single-char glob trick), and good README_JP7.md and README.md Thor cleanup.

However, the README/script gaps from my earlier comprehensive list are still open:

**Deleted `apply_patches_ext.sh` still referenced in:**
- `README_JP4.md:57`
- `README_JP5.md:71`
- `README_JP6.0.md:86`
- `README_JP6.2.md:88`
- `README_tools.md:55, 60, 64`

**Old `rootfs/boot/Image` and `rootfs/boot/*.dtbo` paths in:**
- `README_JP6.0.md` (lines 101, 103-104, 137, 139, 283)
- `README_JP6.2.md` (lines 103, 105-108, 131, 135, 292)
- `README_JP4.md:111-116` (old `arch/arm64/boot/Image` etc.)
- `README_JP5.md:124-138` (old `arch/arm64/boot/Image`, scp paths)
- `README.md:54` (old JP4 dtb path)

**Scripts using old paths:**
- `scripts/install_to_kernel.sh:11, 71-72, 77-84` — `Image` and `boot/*.dtbo` (not `boot/dtb/`)
- `scripts/deploy_kernel.sh:54` — old JP 5.0.2 `arch/arm64/boot/Image` path

Are these planned for a follow-up commit in this PR, or a separate PR? They will leave users with broken instructions and partially-broken deploy/install scripts if merged as-is.

build_all.sh
sareluzi: Inline comment added — thanks, that's exactly what I was asking for. Resolving.

apply_patches.sh
sareluzi: Bumping again — no action on this one in latest commits. Leaving this thread open as a reminder for a future cleanup pass: either drop the add/commit step entirely (per your note it's only there to keep `reset` quiet) or switch to `git add -A` so new files from patches aren't silently dropped. No blocker for this PR.

sareluzi:
## Follow-up after rebase

Good progress — `README_JP6.md` merge, `README_JP7.md` updates, and the `apply_patches_ext.sh` restoration address several earlier concerns. The SUSE depmod inline comment in `build_all.sh` also looks good.

Still open:

**Broken internal README links (new regression from the rename):**
- `README_JP7.md:20-21` — links to the now-deleted `README_JP6.2.md` and `README_JP6.0.md`; should point to `README_JP6.md`
- `README_JP6.md:23-24` — Links section references `README_JP6.0.md` (itself, deleted) and is missing a link to `README_JP7.md`

**`README_JP6.md` residual stale paths (carried over from the pre-merge copy):**
- Line 191: `LINUX /boot/Image` in extlinux example → should be `vmlinuz-*-tegra` (or the `vmlinu?-*-tegra` wildcard you use elsewhere)
- Line 227: `./images/6.2/rootfs/boot/` for DTBOs → should be `./images/6.2/rootfs/boot/dtb/`

**JP4/5 docs and scripts never updated, even though `build_all.sh` now uses the unified `INSTALL_MOD_PATH`/`INSTALL_PATH`/`INSTALL_DTBS_PATH` for JP4/5 too:**
- `README_JP4.md:111-132` — `images/4.6.1/arch/arm64/boot/Image`, `drivers/media/i2c/d4xx.ko` etc. all use the old non-unified tree
- `README_JP5.md:124-230` — same, plus old extlinux `LINUX /boot/Image` references
- `README.md:53` — old JP4 dtb path (`images/4.6.1/arch/arm64/boot/dts/...`)
- `scripts/install_to_kernel.sh:11,71-72,77-84` — help text and paths still reference `Image` and `boot/*.dtbo` (not `boot/dtb/`)
- `scripts/deploy_kernel.sh:54` — JP 5.0.2 branch still uses `${IMG_DIR}/arch/arm64/boot/Image`

If the JP4/5 unification isn't part of this PR's scope, worth narrowing the PR title (it currently says "for JP4/5 and JP6/7"). Otherwise these need updates before merge — `install_to_kernel.sh` in particular will silently fail to copy DTBOs and the kernel on JP6/7 targets because the paths have moved.

apply_patches.sh
Kontra2B: New files do not come from nowhere. They come from our mipi repo as a direct file or from patch. You can not use add -A as after kernel is build once you would add thousands of binaries and temp files like object files. We do not need to commit to kernel repo at all. The repos are unpushable. One big plus of commiting with add -u I missed to highlight is to separate new changes from those introduced by existing patches. Then a developer can see only the new changes wit diff and can easily create a new patch and commit it to our mipi repo. Otherwise it would be cumbersome to separate old/new changes.
To recap:
- no commit is required at all unless to our mipi repo
- can not commit with add -A
- commit with add -u facilitates separating old/new changes and the creation of new patches

Until recently we used to work with kernel and other source packages with no git functionality at all.

sareluzi: Makes sense — `git add -A` would grab compile artifacts, and `add -u` cleanly separates patch changes from build outputs. Resolving.

sareluzi:
## Follow-up after "update deploy scripts" + "update readme" commits

Good progress — both deploy scripts and most READMEs are now updated with the unified paths. `install_to_kernel.sh` and `deploy_kernel.sh` look clean.

**NEW: Copy-paste errors in `README_JP5.md`** — the README was rewritten using the JP6/7 template, but Xavier-specific details weren't adjusted:
- Headers say "AGX Orin™" (line ~36) and "AGX Thor™" (line ~71) — JP5 is **Xavier**
- extlinux example (lines ~157-168) shows `tegra234-p3737-0000+p3701-0000-nv.dtb` — Xavier uses `tegra194-p2888-0001-p2822-0000.dtb` (`tegra234` is Orin, `tegra194` is Xavier)
- Version typo on line ~100: `5.20.120-tegra` should be `5.10.120-tegra`

**Broken internal links (2 remaining):**
- `README_JP5.md:21` — links to deleted `./README_JP6.0.md`; should be `./README_JP6.md`
- `README_JP6.md:24` — links to deleted `./README_JP6.0.md`; should be removed or changed to self-reference `./README_JP6.md`

scripts/deploy_kernel.sh
ejgoldik, line 1: This fails on jetpack 6.0:
"
Install new modules to /lib/modules/5.15.136-tegra
Install new device tree files to /boot/dev/dtb
Install new image to /boot/dev
cp: cannot stat '6.0/boot/vmlinu?-5.15.136-tegra': No such file or directory
Shared connection to 192.168.48.26 closed.
"

ejgoldik (changes requested):
Deploy fails:
nstall new modules to /lib/modules/5.15.136-tegra
Install new device tree files to /boot/dev/dtb
Install new image to /boot/dev
cp: cannot stat '6.0/boot/vmlinu?-5.15.136-tegra': No such file or directory
Shared connection to 192.168.48.26 closed.


After that the device is left without an "image" file in boot/dev.
Rebooting it will cause the boot to fail

scripts/deploy_kernel.sh
Kontra2B: Worked for me:

Creating kernel_mod/6.0/rootfs tarball...
Copy install_to_kernel.sh to kernel_mod
nvidia@fw-orin-2.realsenseai.com's password:
Copying files to fw-orin-2.realsenseai.com host...
rootfs.tar.bzip2                                                                                                                        100%  101MB   5.5MB/s   00:18
install_to_kernel.sh                                                                                                                    100% 1592    17.9KB/s   00:00
Setting permissions on remote host
Execute ./install_to_kernel.sh on fw-orin-2.realsenseai.com host
Extracting rootfs tarball...
Installing 5.15.136-tegra kernel files for JetPack 6.0...
Remove existing /lib/modules/5.15.136-tegra
Install new modules to /lib/modules/5.15.136-tegra
Install new device tree files to /boot/dev/dtb
Install new image to /boot/dev
update-initramfs: Generating /boot/initrd.img-5.15.136-tegra
W: Couldn't identify type of root file system for fsck hook
Shared connection to fw-orin-2.realsenseai.com closed.

nvidia@fw-orin-2 ~> uname -a
Linux fw-orin-2 5.15.136-tegra #4 SMP PREEMPT Mon Apr 27 10:39:42 CEST 2026 aarch64 aarch64 aarch64 GNU/Linux
nvidia@fw-orin-2 ~> ls /dev/video*
/dev/video0   /dev/video15  /dev/video21  /dev/video3  /dev/video-rs-color-0     /dev/video-rs-color-md-3  /dev/video-rs-depth-md-2  /dev/video-rs-ir-1
/dev/video1   /dev/video16  /dev/video22  /dev/video4  /dev/video-rs-color-1     /dev/video-rs-depth-0     /dev/video-rs-depth-md-3  /dev/video-rs-ir-2
/dev/video10  /dev/video17  /dev/video23  /dev/video5  /dev/video-rs-color-2     /dev/video-rs-depth-1     /dev/video-rs-imu-0       /dev/video-rs-ir-3
/dev/video11  /dev/video18  /dev/video24  /dev/video6  /dev/video-rs-color-3     /dev/video-rs-depth-2     /dev/video-rs-imu-1       /dev/video-rs-ir-md-0
/dev/video12  /dev/video19  /dev/video25  /dev/video7  /dev/video-rs-color-md-0  /dev/video-rs-depth-3     /dev/video-rs-imu-2       /dev/video-rs-ir-md-1
/dev/video13  /dev/video2   /dev/video26  /dev/video8  /dev/video-rs-color-md-1  /dev/video-rs-depth-md-0  /dev/video-rs-imu-3       /dev/video-rs-ir-md-2
/dev/video14  /dev/video20  /dev/video27  /dev/video9  /dev/video-rs-color-md-2  /dev/video-rs-depth-md-1  /dev/video-rs-ir-0        /dev/video-rs-ir-md-3
nvidia@fw-orin-2 ~> ls /boot/dev -lrt
total 40800
drwxr-xr-x 2 root root     4096 Apr 27 11:47 dtb
-rw-r--r-- 1 root root 41771520 Apr 27 11:47 vmlinux-5.15.136-tegra
lrwxrwxrwx 1 root root       32 Apr 27 11:47 Image -> /boot/dev/vmlinux-5.15.136-tegra
lrwxrwxrwx 1 root root       31 Apr 27 11:47 initrd -> /boot/initrd.img-5.15.136-tegra
nvidia@fw-orin-2 ~> cat /var/log/dmesg|head -n 10
[    0.000000] kernel: Booting Linux on physical CPU 0x0000000000 [0x410fd421]
[    0.000000] kernel: Linux version 5.15.136-tegra (wojtek@DESKTOP-KJR3ABI) (aarch64-buildroot-linux-gnu-gcc.br_real (Buildroot 2022.08) 11.3.0, GNU ld (GNU Binutils) 2.38) #4 SMP PREEMPT Mon Apr 27 10:39:42 CEST 2026 ()

Kontra2B: @ejgoldik Did you build in parallel many JP versions?

Kontra2B: image

Kontra2B: Indeed an error reproduced.
Fixed and rebased with next push.

scripts/deploy_kernel.sh
```diff
@@ -1,81 +1,52 @@
...
+    echo "  REMOTE_USER       Username on target host (default: administrator)"
+    echo "  COPY_TO_PATH      Remote path to copy files to (default: dev)"
+    echo "  TARGET_FOLDER     Folder name on target host to install kernel files to (default: /boot/dev)"
     echo ""
     echo "Example:"
-    echo "  $0 6.2 192.168.1.100 - pack and copy to /home/administrator/dev, update /boot/dev and reboot TARGET"
-    echo "  $0 6.2 192.168.1.100 nvidia - pack and copy to /home/nvidia/dev, update /boot/dev and reboot TARGET"
-    echo "  $0 6.2 192.168.1.100 nvidia foo - pack and copy to /home/nvidia/foo, update /boot/dev and reboot TARGET"
-    echo "  $0 6.2 192.168.1.100 nvidia foo bar - pack and copy to /home/nvidia/foo, update /boot/bar and reboot TARGET"
+    echo "  $0 6.2 - pack and copy to localhost to /tmp/kernel_mod/dev folder, install to localhost /boot/dev/ and reboot localhost"
```
ejgoldik: Do not change the behaviour of the installation.
e.g:
When running
"./scripts/deploy_kernel.sh 6.2 <host> nvidia ejgoldik dev"
The expected path on the host is:
~/ejgoldik/kernel_mod/6.2
with the rootfs and install_to_kernel.sh in it

Note that deploy kernel (in its latest version) also copies the external sync generator script to this folder, so it shouldn't be removed after installation (it also allows us to have the latest build saved for quick a-b testing)

ejgoldik (changes requested):
Return deploy_kernel host directory to its original path and rebase

scripts/deploy_kernel.sh
Kontra2B: Fixed with new commit

scripts/deploy_kernel.sh
```diff
@@ -1,81 +1,54 @@
...
+tar -cjf ${DEST_DIR}/rootfs.tar.bzip2 -C images/${JETPACK_VERSION}/rootfs boot lib
+echo "Copy install_to_kernel.sh to kernel_mod/"
+cp scripts/install_to_kernel.sh kernel_mod/
+echo "Copy ext_syn_gen.py to kernel_mod/"
+cp scripts/ext_sync_gen.py kernel_mod/${JETPACK_VERSION}/
+
+# Use SSH ControlMaster to reuse a single SSH connection
+CONTROL_PATH="/tmp/ssh-control-${USERNAME}-${TARGET}"
+ssh -MS "${CONTROL_PATH}" -fN ${USERNAME}@${TARGET}
+ssh -S "${CONTROL_PATH}" "${USERNAME}@${TARGET}" "rm -rf kernel_mod/${REMOTE_PATH}/${JETPACK_VERSION} && mkdir -p kernel_mod/${REMOTE_PATH}"
```
ejgoldik: Path is still wrong.
When running:
./scripts/deploy_kernel.sh 6.0 <hostname> nvidia ejgoldik dev
It creates the output under 
"kernel_mod/ejgoldik/6.0"
Instead of
"ejgoldik/kernel_mod/6.0"

ejgoldik (changes requested):

scripts/deploy_kernel.sh
Kontra2B: My last install results:

image

image

ejgoldik: Why is "dev" part of the path?
If you run:
./scripts/deploy_kernel.sh 6.0 <host> nvidia wojtek dev
Where is it copied to?
I would expect:
~/wojtek/kernel_mod/6.0

Please checkout dev branch, and verify your PR and dev branch copy to the same path

Kontra2B: Run scripts/deploy_kernel.sh -h for param description.
In this invocation: ./scripts/deploy_kernel.sh 6.0 nvidia wojtek dev:
nvidia - remote host to ssh to, default localhost
wojtek - remote user, default current user name on local host you run the script on
dev - folder under kernel_mod folder under remote user home folder, default dev
Thus when issued target folder on remote host is ~/kernel_mod/dev

Kontra2B: I guess dev should only specify target folder to install image to? Not the folder under ~/kernel_mod?

Kontra2B: Run from dev branch

image

image

ejgoldik: > I guess dev should only specify target folder to install image to? Not the folder under ~/kernel_mod?

Correct

This line is incorrect, it's missing host name:
./scripts/deploy_kernel.sh 6.0 nvidia wojtek dev

It should be
./scripts/deploy_kernel.sh 6.0 <host_name e.g fw_orin_2> nvidia wojtek dev:
With:
<host_name> - device host name or i.p
nvidia - user (for most of our orins its nvidia)
wojtek - directory under home in which to put the 6.0 directory
dev - directory under /boot/ in which to put the image

Kontra2B: The script from dev:HEAD fails

image

image

Kontra2B: Is ext_sync_get.py script required under each JP version folder (6.2)? Or can be copied to kernel_mod (one level up). Is it any different for each JP version?

PR #409, opened 2026-04-06
Fix Jenkins JP5 build: update apply_patches_ext.sh for renamed DT dirs

sareluzi:
## Summary
- Commit 15291ac (JP7/Thor support) renamed DT patch directories from version-specific names (`4.6.1/`, `5.0.2/`, `5.1.2/`) to generic ones (`4.x/`, `5.x/`) and updated `apply_patches.sh`, but missed `apply_patches_ext.sh` (used by Jenkins)
- Jenkins JP5 builds fail with `ls: cannot access '.../kernel-dts/5.1.2': No such file or directory`
- Uses `$JETPACK_VERSION` (mapped `5.x`/`4.x`) instead of raw `$1` (`5.1.2`) for DT patch lookups
- Adds JP7 exclusions to prevent running JP4/5-only logic on JP7 builds

## Test plan
- [ ] Verify Jenkins JP5 (5.1.2) build passes
- [ ] Verify Jenkins JP4 (4.6.1) build passes
- [ ] Verify JP6 builds are unaffected

### DIFF
--- a/apply_patches_ext.sh
+++ b/apply_patches_ext.sh
@@ -24,16 +24,16 @@ fi
 . scripts/setup-common
 
 # set JP4 devicetree
-if [[ "$JETPACK_VERSION" == "4.6.1" ]]; then
+if [[ "$JETPACK_VERSION" == "4.x" ]]; then
     JP5_D4XX_DTSI="tegra194-camera-d4xx.dtsi"
 fi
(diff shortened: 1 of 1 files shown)


ejgoldik (approved):
Need to add JP7.x support in follow up

PR #416, opened 2026-04-14
Optimize reset and probe

Nikolai-L:
* Refactor I2C handling and remove legacy SERDES recovery code
 > * Updated polling semantics to utilize ds5_read_poll() for transient I2C failures, reducing false warnings and log spam.
 > * Removed legacy SERDES recovery code, simplifying the reset process and focusing on readiness polling.
 > * Adjusted hardware reset parameters to improve stability and recovery times.
* Replace hard link creation for d4xx.c in apply_patches.sh with symlink one
* Increase initial reset delay to reduce poll iterations and to improve probe efficiency


kernel/realsense/d4xx.c
```diff
@@ -1977,22 +1977,16 @@ static int ds5_configure(struct ds5 *state)
 	vc_id = (state->is_depth) ? 0 : (state->is_rgb) ? 1 : (state->is_y8) ? 2 : 3;
 #endif
 
-	fmt = sensor->streaming ? sensor->config.format->data_type : 0;
```
sareluzi, line 1980: why was this removed?

kernel/realsense/d4xx.c
```diff
@@ -2008,15 +2002,13 @@ static int ds5_configure(struct ds5 *state)
 		sensor->cached_md_value = md_value;
 	}
 
-	if (!sensor->streaming)
```
sareluzi, line 2011: why was this removed?

apply_patches.sh
```diff
@@ -102,7 +102,7 @@ fi
 echo "Patches applied successfully"
 
 if [[ "$ACTION" = "apply" ]]; then
-    version_lt "$JETPACK_VERSION" "5.0" || ln -f kernel/realsense/d4xx.c "${BUILD_SRCS}/${D4XX_SRC_DST}/drivers/media/i2c/"
+    version_lt "$JETPACK_VERSION" "5.0" || ln -f -s kernel/realsense/d4xx.c "${BUILD_SRCS}/${D4XX_SRC_DST}/drivers/media/i2c/"
```
sareluzi: **Bug:** `ln -s` resolves the target **relative to the link's directory**, not the working directory. This creates a symlink at e.g. `sources_6.2/nvidia-oot/drivers/media/i2c/d4xx.c` pointing to `kernel/realsense/d4xx.c`, which resolves to `sources_6.2/nvidia-oot/drivers/media/i2c/kernel/realsense/d4xx.c` — a path that doesn't exist. The link will be dangling and the build will fail.

Fix: use an absolute path:
```bash
ln -f -s "$(pwd)/kernel/realsense/d4xx.c" "${BUILD_SRCS}/${D4XX_SRC_DST}/drivers/media/i2c/"
```

kernel/realsense/d4xx.c
```diff
@@ -2359,20 +2351,13 @@ static int ds5_set_calibration_data(struct ds5 *state,
 }
 
 /* HW reset timeout and polling parameters */
-#define DS5_HW_RESET_INITIAL_DELAY_MS	100
+#define DS5_HW_RESET_INITIAL_DELAY_MS	500
```
sareluzi, line 2358: Commit message says *"Reduce initial delay for hardware reset to improve probe efficiency"* but this changes `DS5_HW_RESET_INITIAL_DELAY_MS` from **100 → 500** — a 5× increase. The intent may be that a longer initial wait reduces total poll iterations, but the commit message is misleading. Should be reworded (e.g., "Increase initial reset delay to reduce poll iterations").

kernel/realsense/d4xx.c
```diff
@@ -1955,8 +1953,10 @@ static int ds5_configure(struct ds5 *state)
 			dev_err(&state->client->dev, "No free pipe in %s\n",state->dser_ops->name);
 			return -ENOSR;
 		}
+		mutex_lock(&serdes_lock__);
 		ret = ds5_setup_pipeline(state, data_type1, data_type2,
 					 sensor->pipe_id, vc_id);
```
sareluzi, line 1958: The new `serdes_lock__` around `ds5_setup_pipeline()` is a good addition (it writes shared serializer/deserializer registers). However, `reset_oneshot` a few lines below is called *after* the lock is released. Other `reset_oneshot` calls in the codebase are typically done under `serdes_lock__`. Should the lock extend to cover `reset_oneshot` as well?

kernel/realsense/d4xx.c
```diff
@@ -5211,22 +4895,29 @@ static int ds5_mux_s_stream(struct v4l2_subdev *sd, int on)
...
-					stream_id, stream_cmd, i);
-				continue;
-			}
-			ret = ds5_read(state, stream_status_base, &streaming);
-			if (ret < 0) {
-				dev_warn(&state->client->dev,
-					"stream %d status i2c read failed (%d), retry %u\n", stream_id, ret, i);
+				dev_warn(&state->client->dev, "stream %d cmd 0x%x write failed, retry %d, %dms\n",
+					stream_id, stream_cmd, i, jiffies_to_msecs(jiffies - ts));
 			}
```
sareluzi, line 4900: Behavioral change: the old code always did `continue` after writing the stream command, giving FW one poll interval to process. Now the status is read immediately after write and checked in the same iteration. This will produce extra `"status not as expected"` warnings in the log on most first iterations (FW hasn't processed the command yet).

Also, if the write fails (`ret < 0`), the old code skipped the status read via `continue`. Now it falls through and reads status anyway. Not harmful but slightly noisier.

kernel/realsense/d4xx.c
```diff
@@ -2444,266 +2413,18 @@ static void ds5_reset_streaming_flags(struct ds5_dev *ds5_dev)
...
-	 *	This results in all cameras connected to this deserializer
-	 *	invalidating serializer and ds5 caches,
-	 *	ensuring any stale state from the reset is cleared.
-	*/
-	atomic_inc(dser_get_reset_gen(state));
-
-	return 0;
-}
-#endif /* CONFIG_VIDEO_D4XX_SERDES */
+static int ds5_gmsl_serdes_setup(struct ds5 *state);
```
sareluzi: This forward declaration appears unnecessary — `ds5_gmsl_serdes_setup()` is defined at ~line 4233 and only called from `ds5_probe()` at ~line 4323, both after this insertion point. No code between here and the definition calls it. Looks like a leftover from an earlier iteration — remove it.

kernel/realsense/d4xx.c
```diff
@@ -5235,22 +4926,24 @@ static int ds5_mux_s_stream(struct v4l2_subdev *sd, int on)
...
 				ds5_config_cache_clear(sensor);
 			} else {
 				dev_warn(&state->client->dev,
-					"stream %d config failed after %d retries, aborting\n", stream_id, i);
+					"stream %d config failed after %d retries, aborting, %dms\n",
+					stream_id, i, jiffies_to_msecs(jiffies - ts));
 				break;
 			}
 			continue;
 		}
```
sareluzi, line 4943: Nit: promoting from `dev_dbg` to `dev_info` means this prints on every successful stream start/stop. With 4 subdevices per camera (depth, rgb, ir, imu), that's 4–8 info messages per stream cycle. Fine for development, but consider whether this should stay as `dev_dbg` for production log volume.

kernel/realsense/d4xx.c
```diff
@@ -1977,22 +1977,16 @@ static int ds5_configure(struct ds5 *state)
...
-		 sensor->config.format->data_type == GMSL_CSI_DT_YUV422_8)
+	else if (state->is_y8 && dt_value == GMSL_CSI_DT_YUV422_8)
 		dt_value = 0x32;
 
-	dev_dbg(&state->client->dev, "sensor %p: dt_value=0x%x, cached_dt_value=0x%x, cached_fps_value=%u, framerate=%u\n",
-			sensor, dt_value, sensor->cached_dt_value, sensor->cached_fps_value, sensor->config.framerate);
-
 	if (sensor->cached_dt_value != dt_value) {
 		ret = ds5_write(state, dt_addr, dt_value);
 		if (ret < 0)
```
sareluzi, line 1996: Nit: the removed `dev_dbg` that showed `dt_value`, `cached_dt_value`, `cached_fps_value`, and `framerate` was useful for diagnosing config-cache mismatches. Consider keeping it.

kernel/realsense/d4xx.c
sareluzi: This removes the entire tiered SERDES recovery (Phase 1 serializer re-init + Phase 2 full deserializer reset) and the `retry_after_serdes_recovery` fallback in `ds5_mux_s_stream()`. If a GMSL link drops *outside* of a HW reset (e.g., cable issue, EMI, or delayed camera FW failure), there's now no recovery path — the stream start will just time out and fail.

Have we validated this enough? Specifically — have we seen the old tiered recovery actually trigger in production logs, and if so, was it ever successful? If the logs show it never triggered (or never succeeded), then removing it is justified cleanup. But if it was saving real link-drop scenarios, this is a significant resilience regression.

kernel/realsense/d4xx.c
```diff
@@ -6645,40 +6284,14 @@ static int ds5_probe(struct i2c_client *c
...
-		}
-
-		/* Wait after HW reset before touching MAX9295 serializer registers.
-		 * This delay helps ensure the device is ready.
-		 */
-		msleep(200);
-	}
-
-	/* Verify post-reset format-discovery readiness.
+	/* Verify format-discovery readiness.
```
sareluzi: Removing the probe-time HW reset means the camera inherits whatever state it had from a previous driver load/unload. `ds5_wait_device_type()` will detect an invalid DEVICE_TYPE but just fails the probe rather than recovering.

Please make sure this is validated with `rmmod` / `insmod` cycles (including cases where the previous unload happened mid-stream or after a FW hang).

Nikolai-L (commented):
> why was this removed? 

Unnecessary clutter

kernel/realsense/d4xx.c
Nikolai-L: Unnecessary clutter

kernel/realsense/d4xx.c
Nikolai-L: Done

kernel/realsense/d4xx.c
Nikolai-L: Reverted

kernel/realsense/d4xx.c
Nikolai-L: Didn't make any sense:
The normal flow can't make it here with !sensor->streaming

kernel/realsense/d4xx.c
Nikolai-L: No saving real link-drop scenarios were observed

kernel/realsense/d4xx.c
Nikolai-L: By design, it is command received status and could be read back immidiately

kernel/realsense/d4xx.c
Nikolai-L: 1 info message per stream start with the elapsed time - prefer to keep it

kernel/realsense/d4xx.c
Nikolai-L: I see disadvantages both ways.
Take @ev-mp and decide.

apply_patches.sh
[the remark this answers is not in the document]
Nikolai-L: Right, fixed

Nikolai-L: Right, fixed

kernel/realsense/d4xx.c
sareluzi: Fair — decision on design direction is yours + @ev-mp. From my side, please confirm that the **removal was tested via rmmod/insmod cycles**, including at least one case where the previous unload happened mid-stream or after a FW hang. That's the scenario the probe-time reset was covering. If tested clean, I'm fine with the change.

Nikolai-L: We don't have such tests so I'm keeping the reset

kernel/realsense/d4xx.c
```diff
@@ -1977,22 +1976,19 @@ static int ds5_configure(struct ds5 *state)
 	vc_id = (state->is_depth) ? 0 : (state->is_rgb) ? 1 : (state->is_y8) ? 2 : 3;
 #endif
 
-	fmt = sensor->streaming ? sensor->config.format->data_type : 0;
-
+	dev_dbg(&state->client->dev,
+		"sensor %p: dt_value=0x%x, cached_dt_value=0x%x, cached_fps_value=%u, framerate=%u\n",
```
sareluzi: **Bug introduced by the revert:** the `dev_dbg` was restored, but it now runs *before* `dt_value` is computed.

```c
    dev_dbg(&state->client->dev,
        "sensor %p: dt_value=0x%x, cached_dt_value=0x%x, ...",
        sensor, dt_value, ...);            // dt_value still 0 (declaration initializer)
    /* Determine desired data-type... */
    dt_value = sensor->config.format->data_type;   // actual assignment AFTER the log
    if (state->is_depth && dt_value != 0)
        dt_value = 0x31;
    ...
```

This will always print `dt_value=0x0`. In the original code the log was placed *after* the special-case computation. Please move this `dev_dbg` down to just before the `if (sensor->cached_dt_value != dt_value)` check.

kernel/realsense/d4xx.c
```diff
@@ -2817,11 +2538,13 @@ static int ds5_hw_reset_with_recovery(struct ds5 *state)
...
 
-	/* 5. Brief delay to allow reset to begin */
+	/* 5. Delay to allow reset to complete */
+	ts = jiffies;
 	msleep(DS5_HW_RESET_INITIAL_DELAY_MS);
 
 	/* 6. Poll for control-status defaults to confirm reset completion. */
-	for (retry = 0; retry < DS5_HW_RESET_MAX_RETRIES; retry++, msleep(DS5_HW_RESET_POLL_INTERVAL_MS)) {
+	for (retry = 0, timeout = ts + msecs_to_jiffies(DS5_HW_RESET_TIMEOUT_MS);
+			time_before(jiffies, timeout); retry++, msleep_range(DS5_HW_RESET_POLL_INTERVAL_MS)) {
```
sareluzi: **Two issues with the new time-based loop structure** (the post-loop check and error message at ~d4xx.c:2565-2569 aren't shown in the diff but are affected):

**1. Loop-exit check is semantically inconsistent (Copilot also flagged this).**

The loop now exits on time expiry, but the post-loop check is still counter-based:
```c
for (retry = 0, timeout = ts + msecs_to_jiffies(DS5_HW_RESET_TIMEOUT_MS);
     time_before(jiffies, timeout); retry++, msleep_range(...)) {
    ...
    if (ready_status == DS5_HW_RESET_READY_EXPECTED_VAL) {
        ...
        break;
    }
}

if (retry >= DS5_HW_RESET_MAX_RETRIES) {   // <-- inconsistent with time-based exit
```
Under scheduler pressure `msleep_range` can take longer than nominal; the loop may exit via time expiry with `retry < MAX_RETRIES`, and the function then falls through the `if` block *as if the device was ready* even though it wasn't.

Fix: track readiness explicitly:
```c
bool ready = false;
for (...) {
    ...
    if (ready_status == DS5_HW_RESET_READY_EXPECTED_VAL) {
        ready = true;
        break;
    }
}
if (!ready) {
    ... /* DFU fallback + return -ETIMEDOUT */
}
```

**2. Error message elapsed-time is off.**

The subsequent `dev_err(... "Device did not become ready after %d ms ...", DS5_HW_RESET_INITIAL_DELAY_MS + DS5_HW_RESET_TIMEOUT_MS, ...)` prints 10500 ms, but `ts` was captured *before* the initial `msleep` and the loop budget is `timeout = ts + TIMEOUT_MS`, so actual wall time from `ts` is ~10000 ms. The success path already uses `jiffies_to_msecs(jiffies - ts)` — use the same here for consistency.

kernel/realsense/d4xx.c
Nikolai-L: It wasn't introduced by the revert but still is worth fixing 👍

kernel/realsense/d4xx.c
Nikolai-L: Yep, doesn't look good, fixed

ejgoldik (approved):

PR #421, opened 2026-04-19
Added driver_version.sh script under new jenkins directory 

cinessa:
script increments version of driver


cinessa:
added driver_version.sh file

PR #425, opened 2026-04-23
Fix endless HW reset when camera inaccessible

ejgoldik:
Timeout check was previously made **after** ds5_read_poll, which continues if fails. So if the camera is dead, it will poll forever and never reach the timeout check.


sareluzi (approved):
LGTM; get Niko's approval also

Nikolai-L (approved):
Good catch!

PR #429, opened 2026-04-27
RSDEV-6847: On-demand ESYNC GPIO tunneling for external sync

ipilchin1:
Replace probe-time one-shot GPIO tunneling setup with on-demand enable/disable tied to the sync_mode V4L2 control.

Changes:

d4xx.c: 
Add ds5_set_ser_esync_tunneling() to enable/disable MAX9295A serializer GPIO tunneling.
Enable ESYNC when sync_mode is set to slave (2) or master (3), disable otherwise. 
Disable ESYNC after HW reset to ensure clean state. 
Remove probe-time max9295_setup_gpio_tunneling() call.

max9295 patches (JP 5.x.x, 5.1.2, 6.x.x, 7.x): 
Rename max9295_setup_gpio_tunneling() → max9295_enable_gpio_tunneling(), add max9295_disable_gpio_tunneling() to reset ESYNC registers, export both.

Why: GPIO tunneling was previously enabled unconditionally at probe and never disabled, which could interfere with cameras not using external sync. This makes ESYNC tunneling follow the actual sync mode requested by userspace.

### DIFF
--- a/kernel/nvidia/5.0.2/0015-Add-max9295-GPIO-tunneling-for-the-external-sync-support.patch
+++ b/kernel/nvidia/5.0.2/0015-Add-max9295-GPIO-tunneling-for-the-external-sync-support.patch
@@ -6,7 +6,7 @@ Subject: [PATCH] Add max9295 GPIO tunneling for the external sync support
 Add MAX9295A serializer GPIO tunneling / ESYNC register configuration
 for MAX96712A external sync support.
 
-Includes max9295_setup_gpio_tunneling() export for MAX96712A usage.
+Includes max9295_enable_gpio_tunneling() export for MAX96712A usage.
 
 Signed-off-by: ejgoldik <ehud.joseph.goldik@realsenseai.com>
 ---
@@ -45,11 +45,11 @@ index 4b03b1ef8..d7226af6e 100644
  
  #define MAX9295_MAX_PIPES 0x4
  
-@@ -383,6 +395,61 @@ error:
+@@ -383,6 +395,92 @@ error:
  }
  EXPORT_SYMBOL(max9295_setup_control);
  
-+int max9295_setup_gpio_tunneling(struct device *dev)
++int max9295_enable_gpio_tunneling(struct device *dev)
 +{
 +	struct max9295 *priv = dev_get_drvdata(dev);
 +	int err = 0;
@@ -102,7 +102,38 @@ index 4b03b1ef8..d7226af6e 100644
 +
 +	return err;
 +}
-+EXPORT_SYMBOL(max9295_setup_gpio_tunneling);
++EXPORT_SYMBOL(max9295_enable_gpio_tunneling);
++
++int max9295_disable_gpio_tunneling(struct device *dev)
++{
++	struct max9295 *priv = dev_get_drvdata(dev);
++	int err = 0;
++	unsigned int readback = 0;
++
++	mutex_lock(&priv->lock);
++
++	max9295_write_reg(dev, MAX9295_SRC_PWDN_ADDR, MAX9295_PWDN_GPIO);
++	regmap_read(priv->regmap, MAX9295_SRC_PWDN_ADDR, &readback);
++	dev_info(dev, "%s: reg 0x%04x written 0x%02x readback 0x%02x %s\n",
++		__func__, MAX9295_SRC_PWDN_ADDR, MAX9295_PWDN_GPIO,
++		readback, (readback == MAX9295_PWDN_GPIO) ? "OK" : "MISMATCH");
++
++	max9295_write_reg(dev, MAX9295_SRC_CTRL_ADDR, MAX9295_RESET_SRC);
++	regmap_read(priv->regmap, MAX9295_SRC_CTRL_ADDR, &readback);
++	dev_info(dev, "%s: reg 0x%04x written 0x%02x readback 0x%02x %s\n",
++		__func__, MAX9295_SRC_CTRL_ADDR, MAX9295_RESET_SRC,
++		readback, (readback == MAX9295_RESET_SRC) ? "OK" : "MISMATCH");
++
++	max9295_write_reg(dev, MAX9295_2C0_ADDR, 0x00);
++	max9295_write_reg(dev, MAX9295_2C1_ADDR, 0x00);
++	max9295_write_reg(dev, MAX9295_2C2_ADDR, 0x00);
++	max9295_write_reg(dev, MAX9295_2C3_ADDR, 0x00);
++
++	mutex_unlock(&priv->lock);
++
++	return err;
++}
++EXPORT_SYMBOL(max9295_disable_gpio_tunneling);
 +
  int max9295_reset_control(struct device *dev)
  {
@@ -111,12 +142,13 @@ diff --git a/include/media/max9295.h b/include/media/max9295.h
 index bea15c414..6099859d2 100644
 --- a/include/media/max9295.h
 +++ b/include/media/max9295.h
-@@ -99,6 +99,9 @@ int max9295_sdev_unpair(struct device *dev, struct device *s_dev);
+@@ -99,6 +99,10 @@ int max9295_sdev_unpair(struct device *dev, struct device *s_dev);
  int max9295_setup_streaming(struct device *dev);
  
  int max9295_init_settings(struct device *dev);
 +
-+int max9295_setup_gpio_tunneling(struct device *dev);
++int max9295_enable_gpio_tunneling(struct device *dev);
++int max9295_disable_gpio_tunneling(struct device *dev);
 +
  /** @} */
  
--- a/kernel/nvidia/5.0.2/0015-Add-max9295-GPIO-tunneling-for-the-external-sync-support.patch
+++ b/kernel/nvidia/5.0.2/0015-Add-max9295-GPIO-tunneling-for-the-external-sync-support.patch
@@ -9,10 +9,11 @@ for MAX96712A external sync support.
 Includes max9295_enable_gpio_tunneling() export for MAX96712A usage.
 
 Signed-off-by: ejgoldik <ehud.joseph.goldik@realsenseai.com>
+Signed-off-by: ipilchin1 <inga.pilchin@realsenseai.com>
 ---
- drivers/media/i2c/max9295.c | 67 +++++++++++++++++++++++++++++++++++++++++++++
- include/media/max9295.h     |  3 +++
- 2 files changed, 70 insertions(+)
+ drivers/media/i2c/max9295.c | 69 +++++++++++++++++++++++++++++++++++++++++++++
+ include/media/max9295.h     |  4 +++
+ 2 files changed, 73 insertions(+)
 
 diff --git a/drivers/media/i2c/max9295.c b/drivers/media/i2c/max9295.c
 index 4b03b1ef8..d7226af6e 100644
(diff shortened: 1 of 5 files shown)


sareluzi (commented):
Re-posted as inline comments below.

ejgoldik (dismissed):

kernel/nvidia/5.0.2/0015-Add-max9295-GPIO-tunneling-for-the-external-sync-support.patch
```diff
@@ -102,7 +102,38 @@ index 4b03b1ef8..d7226af6e 100644
...
++		readback, (readback == MAX9295_RESET_SRC) ? "OK" : "MISMATCH");
++
++	max9295_write_reg(dev, MAX9295_2C0_ADDR, 0x00);
++	max9295_write_reg(dev, MAX9295_2C1_ADDR, 0x00);
++	max9295_write_reg(dev, MAX9295_2C2_ADDR, 0x00);
++	max9295_write_reg(dev, MAX9295_2C3_ADDR, 0x00);
++
++	mutex_unlock(&priv->lock);
++
++	return err;
```
sareluzi, line 106: **Always returns 0 — error path is dead code.**

`int err = 0;` (line 110) is never reassigned. `max9295_write_reg()` and `regmap_read()` returns are dropped on the floor; `MISMATCH` is logged but not propagated. As a result, `ds5_set_ser_esync_tunneling()` always sees `ret == 0` and the `if (ret) dev_warn(...)` path is unreachable. Same issue in `max9295_enable_gpio_tunneling()` above.

Suggest either checking `max9295_write_reg()` returns and bailing on the first failure, or — since you already do readback — comparing `readback` to the expected value and `return -EIO;` on mismatch.

kernel/realsense/d4xx.c
```diff
@@ -2416,6 +2416,42 @@ static void ds5_reset_streaming_flags(struct ds5_dev *ds5_dev)
...
+	if (!state || !state->ser_dev)
+		return -EINVAL;
+	if (state->dser_ops != &max96712_interface)
+		return 0;
+
+	dev_info(&state->client->dev,
+		"%s(): serializer ESYNC %s requested\n",
+		__func__, enable ? "enable" : "disable");
+
+	mutex_lock(&serdes_lock__);
```
sareluzi: **Asymmetric `serdes_lock__` acquisition.**

Per CLAUDE.md, `serdes_lock__` covers "scanning or assigning global topology slots." ESYNC enable/disable on an already-bound serializer touches neither, and the pre-PR probe-time call did not hold this lock. `max9295_*_gpio_tunneling()` already serializes via `priv->lock` internally, so this acquisition introduces a new nesting (`serdes_lock__` → `priv->lock`) without an obvious invariant being protected — and increases deadlock surface if any other path takes them in reverse order.

Suggest dropping the `serdes_lock__` acquisition unless there is a specific concurrent path it guards (in which case please add a one-line comment naming it).

kernel/realsense/d4xx.c
```diff
@@ -2623,6 +2659,14 @@ static int ds5_hw_reset_with_recovery(struct ds5 *state)
 		(state->fw_version >> 8) & 0xff, state->fw_version & 0xff,
 		(state->fw_build >> 8) & 0xff, state->fw_build & 0xff);
 
+	dev_info(&state->client->dev,
+		"%s(): disabling serializer ESYNC after HW reset\n", __func__);
+	ret = ds5_set_ser_esync_tunneling(state, false);
```
sareluzi: **V4L2 `sync_mode` control and hardware diverge after this call.**

ESYNC is forcibly disabled here, but `state->ctrls.sync_mode->cur.val` retains its prior value (e.g. `2` = Slave). After the reset, userspace queries the control, sees Slave, but the SerDes is no longer tunneling — and there is no re-application path on the next stream-start.

Two reasonable options:
- Read `sync_mode->cur.val` after reset and call `ds5_set_ser_esync_tunneling()` accordingly to preserve user intent.
- Or zero the V4L2 control via `__v4l2_ctrl_s_ctrl()` so the cached state matches reality.

Either way, "always disable, never tell userspace" leaves a stale control.

kernel/realsense/d4xx.c
```diff
@@ -2416,6 +2416,42 @@ static void ds5_reset_streaming_flags(struct ds5_dev *ds5_dev)
...
+{
+#ifdef CONFIG_VIDEO_D4XX_SERDES
+	int ret;
+
+	if (!state || !state->ser_dev)
+		return -EINVAL;
+	if (state->dser_ops != &max96712_interface)
+		return 0;
+
+	dev_info(&state->client->dev,
```
sareluzi: **Excessive `dev_info` logging on every sync_mode write.**

This dev_info plus the "ESYNC %s OK" dev_info further down (line ~2445), plus the dev_info chain inside `max9295_*_gpio_tunneling()` (~6 readback lines for the 5.x variant), produces ~10 INFO log lines per `VIDIOC_S_CTRL(SYNC_MODE)`. This conflicts with the dmesg-quiet direction the project has been moving toward (cf. the `ds5_read_poll()` note in CLAUDE.md).

Suggest demoting routine success logs to `dev_dbg`, keeping `dev_info` only for the high-level enable/disable outcome and for any error / MISMATCH case.

nvidia-oot/6.0/0004-Add-max9295-GPIO-tunneling-for-the-external-sync-support.patch
```diff
@@ -67,7 +67,27 @@ index e7c77845..eff47e1d 100644
 +
 +	return err;
 +}
-+EXPORT_SYMBOL(max9295_setup_gpio_tunneling);
++EXPORT_SYMBOL(max9295_enable_gpio_tunneling);
++
++int max9295_disable_gpio_tunneling(struct device *dev)
++{
++	struct max9295 *priv = dev_get_drvdata(dev);
++	int err = 0;
```
sareluzi, line 92: **5.x and 6.x patches diverge in detail.**

This 6.x `max9295_disable_gpio_tunneling()` does plain writes only (no readback / `OK`/`MISMATCH` logging), but the 5.0.2 / 5.1.2 patches add full readback verification logging for the same two register writes (~15 extra lines). The same divergence existed pre-PR for `enable`, but the new `disable` doubles it.

If the readback verification has value, it should be in both flavors; if it's redundant, it should be in neither. Worth picking one and consolidating.

kernel/realsense/d4xx.c
```diff
@@ -2928,6 +2972,15 @@ static int ds5_s_ctrl(struct v4l2_ctrl *ctrl)
 			ret = ds5_write(state, base | DS5_CAMERA_SYNC_MODE, ctrl->val);
 			dev_info(&state->client->dev, "%s(): SYNC_MODE command passed to FW, addr: 0x%x, value: %d, ret: %d\n",
 				__func__, base | DS5_CAMERA_SYNC_MODE, ctrl->val, ret);
+			if (!ret) {
+				bool need_esync = (ctrl->val == 2 || ctrl->val == 3);
```
sareluzi: **Magic numbers `2` / `3` for sync_mode values.**

These literals reproduce values from `sync_mode_menu_full[]` at d4xx.c:3435. Sync-mode numeric values now appear in `s_ctrl`, `ds5_adjust_sync_mode_control`, the menu arrays, and here — easy to drift.

Suggest defining named constants:

```c
#define DS5_SYNC_MODE_DEFAULT       0
#define DS5_SYNC_MODE_MASTER        1
#define DS5_SYNC_MODE_SLAVE         2
#define DS5_SYNC_MODE_FULL_SLAVE    3
#define DS5_SYNC_MODE_SUB_PREMASTER 4
#define DS5_SYNC_MODE_FULL_MASTER   5
```

…and use them everywhere these literals appear. Makes the predicate self-documenting and prevents the kind of drift that produced the slave/master mix-up in this PR's description ("slave (2) or master (3)" — value 3 is `Full Slave`).

kernel/nvidia/5.0.2/0015-Add-max9295-GPIO-tunneling-for-the-external-sync-support.patch
```diff
@@ -102,7 +102,38 @@ index 4b03b1ef8..d7226af6e 100644
...
++
++	mutex_lock(&priv->lock);
++
++	max9295_write_reg(dev, MAX9295_SRC_PWDN_ADDR, MAX9295_PWDN_GPIO);
++	regmap_read(priv->regmap, MAX9295_SRC_PWDN_ADDR, &readback);
++	dev_info(dev, "%s: reg 0x%04x written 0x%02x readback 0x%02x %s\n",
++		__func__, MAX9295_SRC_PWDN_ADDR, MAX9295_PWDN_GPIO,
++		readback, (readback == MAX9295_PWDN_GPIO) ? "OK" : "MISMATCH");
++
++	max9295_write_reg(dev, MAX9295_SRC_CTRL_ADDR, MAX9295_RESET_SRC);
```
sareluzi: **`MAX9295_RESET_SRC` is overloaded as the "disable" value.**

`MAX9295_RESET_SRC` (here) and `MAX9295_PWDN_GPIO` (line above) are written as the disable values for the same registers the enable function writes `MAX9295_RESET_ESYNC` / `MAX9295_PWDN_ESYNC_EXT` to. The constant names imply "default reset state of the SRC GPIO," not "ESYNC tunneling disable" — a future reader has to reverse this from context.

A one-line comment ("restore default SRC pin function — plain GPIO, no ESYNC tunneling") would help.

nvidia-oot/6.0/0004-Add-max9295-GPIO-tunneling-for-the-external-sync-support.patch
```diff
@@ -6,7 +6,7 @@ Subject: [PATCH] Add max9295 GPIO tunneling for the external sync support
 Add MAX9295A serializer GPIO tunneling / ESYNC register configuration
 for MAX96712A external sync support.
 
-Includes max9295_setup_gpio_tunneling() export for MAX96712A usage.
+Includes max9295_enable_gpio_tunneling() export for MAX96712A usage.
 
 ---
```
sareluzi, line 13: **Missing `Signed-off-by:` line.**

The 5.0.2 / 5.1.2 patches retain `ejgoldik`'s `Signed-off-by:`, but this 6.0 patch has none. Since this patch now contains new code (the `disable` function), please add a `Signed-off-by:` from the patch author.

kernel/nvidia/5.0.2/0015-Add-max9295-GPIO-tunneling-for-the-external-sync-support.patch
ipilchin1: done

kernel/realsense/d4xx.c
ipilchin1: done

kernel/realsense/d4xx.c
ipilchin1: done

kernel/realsense/d4xx.c
ipilchin1: done

nvidia-oot/6.0/0004-Add-max9295-GPIO-tunneling-for-the-external-sync-support.patch
ipilchin1: done

kernel/realsense/d4xx.c
ipilchin1: done

kernel/nvidia/5.0.2/0015-Add-max9295-GPIO-tunneling-for-the-external-sync-support.patch
ipilchin1: done

nvidia-oot/6.0/0004-Add-max9295-GPIO-tunneling-for-the-external-sync-support.patch
ipilchin1: done

kernel/nvidia/5.0.2/0015-Add-max9295-GPIO-tunneling-for-the-external-sync-support.patch
ipilchin1: done

sareluzi (approved):

PR #431, opened 2026-04-29
[RSDSO-21439][RSDSO-21440] - Align JP5 esync configuration

ejgoldik:


sareluzi (approved):
why not keep some prints as debug (instead of info)?

PR #432, opened 2026-05-04
[RSDSO-21407] - External sync using tsc PWM

ejgoldik:
This pull request adds comprehensive support for external multi-camera frame synchronization on the fg12-16ch board, specifically targeting Jetson platforms with D4XX cameras. The changes introduce device tree overlay updates and new documentation to enable and explain both Orin TSC-based and external signal generator-based synchronization. This includes configuration of TSC signal generators, pinmux settings, and clear instructions for users.

The most important changes are:
* Added TSC signal generator configuration and pinmux setup to FG overlay files 
* Configured pinmux for relevant pins: routed TSC-capable GPIOs to the TSC function where available, and set other pins (e.g., for MAX96712_3 and shared sync) as high-impedance inputs with pull-downs for external wiring.
* Patched cam_cdi_tsc.c - For which 6.1 patch dir diverged from 6.0 and is no longer a directory link


ashrafk93 (commented):
generated by rs-agentic

## Code Review — `[RSDSO-21407] External sync using tsc PWM`

This PR adds hardware-level external sync for multi-camera setups via Tegra TSC PWM signal generators. The DTS overlays are well-organized and the Python utility is clean. Below are observations.

---

### DTS Overlay Files (`.dts`)

1. **Fragment numbering gap / ordering**: In the 2-camera files (`cams-0-1`, `cams-0-4`), the new fragments are numbered `@2` and `@3`, which are _higher_ than the existing `@0` and `@1`. DT overlay fragment order is not guaranteed to be numeric, but many downstream tooling scripts assume ascending contiguous numbering starting at 0. Confirm that the Jetson DTBO toolchain handles non-contiguous fragment numbers without issue, or renumber so the new TSC fragments are `@0`/`@1` and the existing camera fragments are shifted.

2. **`fragment@3` appears in all files but references the same `tsc_sig_gen@c6a0000` target**: In the 2-camera variants, only one generator (`@380`) is configured inside the node, while the 4-camera variant (`cams-0-4-8-12`) configures three generators. This is correct, but a comment in the single/dual variants explaining _why_ only generator `@380` is used (i.e., only MAX1 is present) would help maintainers avoid accidentally duplicating to the wrong variant.

3. **`fragment@2` — pinmux `pinctrl-0` accumulation risk**: Each overlay appends `pinctrl-0 = <&tsc_sync_pinmux>` to the `pinmux@c300000` node. If multiple overlays are applied simultaneously (e.g., a user loads both `cams-0-1` and `cams-0-4`), this will conflict. Document that these overlays are mutually exclusive.

4. **Removed `CAM1_PWDN` / `CAM2_PWDN` / `CAM3_PWDN` `#define`s**: The commented-out defines from the previous version are removed without comment. If they were intentionally disabled (e.g., power sequencing is no longer handled via GPIO), add a brief comment to prevent future re-introduction.

---

### `nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch`

5. **Typo fix `MPF6` → `MFP6`**: The rename of `MAX96712_LINKA_MPF6_*` to `MAX96712_LINKA_MFP6_*` (Multi-Function Pin, not Multi-Pin Function) is a correct naming fix. The ESYNC constants are correspondingly unified to `MAX96712_LINKA_MFP_A/B_ESYNC` — this shared prefix for both MFP2 and MFP6 is fine, but a comment stating "same ESYNC values are written to both MFP2 and MFP6" would clarify intent.

6. **`max96712_set_sync` — no error propagation**: The function accumulates errors with `err |= ...` but the function signature returns `int`. A non-zero `err` on return should be treated as a failure by callers. Verify that `max96712_set_sync` callers check the return value and handle errors appropriately.

7. **New `max96712_write_reg` wrapper function added**: The diff shows a new static helper being added around line 201. Ensure the commit squashes the temporary "Add script to parse CSV" placeholder commit (unrelated title visible in the patch header).

---

### `scripts/ext_sync_gen_root.py`

8. **`/dev/mem` safety**: The script opens `/dev/mem` and maps `TSC_CONTROLLER_BASE` with `PROT_READ | PROT_WRITE`. A `CONFIG_STRICT_DEVMEM`-enabled kernel will refuse this. Document in the script header or README what kernel config is required, and consider adding an error message specific to that failure mode.

9. **`TSC_REG_SIZE = 0x10000` vs generator offsets**: The generator bases extend to `0xc6a0480 + 0x1C = 0xc6a049C`. Relative to `TSC_CONTROLLER_BASE = 0xc6a0000`, the furthest offset is `0x49C`, well within `0x10000`. This is correct.

10. **`START_OFFSET_MS` of 100 ms for synchronized start**: Using a shared future TSC timestamp for all generators is an elegant approach to hardware-synchronized startup. Consider logging the computed `start_ticks` so operators can verify synchronization in the field.

11. **`duty_cycle` bounds check**: The check `duty_cycle <= 0 or duty_cycle >= 100` rejects 0% and 100% duty cycles. Consider whether 0% (always-low output) should be allowed as a "stop" alternative.

---

### Minor / Nit

- Missing `__all__` or module docstring in `ext_sync_gen_root.py` — minor since it's a CLI script.
- The `MAX3_CSI_SYNC` and `MAX_ALL_CAM_SYNC` pins are correctly documented as high-Z inputs with pull-down; the comments in `cams-0-4-8-12` are especially clear.

Overall the PR is in good shape. Addressing items 1, 3, 6, and 8 before merge is recommended.

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

sareluzi (commented):
_(superseded — see consolidated review)_

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.dts
```diff
@@ -24,6 +28,82 @@
...
+	};
+
+	/* Muxing MAX<1/2/4>_CSI_SYNC to TSC function
+	 * MAX3_CSI_SYNC (PCC.02) & MAX_ALL_CAM_SYNC (PCC.00) do not have
+	 * this routing in Orin, so we define them as high-Z input with pull-down for external wiring
+	 * use jumper cable from MAX<1/2/4>_CSI_SYNC to MAX3_CSI_SYNC to use Orin TSC for sync signal or
+	 * MAX_ALL_CAM_SYNC connected to an external signal generator
+	 */
+	fragment@2 {
+		target-path = "/bus@0/pinmux@c300000";
```
sareluzi, line 66: **C1 [BLOCKING]** Pinmux target path may be wrong — AON pads under MAIN pinmux node

`target-path = "/bus@0/pinmux@c300000"` — but every pin listed below (`soc_gpio50_pbb2`, `can0_stb_paa4`, `can0_err_paa7`, `spi2_mosi_pcc2`, `spi2_sck_pcc0`) is in the **AON** GPIO bank (PAA / PBB / PCC). On T234 the AON pinmux is a separate controller (typically `pinmux@c302000`); MAIN pinmux at `c300000` doesn't own these pads and won't recognize the names.

Symptom: `nvidia,pins` silently doesn't match anything and your bias-pull-down + tsc-function settings never reach the register — exactly the failure mode that explains "PCC.00 still shows 1.8 V even after the pinmux override".

Verify on a running unit with `sudo cat /sys/kernel/debug/pinctrl/*pinmux*/pinconf-pins | grep -A3 'pcc0\|pbb2'` before merging.

_Same issue in 0-1.dts, 0-1-2-3.dts, 0-4.dts._

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.dts
```diff
@@ -24,6 +28,82 @@
...
+		__overlay__ {
+			pinctrl-names = "default";
```
sareluzi, line 68: **C2 [BLOCKING]** `pinctrl-names`/`pinctrl-0` set on the pinmux controller node itself

Setting `pinctrl-names = "default"` and `pinctrl-0 = <&tsc_sync_pinmux>` on the pinmux@<addr> node *replaces* whatever the platform DT had on its own default state list. Since the platform default usually lists every pad on the SoC, this overlay risks unconfiguring the rest of the board on boot.

The Tegra pinctrl driver auto-applies child state nodes regardless of `pinctrl-0`, so just add the child node — drop the `pinctrl-names`/`pinctrl-0` lines entirely.

_Same issue in 0-1.dts, 0-1-2-3.dts, 0-4.dts._

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.dts
```diff
@@ -24,6 +28,82 @@
...
+					gpio_pinmux = <&gpio_aon MAX4_CSI_SYNC GPIO_ACTIVE_LOW>;
+				};
+			};
+		};
+	};
+
+	/* Muxing MAX<1/2/4>_CSI_SYNC to TSC function
+	 * MAX3_CSI_SYNC (PCC.02) & MAX_ALL_CAM_SYNC (PCC.00) do not have
+	 * this routing in Orin, so we define them as high-Z input with pull-down for external wiring
+	 * use jumper cable from MAX<1/2/4>_CSI_SYNC to MAX3_CSI_SYNC to use Orin TSC for sync signal or
```
sareluzi, line 62: **C5 [MEDIUM]** Documented hardware jumper hidden in a DT comment

A board-mod requirement (jumper wire from MAX<1/2/4>_CSI_SYNC to MAX3_CSI_SYNC) buried in a DTS comment will get missed. Move it to the FG12 README/wiki and link from the overlay description, or `dev_warn()` from the driver when only 3 of 4 deserializers report a working sync source.

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.dts
```diff
@@ -24,6 +28,82 @@
 	jetson-header-name = "Jetson AGX CSI Connector";
 	compatible = JETSON_COMPATIBLE;
 
+	/* TSC signal generators for MAX96712 MFP2 sync (one per deserializer). */
+	fragment@3 {
```
sareluzi, line 32: **C6 [MEDIUM]** Duplicated TSC + pinmux fragments across overlays

`fragment@3` (TSC generators) and `fragment@2` (pinmux) are ~70% identical across 0-1.dts, 0-1-2-3.dts, 0-4.dts, 0-4-8-12.dts. Factor them into a shared `tegra234-d4xx-fg12-syncgen.dtsi` and `#include` from each, parameterized by which generators/pins are active. Also closes the calib-variant gap in C3.

---

**C8 [NIT]** Fragment ordering reversed — fragments appear `@3, @2, @1, @0`. Cosmetic; other overlays in the tree go ascending. Not blocking.

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.dts
```diff
@@ -24,6 +28,82 @@
...
+		target-path = "/";
+		__overlay__ {
+			tsc_sig_gen@c6a0000 {
+				status = "okay";
+				generator@380 {
+					status = "okay";
+					freq_hz = <30>;
+					duty_cycle = <25>;
+					gpio_pinmux = <&gpio_aon MAX1_CSI_SYNC GPIO_ACTIVE_LOW>;
```
sareluzi, line 41: **C7 [LOW]** `&gpio_aon` phandle reference

Relies on the platform DT having a label `gpio_aon`. NVIDIA's stock T234 DT uses `tegra_aon_gpio` in some L4T releases. Sanity-check the resolved DTB or the overlay will fail to apply.

sareluzi (commented):
Review (numbered for reference). Inline comments below cover most items; cross-file or out-of-diff items are listed here.

_(Note: please ignore the 20 "superseded" reviews above — those were artifacts of an API tooling issue while I was bisecting which line numbers GitHub would accept for inline comments. All actual content is in this single review.)_

## Cross-file / out-of-diff

### C3 [HIGH] `.calib.dts` siblings not updated
The matching `.calib` overlays for the three modified DTS files were not updated:
- `tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4.calib.dts`
- `tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.calib.dts`
- `tegra234-camera-d4xx-overlay-fg12-16ch.calib.dts`

These are the same content as the `.dts` siblings with `embedded_metadata_height = "0"` and a different overlay name. Calibration runs that load the `.calib` overlay will not get the new TSC sync configuration. Either propagate the changes or factor the shared blocks into a `.dtsi` include (see C6).

## Things I'd want answered before approving

### Q1 — Pinmux verification (you said 'minimal testing, works')
"Works" likely confirms only the TSC→MFP2 path. The TSC kernel driver applies its own per-generator `gpio_pinmux` independently of `fragment@2`, so PBB.2 routing can succeed even if `fragment@2` is silently dropped (see C1). The `PCC.00` pull-down lives only in `fragment@2` and is not exercised by your sync test.

Quick disambiguator on a running unit:
```bash
sudo cat /sys/kernel/debug/pinctrl/*pinmux*/pinconf-pins | grep -A3 'pcc0\|pbb2'
```
- If `pcc0` shows `pull=down` → C1 unfounded, fragment@2 was applied.
- If `pcc0` still shows `pull=up` → C1 real, the test only worked because it didn't use that pin.

### Q2 — Regression on the existing 4-port board (not tested yet)
C4 (unconditional MFP2 + MFP6 writes) changes behavior for every existing FG12 deployment, including the 0-4-8-12 board already in v1.0.2.32. Please re-run streaming on that config and probe MFP2 of one deserializer to confirm no phantom edge from the new ESYNC writes.

If time-tight, gate the new MFP2 writes behind a DT property (`realsense,mfp2-sync;`) default-off — keeps the 0-4-8-12 path bit-identical to PR #430 and makes the change opt-in for the new boards.

## Things that look good
- **G1** `MPF` → `MFP` rename matches the MAX96712 datasheet spelling.
- **G2** Two parallel sync paths (MFP6 for 1-cam-per-deser, MFP2 for 4-cams-on-1-deser, PCC.00 for external generator) match the hardware reality.
- **G3** Inline DT comment is the right place to record the PCC.00/PCC.02 no-TSC-routing constraint on Orin (with C5 caveat).
- **G4** `pull = TEGRA_PIN_PULL_DOWN` on PCC.00 is the right fix for the 1.8 V phantom voltage we chased in the parallel debug thread — assuming C1 is resolved.

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-1-2-3.dts
```diff
@@ -24,6 +23,51 @@
...
+					freq_hz = <30>;
+					duty_cycle = <25>;
+					gpio_pinmux = <&gpio_aon MAX1_CSI_SYNC GPIO_ACTIVE_LOW>;
+				};
+			};
+		};
+	};
+
+	/* Muxing MAX1_CSI_SYNC to TSC function
+	 * MAX_ALL_CAM_SYNC (PCC.00) does not have This routing in Orin
```
sareluzi: **C10 [NIT]** Comment capitalization / wording

`does not have This routing in Orin` → `does not have this routing on Orin`. Same comment in 0-1.dts, 0-4.dts, 0-4-8-12.dts.

nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch
```diff
@@ -619,12 +625,19 @@ index 8e237eef..1a03343f 100644
...
 +
 +	/* Configure ESYNC registers for GPIO tunneling */
-+	err |= max96712_write_reg(priv, MAX96712_LINKA_MPF6_A_ADDR, MAX96712_LINKA_MPF6_A_ESYNC);
-+
-+	err |= max96712_write_reg(priv, MAX96712_LINKA_MPF6_B_ADDR, MAX96712_LINKA_MPF6_B_ESYNC);
-+	err |= max96712_write_reg(priv, MAX96712_LINKB_MPF6_B_ADDR, MAX96712_LINKB_MPF6_B_ESYNC);
-+	err |= max96712_write_reg(priv, MAX96712_LINKC_MPF6_B_ADDR, MAX96712_LINKC_MPF6_B_ESYNC);
-+	err |= max96712_write_reg(priv, MAX96712_LINKD_MPF6_B_ADDR, MAX96712_LINKD_MPF6_B_ESYNC);
++	/* MFP6 configuration for multi-deserializer sync */
++	err |= max96712_write_reg(priv, MAX96712_LINKA_MFP6_A_ADDR, MAX96712_LINKA_MFP_A_ESYNC);
```
sareluzi, line 629: **C4 [HIGH]** ESYNC unconditionally programmed on both MFP2 and MFP6

Every deserializer is configured with both ESYNC paths regardless of which pin is actually wired on a given board. On a board where only MFP2 is wired and MFP6 is floating you may pick up noise on the unused pin and inject phantom syncs (or vice-versa).

Suggestion: drive this from a DT property on the deserializer node (e.g. `realsense,sync-pin = "mfp2"|"mfp6"|"both"`) and only program the requested pin. See Q2 in the review summary — without this, the existing 0-4-8-12 board behavior changes and needs full regression.

nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch
```diff
@@ -1,39 +1,45 @@
...
++#define MAX96712_LINKC_MFP2_B_ADDR 0x374
++#define MAX96712_LINKD_MFP2_B_ADDR 0x3AA
 +
 +/* ESYNC data defines */
-+#define MAX96712_LINKA_MPF6_A_ESYNC 0x2B
-+#define MAX96712_LINKA_MPF6_B_ESYNC 0x17
-+#define MAX96712_LINKB_MPF6_B_ESYNC 0x77
-+#define MAX96712_LINKC_MPF6_B_ESYNC 0x77
-+#define MAX96712_LINKD_MPF6_B_ESYNC 0x77
++#define MAX96712_LINKA_MFP_A_ESYNC 0x2B
```
sareluzi, line 38: **C11 [MEDIUM]** Macro rename `*_MPF6_*_ESYNC` → `*_MFP_*_ESYNC` (no pin number) needs a justifying comment

Dropping the pin number from the macro name implies the same ESYNC payload is correct for both MFP2 and MFP6, and you indeed write the same value to both. That's only true if the MAX96712 GPIO routing matrix is symmetric for those two pins. Add a one-line comment citing the MAX96712 datasheet section that says ESYNC payload is pin-agnostic, otherwise this looks like a copy-paste that happens to work.

nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch
```diff
@@ -619,12 +625,19 @@ index 8e237eef..1a03343f 100644
...
++	err |= max96712_write_reg(priv, MAX96712_LINKB_MFP6_B_ADDR, MAX96712_LINKB_MFP_B_ESYNC);
++	err |= max96712_write_reg(priv, MAX96712_LINKC_MFP6_B_ADDR, MAX96712_LINKC_MFP_B_ESYNC);
++	err |= max96712_write_reg(priv, MAX96712_LINKD_MFP6_B_ADDR, MAX96712_LINKD_MFP_B_ESYNC);
++
++	/* MFP2 configuration for single-deserializer sync */
++	err |= max96712_write_reg(priv, MAX96712_LINKA_MFP2_A_ADDR, MAX96712_LINKA_MFP_A_ESYNC);
++	err |= max96712_write_reg(priv, MAX96712_LINKA_MFP2_B_ADDR, MAX96712_LINKA_MFP_B_ESYNC);
++	err |= max96712_write_reg(priv, MAX96712_LINKB_MFP2_B_ADDR, MAX96712_LINKB_MFP_B_ESYNC);
++	err |= max96712_write_reg(priv, MAX96712_LINKC_MFP2_B_ADDR, MAX96712_LINKC_MFP_B_ESYNC);
++	err |= max96712_write_reg(priv, MAX96712_LINKD_MFP2_B_ADDR, MAX96712_LINKD_MFP_B_ESYNC);
```
sareluzi, line 640: **C12 [LOW]** Stale TODOs in committed code (refers to lines ~650, ~657, ~664 below — GitHub truncates inline comment ability past line ~642)

Three `/* TODO EHUD: Is needed? */` blocks in `max96712_reset_control`, `max96712_sdev_register`, `max96712_sdev_unregister`. Not introduced by this PR but they ride along — drop the TODO and document the no-op intent, or open a follow-up ticket and reference its ID in the comment.

scripts/ext_sync_gen_root.py
```diff
@@ -0,0 +1,246 @@
...
+TSC_NS_PER_TICK       = 32
+NS_PER_MS             = 1000000
+START_OFFSET_MS       = 100
+MAX_FREQ_HZ           = 120
+
+
+class TSCController:
+    def __init__(self):
+        self.fd = os.open("/dev/mem", os.O_RDWR | os.O_SYNC)
+        self.mm = mmap.mmap(self.fd, TSC_REG_SIZE,
```
sareluzi: **C13 [HIGH]** Userspace `/dev/mem` poking duplicates what the kernel TSC driver should do

The DT overlay enables `tsc_sig_gen@c6a0000` with `freq_hz = <30>; duty_cycle = <25>;` and `gpio_pinmux = <&gpio_aon ...>;` per generator. NVIDIA's in-kernel `tegra-tsc-signal-generator` driver consumes those at probe and starts the generators on its own. If the kernel driver runs, this script *races* it (and could re-enable it after the kernel disabled it). If the kernel driver isn't loaded, the DT properties are dead.

Pick one and stick with it:
- **Kernel-managed**: drop the script (or keep it as a debug-only tool with a big warning) and rely on DT.
- **Userspace-managed**: in DT, mark the generators `status = "disabled"` or omit `freq_hz`/`duty_cycle` so the kernel leaves them alone, and document `ext_sync_gen_root.py --enable` as a required runtime step.

---

**C14 [MEDIUM]** No mutual exclusion / lock with the kernel TSC driver — this `mmap` succeeds even if the kernel driver has the device live. Refuse to run if `/sys/bus/platform/drivers/tegra-tsc-sig-gen/c6a0000.tsc_sig_gen` (or the actual bind path) shows the device bound — or use the kernel driver's sysfs/debugfs interface instead.

scripts/ext_sync_gen_root.py
```diff
@@ -0,0 +1,246 @@
...
+TSC_GENX_EDGE1  = 0x1C
+
+# Bit definitions
+TSC_GENX_CTRL_ENABLE      = (1 << 0)
+TSC_GENX_CTRL_INITIAL_VAL = (1 << 1)
+
+TSC_GENX_STATUS_WAITING = (1 << 0)
+TSC_GENX_STATUS_RUNNING = (1 << 1)
+
+TSC_GENX_EDGEX_TOGGLE = (1 << 29)
```
sareluzi: **C15 [LOW]** Hard-coded register layout duplicated in Python

Register offsets and bit defs (`TSC_GENX_EDGEX_TOGGLE`, `TSC_GENX_CTRL_INITIAL_VAL`, etc.) are duplicated from kernel headers. Will silently break if NVIDIA changes the IP-block layout in a future L4T. Add a comment pointing at the kernel source/manual + TRM section reference.

scripts/ext_sync_gen_root.py
```diff
@@ -0,0 +1,246 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: GPL-2.0-only
+# SPDX-FileCopyrightText: Copyright (c) 2024, INTEL CORPORATION.  All rights reserved.
```
sareluzi: **C16 [LOW]** SPDX header copyright

Should be 2026, and likely "RealSense Technologies" — match whatever convention the rest of the repo is using.

scripts/ext_sync_gen_root.py
```diff
@@ -0,0 +1,246 @@
...
+#
+# tsc_sync.py - Control TSC signal generators on Tegra234 for camera sync.
```
sareluzi: **C17 [LOW]** Module-level docstring labels file as `tsc_sync.py`

Filename is `ext_sync_gen_root.py`. Cosmetic.

scripts/ext_sync_gen_root.py
```diff
@@ -0,0 +1,246 @@
...
+
+TSC_GENX_EDGEX_TOGGLE = (1 << 29)
+TSC_GENX_EDGEX_LOOP   = (1 << 28)
+TSC_GENX_EDGEX_OFFSET_MASK = 0x0FFFFFFF
+
+TSC_TICKS_PER_HZ     = 31250000
+TSC_NS_PER_TICK       = 32
+NS_PER_MS             = 1000000
+START_OFFSET_MS       = 100
+MAX_FREQ_HZ           = 120
```
sareluzi: **C18 [NIT]** `MAX_FREQ_HZ = 120` — magic constant

Why 120? Document the upper bound (CSI hsync rate? deserializer ESYNC max? framerate ceiling for D457?) so future-you can decide whether to raise it.

scripts/ext_sync_gen_root.py
```diff
@@ -0,0 +1,246 @@
...
+                       help="Enable TSC signal output")
+    group.add_argument("--disable", action="store_true",
+                       help="Disable TSC signal output")
+    group.add_argument("--status", action="store_true",
+                       help="Print generator status")
+    parser.add_argument("--fps", type=int, default=30,
+                        help="Signal frequency in Hz (default: 30)")
+    parser.add_argument("--duty", type=int, default=25,
+                        help="Duty cycle in percent (default: 25)")
+    parser.add_argument("--gen", type=int, nargs="+", metavar="N",
```
sareluzi: **C19 [NIT]** `--gen` doesn't echo which generators were selected

When defaulting to all, a `print(f"Targeting generators {gens or 'ALL'}")` at the top of the action would aid debugging.

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.dts
ejgoldik: c300000 is the AON pinmux.
2430000 is the Main
c302000 does not exist

sudo cat /sys/kernel/debug/pinctrl/c300000.pinmux/pinconf-pins
Pin config settings per pin
Format: pin (name): configs
pin 0 (CAN0_DOUT_PAA0):
pin 1 (CAN0_DIN_PAA1):
pin 2 (CAN1_DOUT_PAA2):
pin 3 (CAN1_DIN_PAA3):
pin 4 (CAN0_STB_PAA4):
pin 5 (CAN0_EN_PAA5):
pin 6 (SOC_GPIO49_PAA6):
pin 7 (CAN0_ERR_PAA7):
pin 8 (CAN1_STB_PBB0):
pin 9 (CAN1_EN_PBB1):
pin 10 (SOC_GPIO50_PBB2):
pin 11 (CAN1_ERR_PBB3):
pin 12 (SPI2_SCK_PCC0):
pin 13 (SPI2_MISO_PCC1):
pin 14 (SPI2_MOSI_PCC2):
pin 15 (SPI2_CS0_PCC3):
pin 16 (TOUCH_CLK_PCC4):
pin 17 (UART3_TX_PCC5):
pin 18 (UART3_RX_PCC6):
pin 19 (GEN2_I2C_SCL_PCC7):
pin 20 (GEN2_I2C_SDA_PDD0):
pin 21 (GEN8_I2C_SCL_PDD1):
pin 22 (GEN8_I2C_SDA_PDD2):
pin 23 (SCE_ERROR_PEE0):
pin 24 (VCOMP_ALERT_PEE1):
pin 25 (AO_RETENTION_N_PEE2):
pin 26 (BATT_OC_PEE3):
pin 27 (POWER_ON_PEE4):
pin 28 (SOC_GPIO26_PEE5):
pin 29 (SOC_GPIO27_PEE6):
pin 30 (BOOTV_CTL_N_PEE7):
pin 31 (HDMI_CEC_PGG0):

scripts/ext_sync_gen.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: GPL-2.0-only
+# SPDX-FileCopyrightText: Copyright (c) 2024, INTEL CORPORATION.  All rights reserved.
```
sareluzi: should use RealSense copyright

nvidia-oot/6.1/0001-Porting-nvidia-driver-patches-to-jetpack-6.0.patch
```diff
@@ -0,0 +1 @@
+../6.0/0001-Porting-nvidia-driver-patches-to-jetpack-6.0.patch
```
sareluzi, line 1: how is this related to the change?

ejgoldik:
> generated by rs-agentic
> 
> ## Code Review — `[RSDSO-21407] External sync using tsc PWM`
> This PR adds hardware-level external sync for multi-camera setups via Tegra TSC PWM signal generators. The DTS overlays are well-organized and the Python utility is clean. Below are observations.
> 
> ### DTS Overlay Files (`.dts`)
> 1. **Fragment numbering gap / ordering**: In the 2-camera files (`cams-0-1`, `cams-0-4`), the new fragments are numbered `@2` and `@3`, which are _higher_ than the existing `@0` and `@1`. DT overlay fragment order is not guaranteed to be numeric, but many downstream tooling scripts assume ascending contiguous numbering starting at 0. Confirm that the Jetson DTBO toolchain handles non-contiguous fragment numbers without issue, or renumber so the new TSC fragments are `@0`/`@1` and the existing camera fragments are shifted.
> 2. **`fragment@3` appears in all files but references the same `tsc_sig_gen@c6a0000` target**: In the 2-camera variants, only one generator (`@380`) is configured inside the node, while the 4-camera variant (`cams-0-4-8-12`) configures three generators. This is correct, but a comment in the single/dual variants explaining _why_ only generator `@380` is used (i.e., only MAX1 is present) would help maintainers avoid accidentally duplicating to the wrong variant.
> 3. **`fragment@2` — pinmux `pinctrl-0` accumulation risk**: Each overlay appends `pinctrl-0 = <&tsc_sync_pinmux>` to the `pinmux@c300000` node. If multiple overlays are applied simultaneously (e.g., a user loads both `cams-0-1` and `cams-0-4`), this will conflict. Document that these overlays are mutually exclusive.
> 4. **Removed `CAM1_PWDN` / `CAM2_PWDN` / `CAM3_PWDN` `#define`s**: The commented-out defines from the previous version are removed without comment. If they were intentionally disabled (e.g., power sequencing is no longer handled via GPIO), add a brief comment to prevent future re-introduction.
> 
> ### `nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch`
> 5. **Typo fix `MPF6` → `MFP6`**: The rename of `MAX96712_LINKA_MPF6_*` to `MAX96712_LINKA_MFP6_*` (Multi-Function Pin, not Multi-Pin Function) is a correct naming fix. The ESYNC constants are correspondingly unified to `MAX96712_LINKA_MFP_A/B_ESYNC` — this shared prefix for both MFP2 and MFP6 is fine, but a comment stating "same ESYNC values are written to both MFP2 and MFP6" would clarify intent.
> 6. **`max96712_set_sync` — no error propagation**: The function accumulates errors with `err |= ...` but the function signature returns `int`. A non-zero `err` on return should be treated as a failure by callers. Verify that `max96712_set_sync` callers check the return value and handle errors appropriately.
> 7. **New `max96712_write_reg` wrapper function added**: The diff shows a new static helper being added around line 201. Ensure the commit squashes the temporary "Add script to parse CSV" placeholder commit (unrelated title visible in the patch header).
> 
> ### `scripts/ext_sync_gen_root.py`
> 8. **`/dev/mem` safety**: The script opens `/dev/mem` and maps `TSC_CONTROLLER_BASE` with `PROT_READ | PROT_WRITE`. A `CONFIG_STRICT_DEVMEM`-enabled kernel will refuse this. Document in the script header or README what kernel config is required, and consider adding an error message specific to that failure mode.
> 9. **`TSC_REG_SIZE = 0x10000` vs generator offsets**: The generator bases extend to `0xc6a0480 + 0x1C = 0xc6a049C`. Relative to `TSC_CONTROLLER_BASE = 0xc6a0000`, the furthest offset is `0x49C`, well within `0x10000`. This is correct.
> 10. **`START_OFFSET_MS` of 100 ms for synchronized start**: Using a shared future TSC timestamp for all generators is an elegant approach to hardware-synchronized startup. Consider logging the computed `start_ticks` so operators can verify synchronization in the field.
> 11. **`duty_cycle` bounds check**: The check `duty_cycle <= 0 or duty_cycle >= 100` rejects 0% and 100% duty cycles. Consider whether 0% (always-low output) should be allowed as a "stop" alternative.
> 
> ### Minor / Nit
> * Missing `__all__` or module docstring in `ext_sync_gen_root.py` — minor since it's a CLI script.
> * The `MAX3_CSI_SYNC` and `MAX_ALL_CAM_SYNC` pins are correctly documented as high-Z inputs with pull-down; the comments in `cams-0-4-8-12` are especially clear.
> 
> Overall the PR is in good shape. Addressing items 1, 3, 6, and 8 before merge is recommended.



> generated by rs-agentic
> 
> ## Code Review — `[RSDSO-21407] External sync using tsc PWM`
> This PR adds hardware-level external sync for multi-camera setups via Tegra TSC PWM signal generators. The DTS overlays are well-organized and the Python utility is clean. Below are observations.
> 
> ### DTS Overlay Files (`.dts`)
> 1. **Fragment numbering gap / ordering**: In the 2-camera files (`cams-0-1`, `cams-0-4`), the new fragments are numbered `@2` and `@3`, which are _higher_ than the existing `@0` and `@1`. DT overlay fragment order is not guaranteed to be numeric, but many downstream tooling scripts assume ascending contiguous numbering starting at 0. Confirm that the Jetson DTBO toolchain handles non-contiguous fragment numbers without issue, or renumber so the new TSC fragments are `@0`/`@1` and the existing camera fragments are shifted.
> 2. **`fragment@3` appears in all files but references the same `tsc_sig_gen@c6a0000` target**: In the 2-camera variants, only one generator (`@380`) is configured inside the node, while the 4-camera variant (`cams-0-4-8-12`) configures three generators. This is correct, but a comment in the single/dual variants explaining _why_ only generator `@380` is used (i.e., only MAX1 is present) would help maintainers avoid accidentally duplicating to the wrong variant.
> 3. **`fragment@2` — pinmux `pinctrl-0` accumulation risk**: Each overlay appends `pinctrl-0 = <&tsc_sync_pinmux>` to the `pinmux@c300000` node. If multiple overlays are applied simultaneously (e.g., a user loads both `cams-0-1` and `cams-0-4`), this will conflict. Document that these overlays are mutually exclusive.
> 4. **Removed `CAM1_PWDN` / `CAM2_PWDN` / `CAM3_PWDN` `#define`s**: The commented-out defines from the previous version are removed without comment. If they were intentionally disabled (e.g., power sequencing is no longer handled via GPIO), add a brief comment to prevent future re-introduction.
> 
> ### `nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch`
> 5. **Typo fix `MPF6` → `MFP6`**: The rename of `MAX96712_LINKA_MPF6_*` to `MAX96712_LINKA_MFP6_*` (Multi-Function Pin, not Multi-Pin Function) is a correct naming fix. The ESYNC constants are correspondingly unified to `MAX96712_LINKA_MFP_A/B_ESYNC` — this shared prefix for both MFP2 and MFP6 is fine, but a comment stating "same ESYNC values are written to both MFP2 and MFP6" would clarify intent.
> 6. **`max96712_set_sync` — no error propagation**: The function accumulates errors with `err |= ...` but the function signature returns `int`. A non-zero `err` on return should be treated as a failure by callers. Verify that `max96712_set_sync` callers check the return value and handle errors appropriately.
> 7. **New `max96712_write_reg` wrapper function added**: The diff shows a new static helper being added around line 201. Ensure the commit squashes the temporary "Add script to parse CSV" placeholder commit (unrelated title visible in the patch header).
> 
> ### `scripts/ext_sync_gen_root.py`
> 8. **`/dev/mem` safety**: The script opens `/dev/mem` and maps `TSC_CONTROLLER_BASE` with `PROT_READ | PROT_WRITE`. A `CONFIG_STRICT_DEVMEM`-enabled kernel will refuse this. Document in the script header or README what kernel config is required, and consider adding an error message specific to that failure mode.
> 9. **`TSC_REG_SIZE = 0x10000` vs generator offsets**: The generator bases extend to `0xc6a0480 + 0x1C = 0xc6a049C`. Relative to `TSC_CONTROLLER_BASE = 0xc6a0000`, the furthest offset is `0x49C`, well within `0x10000`. This is correct.
> 10. **`START_OFFSET_MS` of 100 ms for synchronized start**: Using a shared future TSC timestamp for all generators is an elegant approach to hardware-synchronized startup. Consider logging the computed `start_ticks` so operators can verify synchronization in the field.
> 11. **`duty_cycle` bounds check**: The check `duty_cycle <= 0 or duty_cycle >= 100` rejects 0% and 100% duty cycles. Consider whether 0% (always-low output) should be allowed as a "stop" alternative.
> 
> ### Minor / Nit
> * Missing `__all__` or module docstring in `ext_sync_gen_root.py` — minor since it's a CLI script.
> * The `MAX3_CSI_SYNC` and `MAX_ALL_CAM_SYNC` pins are correctly documented as high-Z inputs with pull-down; the comments in `cams-0-4-8-12` are especially clear.
> 
> Overall the PR is in good shape. Addressing items 1, 3, 6, and 8 before merge is recommended.

1. Fragment numbering: This is the convention with all our overlays. Can be changed but it is not relevant to this PR.
2. Not necessary to explain why we only include what we include in every overlay. 
3. Fixed
4. Comments removed as they were mistakenly there (incomplete cleanup when creating these overlays)
5. Unnecessary clarification as user can see it pertains to both
6. Does not exist
7. Not relevant to the patch
8 - 11. Irrelevant, script removed

ejgoldik:
> Review (numbered for reference). Inline comments below cover most items; cross-file or out-of-diff items are listed here.
> 
> _(Note: please ignore the 20 "superseded" reviews above — those were artifacts of an API tooling issue while I was bisecting which line numbers GitHub would accept for inline comments. All actual content is in this single review.)_
> 
> ## Cross-file / out-of-diff
> ### C3 [HIGH] `.calib.dts` siblings not updated
> The matching `.calib` overlays for the three modified DTS files were not updated:
> 
> * `tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4.calib.dts`
> * `tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.calib.dts`
> * `tegra234-camera-d4xx-overlay-fg12-16ch.calib.dts`
> 
> These are the same content as the `.dts` siblings with `embedded_metadata_height = "0"` and a different overlay name. Calibration runs that load the `.calib` overlay will not get the new TSC sync configuration. Either propagate the changes or factor the shared blocks into a `.dtsi` include (see C6).
> 
> ## Things I'd want answered before approving
> ### Q1 — Pinmux verification (you said 'minimal testing, works')
> "Works" likely confirms only the TSC→MFP2 path. The TSC kernel driver applies its own per-generator `gpio_pinmux` independently of `fragment@2`, so PBB.2 routing can succeed even if `fragment@2` is silently dropped (see C1). The `PCC.00` pull-down lives only in `fragment@2` and is not exercised by your sync test.
> 
> Quick disambiguator on a running unit:
> 
> ```shell
> sudo cat /sys/kernel/debug/pinctrl/*pinmux*/pinconf-pins | grep -A3 'pcc0\|pbb2'
> ```
> 
> * If `pcc0` shows `pull=down` → C1 unfounded, fragment@2 was applied.
> * If `pcc0` still shows `pull=up` → C1 real, the test only worked because it didn't use that pin.
> 
> ### Q2 — Regression on the existing 4-port board (not tested yet)
> C4 (unconditional MFP2 + MFP6 writes) changes behavior for every existing FG12 deployment, including the 0-4-8-12 board already in v1.0.2.32. Please re-run streaming on that config and probe MFP2 of one deserializer to confirm no phantom edge from the new ESYNC writes.
> 
> If time-tight, gate the new MFP2 writes behind a DT property (`realsense,mfp2-sync;`) default-off — keeps the 0-4-8-12 path bit-identical to PR #430 and makes the change opt-in for the new boards.
> 
> ## Things that look good
> * **G1** `MPF` → `MFP` rename matches the MAX96712 datasheet spelling.
> * **G2** Two parallel sync paths (MFP6 for 1-cam-per-deser, MFP2 for 4-cams-on-1-deser, PCC.00 for external generator) match the hardware reality.
> * **G3** Inline DT comment is the right place to record the PCC.00/PCC.02 no-TSC-routing constraint on Orin (with C5 caveat).
> * **G4** `pull = TEGRA_PIN_PULL_DOWN` on PCC.00 is the right fix for the 1.8 V phantom voltage we chased in the parallel debug thread — assuming C1 is resolved.



> Review (numbered for reference). Inline comments below cover most items; cross-file or out-of-diff items are listed here.
> 
> _(Note: please ignore the 20 "superseded" reviews above — those were artifacts of an API tooling issue while I was bisecting which line numbers GitHub would accept for inline comments. All actual content is in this single review.)_
> 
> ## Cross-file / out-of-diff
> ### C3 [HIGH] `.calib.dts` siblings not updated
> The matching `.calib` overlays for the three modified DTS files were not updated:
> 
> * `tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4.calib.dts`
> * `tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.calib.dts`
> * `tegra234-camera-d4xx-overlay-fg12-16ch.calib.dts`
> 
> These are the same content as the `.dts` siblings with `embedded_metadata_height = "0"` and a different overlay name. Calibration runs that load the `.calib` overlay will not get the new TSC sync configuration. Either propagate the changes or factor the shared blocks into a `.dtsi` include (see C6).
> 
> ## Things I'd want answered before approving
> ### Q1 — Pinmux verification (you said 'minimal testing, works')
> "Works" likely confirms only the TSC→MFP2 path. The TSC kernel driver applies its own per-generator `gpio_pinmux` independently of `fragment@2`, so PBB.2 routing can succeed even if `fragment@2` is silently dropped (see C1). The `PCC.00` pull-down lives only in `fragment@2` and is not exercised by your sync test.
> 
> Quick disambiguator on a running unit:
> 
> ```shell
> sudo cat /sys/kernel/debug/pinctrl/*pinmux*/pinconf-pins | grep -A3 'pcc0\|pbb2'
> ```
> 
> * If `pcc0` shows `pull=down` → C1 unfounded, fragment@2 was applied.
> * If `pcc0` still shows `pull=up` → C1 real, the test only worked because it didn't use that pin.
> 
> ### Q2 — Regression on the existing 4-port board (not tested yet)
> C4 (unconditional MFP2 + MFP6 writes) changes behavior for every existing FG12 deployment, including the 0-4-8-12 board already in v1.0.2.32. Please re-run streaming on that config and probe MFP2 of one deserializer to confirm no phantom edge from the new ESYNC writes.
> 
> If time-tight, gate the new MFP2 writes behind a DT property (`realsense,mfp2-sync;`) default-off — keeps the 0-4-8-12 path bit-identical to PR #430 and makes the change opt-in for the new boards.
> 
> ## Things that look good
> * **G1** `MPF` → `MFP` rename matches the MAX96712 datasheet spelling.
> * **G2** Two parallel sync paths (MFP6 for 1-cam-per-deser, MFP2 for 4-cams-on-1-deser, PCC.00 for external generator) match the hardware reality.
> * **G3** Inline DT comment is the right place to record the PCC.00/PCC.02 no-TSC-routing constraint on Orin (with C5 caveat).
> * **G4** `pull = TEGRA_PIN_PULL_DOWN` on PCC.00 is the right fix for the 1.8 V phantom voltage we chased in the parallel debug thread — assuming C1 is resolved.

Q1: Indeed it is not applied without pinctrl-0 definition. With pinctrl-0 definition they are applied together and probing MFP6 pin on the board shows that it is '0' (and not 1.8v) meaning fragment 2 was applied correctly

Q2: Tested both MFP2 and MFP6 path. I can see D457 failures (failed to start stream) when connected through MFP6 even when I disable MFP2 path completely (in the deserializer code) and I see the same failures in MFP2 path even when I disable MFP2 path (in the deserializer code) so no evidence of cross-talk.

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.dts
ejgoldik: Verified it now. Without pinctrl-names + pinctrl-0 the fragment is ignored. Restored these

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.dts
ejgoldik: Added readme

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.dts
ejgoldik: C6 - should be addressed as part of a general task to make our overlays modular
C8 - All our fragments in all our overlays are in descending order

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.dts
ejgoldik: C7 - Incorrect. tegra_aon_gpio does not exist. gpio_aon is used throughout the device tree.

nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch
ejgoldik: Testing shows no difference in behavior

nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch
ejgoldik: Better to have a follow-up to cleanup comments instead of touching regions of the code not related to this PR

scripts/ext_sync_gen.py
ejgoldik: None of our other scripts have copyright headers, and I'm not sure this even holds in an open-source repo. Will remove completely.
(Side note- our d4xx.c still has an Intel copyright header)

nvidia-oot/6.1/0001-Porting-nvidia-driver-patches-to-jetpack-6.0.patch
ejgoldik: I needed to add a patch to cdi_tsc driver which is different in JP6.1 to JP6.0. That means that our assumption that the patch dir for 6.1 can be a link to the patch dir of 6.0 no longer holds. So now 6.1 has its own dir but the internal files are links where applicable (all of them except the new one "0010-Runtime-tsc-rate-config.patch"

nvidia-oot/6.0/0006-Adding-max96712-support-for-D4xx.patch
ejgoldik: All MFP pins have the same structure, Do we need a comment for that?

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-1-2-3.dts
ejgoldik: Fixed

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4.dts
```diff
@@ -24,6 +25,61 @@
...
+		__overlay__ {
+			tsc_sync_pinmux: tsc-sync-pins {
+				max1-csi-sync {
+					nvidia,pins = "soc_gpio50_pbb2";
+					nvidia,function = "tsc";
+					nvidia,tristate = <TEGRA_PIN_DISABLE>;
+					nvidia,enable-input = <TEGRA_PIN_DISABLE>;
+				};
+				max2-csi-sync {
+					nvidia,pins = "can0_stb_paa7";
```
sareluzi: **N1 [HIGH — real bug]** Pinmux pin-name typo

```dts
max2-csi-sync {
    nvidia,pins = "can0_stb_paa7";   // ← invalid combination
```

On T234 the pad at PAA.7 is named `can0_err_paa7` (CAN0_ERR). `can0_stb` is PAA.4. When `MAX2_CSI_SYNC` was bumped from `AON_GPIO(AA, 4)` → `AON_GPIO(AA, 7)` in this file, the corresponding pinmux pin-name string wasn't updated.

Sister files have it correct: `0-4-8-12.dts:76` and `PWR-only.dts:74` both use `can0_err_paa7` for the same pad.

The pinctrl driver will silently fail to match `can0_stb_paa7` and leave PAA.7 in its POR/default state, so MAX2 sync on the cams-0-4 board will not get TSC routing.

Fix: `nvidia,pins = "can0_err_paa7";`

scripts/ext_sync_gen.py
```diff
@@ -0,0 +1,82 @@
...
+#
+# tsc_sync_ioctl.py - Control TSC signal generators via /dev/cdi_tsc.
```
sareluzi: **N4 [LOW]** Docstring filename mismatch

```python
# tsc_sync_ioctl.py - Control TSC signal generators via /dev/cdi_tsc.
```

Filename is `ext_sync_gen.py`. Same kind of stale-name issue as the previously-resolved C17 — just inherited a different name. Cosmetic.

sareluzi (commented):
Re-review pass on commit e731ff97. Resolved 7 threads (C2, C13, C15, C16, C17, C18, C19) — all addressed by the rewrite to ioctl-based control or the removal of the `pinctrl-0` override. Posting 6 new findings below (N1 is a real pin-name typo bug; N6 is a real ioctl error-path bug; the rest are LOW/cosmetic).

_Carry-over still open from prior review:_ C1, C4, C5, C6+C8, C7, C10, C11, C12 — see the existing inline threads.

nvidia-oot/6.0/0010-Runtime-tsc-rate-config.patch
```diff
@@ -0,0 +1,98 @@
...
+ 		}
+ 		break;
++	case _IOC_NR(CDI_TSC_SET_RATE): {
++		struct cdi_tsc_rate rate;
++		struct tsc_signal_generator *gen;
++
++		if (copy_from_user(&rate, ptr, sizeof(rate)) != 0U)
++			break;
++		if (rate.freq_hz == 0 ||
++		    rate.freq_hz > controller->features->rational_locking.max_freq_hz_lcm) {
```
sareluzi, line 72: **N5 [LOW]** `freq_hz` upper bound uses `rational_locking.max_freq_hz_lcm`

```c
if (rate.freq_hz == 0 ||
    rate.freq_hz > controller->features->rational_locking.max_freq_hz_lcm) {
```

`max_freq_hz_lcm` is NVIDIA's LCM bound used for phase-locking multiple generators in *rational locking* mode — it's not the absolute hardware max. For our external-sync use case (single generator from the kernel's POV, or independent generators) this may artificially cap `freq_hz` below what the hardware can actually drive.

Either use a more permissive bound (e.g. the per-generator max from the features table), or add a one-line comment justifying why the LCM bound is the right one here.

_Same code in 6.1/0010 and 7.0/0010._

nvidia-oot/6.0/0010-Runtime-tsc-rate-config.patch
```diff
@@ -0,0 +1,98 @@
...
+ 
+@@ -494,6 +495,34 @@ static long cdi_tsc_chardev_ioctl(struct file *file, unsigned int cmd, unsigned
+ 			err = cdi_tsc_stop_generators(controller);
+ 		}
+ 		break;
++	case _IOC_NR(CDI_TSC_SET_RATE): {
++		struct cdi_tsc_rate rate;
++		struct tsc_signal_generator *gen;
++
++		if (copy_from_user(&rate, ptr, sizeof(rate)) != 0U)
```
sareluzi, line 69: **N6 [LOW — real bug]** `copy_from_user` failure swallowed

```c
case _IOC_NR(CDI_TSC_SET_RATE): {
    struct cdi_tsc_rate rate;
    struct tsc_signal_generator *gen;

    if (copy_from_user(&rate, ptr, sizeof(rate)) != 0U)
        break;          // ← err is not set; ioctl returns 0 (success)
```

`err` is initialised to 0 at function entry, so on `copy_from_user` failure this case `break`s out and the ioctl returns success without having done anything. Userspace gets no signal that the rate write didn't take effect.

Fix: `err = -EFAULT; break;` (or `return -EFAULT;`).

_Same pattern in 6.1/0010 and 7.0/0010._

scripts/ext_sync_gen.py
```diff
@@ -0,0 +1,82 @@
...
+    if not os.path.exists(CDI_TSC_DEV):
+        print(f"Error: {CDI_TSC_DEV} not found (TSC driver not loaded?)",
+              file=sys.stderr)
+        sys.exit(1)
+
+    try:
+        fd = os.open(CDI_TSC_DEV, os.O_RDWR)
+        try:
+            if args.enable:
+                if args.fps is not None or args.duty is not None:
```
sareluzi, line 57: **N8 [LOW]** `--enable` silently inherits DT default rate when neither `--fps` nor `--duty` is given

```python
if args.fps is not None or args.duty is not None:
    fps = args.fps if args.fps is not None else 30
    duty = args.duty if args.duty is not None else 25
    tsc_set_rate(fd, fps, duty)
    print(f"Rate set: {fps} Hz, {duty}% duty")
tsc_fsync(fd, 1)
```

If the user runs `ext_sync_gen.py --enable` with no rate flags, the script never calls `tsc_set_rate` and the generators run at whatever the DT configured (30 Hz / 25%). The `--help` text doesn't mention this fall-through.

Options:
1. Document the fall-through in `--help` (`"...if omitted, uses DT default (30 Hz / 25%)"`) and the script docstring.
2. Or always issue `tsc_set_rate` with the 30/25 default so behaviour is identical regardless of how DT was configured.

Preference for option 2 — makes the script's effect deterministic and independent of which overlay was loaded.

nvidia-oot/6.1/0010-Runtime-tsc-rate-config.patch
```diff
@@ -0,0 +1,144 @@
+From 20b317ea3a831ed164c29c7e26563ff291356a58 Mon Sep 17 00:00:00 2001
```
sareluzi, line 1: **N9 [LOW]** Document the JP6.1 directory restructure in the commit body

The `nvidia-oot/6.1` directory was previously a single symlink → `6.0`. This PR replaces it with a real directory of per-file symlinks (`0001`, `0003`, `0004`, `0005`, `0006`, `0007`, `0008`, `0009`) plus this real `0010-Runtime-tsc-rate-config.patch` that diverges from 6.0's version (the JP6.1 patch additionally drops the `controller->opened` single-open guard and the auto-stop on close).

Worth a one-line note in the commit body explaining:
1. **Why** 6.1 needs a divergent runtime-tsc patch (the cdi_tsc lifecycle in JP6.1's NVIDIA tree differs from 6.0's, specifically around the `opened` flag).
2. **How** future updates to `6.0/0001..0009` should be propagated — these are now symlinks but a non-trivial change in 6.0 may or may not apply cleanly via the symlinked path.

_(Complements your existing thread on `6.1/0001-…` asking how it's related.)_

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4.dts
ejgoldik: Fixed

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.dts
sareluzi: Re-opening: commit 71438d91 "Restore pinctrl-0 definition" puts the override back on the pinmux node. Same concern as before — this replaces the platform default state list rather than adding to it. Tegra pinctrl applies child state nodes regardless of `pinctrl-0`, so the `tsc_sync_pinmux: tsc-sync-pins {…}` child alone is sufficient.

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.dts
sareluzi: Resolving: the new `docs/external-sync-fg12-16ch.md` (added in 6423cf4b) covers Option A (TSC→MFP2) and Option B (external gen→MFP6) with signal paths, so the discoverability concern is addressed. The DT comment can stay as-is.

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-1-2-3.dts
sareluzi: Resolving: fixed by 6d9b210d "Overlay wording fix" — comment now reads `this routing on Orin`.

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4.dts
sareluzi: Re-anchor note: after commit 6d9b210d, the typo line moved from 67 → **69** in the file (the wording fix added 2 lines above it). The bug itself is unchanged — `nvidia,pins = "can0_stb_paa7"` is still there at line 69 and still needs to be `can0_err_paa7`.

scripts/ext_sync_gen.py
ejgoldik: Fixed

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.dts
ejgoldik: Verified, it is necessary

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4-8-12.dts
sareluzi: Follow-up: https://rsjira.realsenseai.com/browse/RSDSO-21457

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4.dts
ejgoldik: Fixed

nvidia-oot/6.0/0010-Runtime-tsc-rate-config.patch
sareluzi: ignore

nvidia-oot/6.0/0010-Runtime-tsc-rate-config.patch
sareluzi: To follow-up - https://rsjira.realsenseai.com/browse/RSDSO-21478

sareluzi (approved):

PR #437, opened 2026-05-11
Add support for D5xx cameras with MAX96717/MAX96724

pinquanw:
## Summary
- add a Tegra234 device-tree overlay for RealSense D5xx / D585 camera bring-up
- define 4 CSI virtual channels for depth, IR-left, IR-right, and RGB streams
- configure MAX96724, MAX96717, HKR, and sensor aliasing for the AGX Orin CSI path

## Validation
- build_all.sh 6.2 ./Linux_for_Tegra/source


ymodlin:
@ejgoldik please add  your review

PR #440, opened 2026-05-12
[FIX] Ci instability

Nikolai-L:
[FIX] CI stability
* Implement SerDes startup I2C readiness polling with timeout and retry mechanism
* Add script to power-cycle test agent using Raritan PDU outlet

### DIFF
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -3588,6 +3588,55 @@ static void ds5_init_ds5_dev(struct ds5 *state, struct ds5_dev *ds5_dev)
 	ds5_reset_streaming_flags(ds5_dev);
 }
 
+/* Caller must hold serdes_lock__. */
+static bool ds5_release_slot(struct ds5 *state)
+{
+	struct dser_control *dser_control;
+	bool has_other_users = false;
+	bool released = false;
+	int i;
+
+	if (!state->ds5_dev)
+		return false;
+
+	mutex_lock(&state->ds5_dev->lock);
+	if (state->ds5_dev->ds5_primary == state) {
+		dser_control = state->ds5_dev->dser_control;
+		state->ds5_dev->ds5_primary = NULL;
+		state->ds5_dev->serdes_setup_complete = false;
+		state->ds5_dev->dser_control = NULL;
+		released = true;
+	} else {
+		dser_control = NULL;
+	}
+	mutex_unlock(&state->ds5_dev->lock);
+
+	if (!released || !dser_control)
+		return released;
+
+	for (i = 0; i < MAX_DS5_NUM; i++) {
+		bool in_use;
+
+		mutex_lock(&ds5_inited[i].lock);
+		in_use = ds5_inited[i].ds5_primary &&
+			ds5_inited[i].dser_control == dser_control;
+		mutex_unlock(&ds5_inited[i].lock);
+		if (in_use) {
+			has_other_users = true;
+			break;
+		}
+	}
+
+	if (!has_other_users) {
+		mutex_lock(&dser_control->lock);
+		if (dser_control->dser_dev == state->dser_dev)
+			dser_control->dser_dev = NULL;
+		mutex_unlock(&dser_control->lock);
+	}
+
+	return released;
+}
+
 #ifdef CONFIG_VIDEO_D4XX_SERDES
 static int ds5_setup_and_link(struct ds5 *state)
 {
@@ -4120,14 +4169,8 @@ serdes_setup_end:
 	if (ret) {
 		max9295_sdev_unpair(state->ser_dev, state->g_ctx.s_dev);
 		state->dser_ops->sdev_unregister(state->dser_dev, state->g_ctx.s_dev);
-		if (state->serdes_primary) {
-			mutex_lock(&state->ds5_dev->lock);
-			if (state->ds5_dev->ds5_primary == state) {
-				state->ds5_dev->ds5_primary = NULL;
-				state->ds5_dev->serdes_setup_complete = false;
-			}
-			mutex_unlock(&state->ds5_dev->lock);
-		}
+		if (state->serdes_primary)
+			ds5_release_slot(state);
 	} else if (state->serdes_primary) {
 		mutex_lock(&state->ds5_dev->lock);
 		if (state->ds5_dev->ds5_primary == state)
@@ -6465,15 +6508,7 @@ static void ds5_remove(struct i2c_client *c)
 		bool do_cleanup = false;
 
 		mutex_lock(&serdes_lock__);
-		mutex_lock(&state->ds5_dev->lock);
-
-		if (state->ds5_dev->ds5_primary) {
-			state->ds5_dev->ds5_primary = NULL;
-			state->ds5_dev->serdes_setup_complete = false;
-			do_cleanup = true;
-		}
-
-		mutex_unlock(&state->ds5_dev->lock);
+		do_cleanup = ds5_release_slot(state);
 
 		if (do_cleanup) {
 			ret = max9295_reset_control(state->ser_dev);
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -4038,8 +4038,8 @@ static int ds5_gmsl_serdes_setup(struct ds5 *state)
 	}
 	if (err) {
 		dev_err(dev,
-			"%s(): serializer setup timed out after %d ms, deferring probe\n",
-			__func__, DS5_SERDES_STARTUP_TIMEOUT_MS);
+			"%s(): serializer setup failed (err=%d) after %d ms, deferring probe\n",
+			__func__, err, DS5_SERDES_STARTUP_TIMEOUT_MS);
 		err = -EPROBE_DEFER;
 		goto error;
 	}
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -3599,6 +3599,8 @@ static bool ds5_release_slot(struct ds5 *state)
 	if (!state->ds5_dev)
 		return false;
 
+	mutex_lock(&serdes_lock__);
+
 	mutex_lock(&state->ds5_dev->lock);
 	if (state->ds5_dev->ds5_primary == state) {
 		dser_control = state->ds5_dev->dser_control;
@@ -3611,8 +3613,10 @@ static bool ds5_release_slot(struct ds5 *state)
 	}
 	mutex_unlock(&state->ds5_dev->lock);
 
-	if (!released || !dser_control)
+	if (!released || !dser_control) {
+		mutex_unlock(&serdes_lock__);
 		return released;
+	}
 
 	for (i = 0; i < MAX_DS5_NUM; i++) {
 		bool in_use;
@@ -3634,6 +3638,7 @@ static bool ds5_release_slot(struct ds5 *state)
 		mutex_unlock(&dser_control->lock);
 	}
 
+	mutex_unlock(&serdes_lock__);
 	return released;
 }
 
@@ -6507,7 +6512,6 @@ static void ds5_remove(struct i2c_client *c)
 		int ret;
 		bool do_cleanup = false;
 
-		mutex_lock(&serdes_lock__);
 		do_cleanup = ds5_release_slot(state);
 
 		if (do_cleanup) {
@@ -6531,9 +6535,6 @@ static void ds5_remove(struct i2c_client *c)
 					"failed to %s unregister sdev\n", state->dser_ops->name);
 			state->dser_ops->power_off(state->dser_dev);
 		}
-
-		mutex_unlock(&serdes_lock__);
-
 	}
 #ifndef CONFIG_OF
 	if (state->ser_i2c)
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -4045,9 +4045,6 @@ static int ds5_gmsl_serdes_setup(struct ds5 *state)
 	}
 
 	/* proceed even if ser setup failed, to setup deser correctly */
-	if (err)
-		dev_err(dev, "gmsl serializer setup failed\n");
-
 	des_err = state->dser_ops->setup_control(state->dser_dev, &state->client->dev);
 	if (des_err) {
 		dev_err(dev, "gmsl deserializer setup failed\n");
@@ -4116,6 +4113,10 @@ static int ds5_serdes_setup(struct ds5 *state)
 	}
 
 serdes_setup_end:
+	/* Set/clear serdes_setup_complete from the same exit gate: error branch
+	 * clears it, success branch sets it. This ensures flag state is
+	 * synchronized with actual setup completion status.
+	 */
 	if (ret) {
 		max9295_sdev_unpair(state->ser_dev, state->g_ctx.s_dev);
 		state->dser_ops->sdev_unregister(state->dser_dev, state->g_ctx.s_dev);
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -134,6 +134,10 @@ struct dser_interface {
 #define DS5_RGB_FPS				0x402C
 #define DS5_RGB_CONTROL_STATUS 	0x402E
 
+/* SerDes startup I2C readiness polling (defer probe if not responsive) */
+#define DS5_SERDES_STARTUP_TIMEOUT_MS 2000
+#define DS5_SERDES_STARTUP_RETRY_DELAY_MS 100
+
 #define DS5_IMU_STREAM_DT		0x4040
 #define DS5_IMU_STREAM_MD		0x4042
 #define DS5_IMU_RES_WIDTH		0x4044
@@ -540,6 +544,7 @@ struct ds5_dev {
 
 	/* Pointer to the primary DS5 struct */
 	struct ds5 *ds5_primary;
+	bool serdes_setup_complete;
 
 	bool depth_streaming;
 	bool ir_streaming;
@@ -3577,6 +3582,7 @@ static void ds5_init_ds5_dev(struct ds5 *state, struct ds5_dev *ds5_dev)
 	state->ds5_dev = ds5_dev;
 	mutex_lock(&ds5_dev->lock);
 	ds5_dev->ds5_primary = state;
+	ds5_dev->serdes_setup_complete = false;
 	ds5_dev->cached_device_type = DS5_DEVICE_TYPE_UNKNOWN;
 	mutex_unlock(&ds5_dev->lock);
 	ds5_reset_streaming_flags(ds5_dev);
@@ -3595,12 +3601,18 @@ static int ds5_setup_and_link(struct ds5 *state)
 	/* Look for existing DS5 instances */
 	for (i = 0; i < MAX_DS5_NUM; i++) {
 		bool match;
+		bool ready;
 
 		mutex_lock(&ds5_inited[i].lock);
 		match = ds5_inited[i].ds5_primary &&
 			ds5_inited[i].ds5_primary->ser_dev == state->ser_dev;
+		ready = ds5_inited[i].serdes_setup_complete;
 		mutex_unlock(&ds5_inited[i].lock);
 		if (match) { /* Same camera, different stream instance. */
+			if (!ready) {
+				err = -EPROBE_DEFER;
+				goto out_unlock;
+			}
 			state->serdes_primary = false;
 			state->ds5_dev = &ds5_inited[i];
 			break;
@@ -3986,6 +3998,8 @@ static int ds5_gmsl_serdes_setup(struct ds5 *state)
 {
 	int err = 0;
 	int des_err = 0;
+	int attempts;
+	int retry;
 	struct device *dev;
 
 	if (!state || !state->ser_dev || !state->dser_dev || !state->client)
@@ -4012,7 +4026,23 @@ static int ds5_gmsl_serdes_setup(struct ds5 *state)
 		goto error;
 	}
 	msleep(100);
-	err = max9295_setup_control(state->ser_dev);
+	attempts = (DS5_SERDES_STARTUP_TIMEOUT_MS +
+		DS5_SERDES_STARTUP_RETRY_DELAY_MS - 1) /
+		DS5_SERDES_STARTUP_RETRY_DELAY_MS;
+	for (retry = 0; retry < attempts; retry++) {
+		err = max9295_setup_control(state->ser_dev);
+		if (!err)
+			break;
+		if (retry < attempts - 1)
+			msleep(DS5_SERDES_STARTUP_RETRY_DELAY_MS);
+	}
+	if (err) {
+		dev_err(dev,
+			"%s(): serializer setup timed out after %d ms, deferring probe\n",
+			__func__, DS5_SERDES_STARTUP_TIMEOUT_MS);
+		err = -EPROBE_DEFER;
+		goto error;
+	}
 
 	/* proceed even if ser setup failed, to setup deser correctly */
 	if (err)
@@ -4089,6 +4119,19 @@ serdes_setup_end:
 	if (ret) {
 		max9295_sdev_unpair(state->ser_dev, state->g_ctx.s_dev);
 		state->dser_ops->sdev_unregister(state->dser_dev, state->g_ctx.s_dev);
+		if (state->serdes_primary) {
+			mutex_lock(&state->ds5_dev->lock);
+			if (state->ds5_dev->ds5_primary == state) {
+				state->ds5_dev->ds5_primary = NULL;
+				state->ds5_dev->serdes_setup_complete = false;
+			}
+			mutex_unlock(&state->ds5_dev->lock);
+		}
+	} else if (state->serdes_primary) {
+		mutex_lock(&state->ds5_dev->lock);
+		if (state->ds5_dev->ds5_primary == state)
+			state->ds5_dev->serdes_setup_complete = true;
+		mutex_unlock(&state->ds5_dev->lock);
 	}
 
 	return ret;
@@ -6425,6 +6468,7 @@ static void ds5_remove(struct i2c_client *c)
 
 		if (state->ds5_dev->ds5_primary) {
 			state->ds5_dev->ds5_primary = NULL;
+			state->ds5_dev->serdes_setup_complete = false;
 			do_cleanup = true;
 		}
 
--- a/scripts/power_cycle_pdu.py
+++ b/scripts/power_cycle_pdu.py
@@ -54,7 +54,10 @@ def rpc(path, method, params=None):
     if params is not None: d["params"] = params
     req = u.Request(f"https://{host}{path}", data=json.dumps(d).encode(),
                     headers={"Authorization": auth, "Content-Type": "application/json"})
-    return json.loads(u.urlopen(req, context=ctx, timeout=20).read().decode())
+    resp = json.loads(u.urlopen(req, context=ctx, timeout=20).read().decode())
+    if resp.get("error"):
+        raise RuntimeError(f"JSON-RPC error: {resp['error']}")
+    return resp
 
 try:
     outs = rpc("/model/pdu/0", "getOutlets")["result"]["_ret_"]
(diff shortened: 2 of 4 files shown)


scripts/power_cycle_pdu.py
```diff
@@ -0,0 +1,84 @@
...
+
+print(f"Agent name: {agent_name}")
+print(f"Agent short name: {agent_short_name}")
+print(f"Delay before power-cycle: {delay_sec}s")
+
+# Wait before cycling
+time.sleep(delay_sec)
+
+ctx = ssl._create_unverified_context()
+auth = "Basic " + base64.b64encode(f"{user}:{pw}".encode()).decode()
```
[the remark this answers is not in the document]
Nikolai-L: For intranet use only

sareluzi (commented):
Code review — SerDes I2C readiness polling + PDU script.

### `kernel/realsense/d4xx.c`

- **L4035-4042 — 🟡 risk:** `-EPROBE_DEFER` path makes the following `/* proceed even if ser setup failed */` comment + `if (err) dev_err(...)` block dead code. Either delete the stale fallback or drop the defer and keep the original behavior. Mixed intent will confuse future readers.
- **L4029-4035 — 🟡 risk:** retry loop re-runs `max9295_setup_control` without redoing `power_off`/`power_on`/`setup_link`. If the serializer is wedged (not just slow), 20 × 100 ms are wasted before deferring. Consider exiting the loop early on non-`-ETIMEDOUT`/`-ENXIO` errors.
- **L4036 — 🟡 risk:** `dser_ops->setup_link` already mutated deserializer state before we return `-EPROBE_DEFER`. Nothing undoes it before `goto error`; relies on the next probe's `power_off`+`power_on` to reset. Either call `dser_ops->power_off` before the goto or document the assumption.
- **L4029 — 🔵 nit:** use `DIV_ROUND_UP(DS5_SERDES_STARTUP_TIMEOUT_MS, DS5_SERDES_STARTUP_RETRY_DELAY_MS)` instead of open-coded `(a+b-1)/b`.
- **L4032 — 🔵 nit:** log each attempt at `dev_dbg` — when CI flakes again, current logging only shows the final failure and hides retry pattern.
- **L3608-3611 — 🟡 risk:** sibling defers indefinitely while `!ready`. If primary's probe is itself stuck deferred, both block until `deferred_probe_timeout`. Add a one-shot `dev_info` when first deferring so it's diagnosable from dmesg.
- **L4122-4127 — 🟡 risk:** on primary error you clear `ds5_primary=NULL` but leave `dser_inited[j].dser_dev` still set (assigned in `ds5_setup_and_link`). Slot is half-released — next probe sees stale state. Mirror `ds5_remove`'s full slot cleanup or release the dser slot here too.
- **L4119 — ❓ q:** success branch sets `serdes_setup_complete=true` before `max9295_init_settings` / `dser_ops->init_settings` (called later in `ds5_serdes_setup` at L4329-4340). If those fail and we `goto serdes_setup_end`, `ret` is set so we hit the error branch and clear the flag — OK. Worth a comment that the flag is set/cleared from the same exit gate.

### `scripts/power_cycle_pdu.py`

- **L7 — 🟡 risk:** password passed as argv → visible in `ps` and shell history. Read from env var (`PDU_PASS`) or stdin instead.
- **L43 — 🟡 risk:** `ssl._create_unverified_context()` disables certificate validation unconditionally. Acceptable for a lab PDU but make it an explicit `--insecure` flag rather than hard-coding.
- **L47-49 — 🔵 nit:** `time.sleep(delay_sec)` runs before any auth/network check. Bad CLI args force you to wait the full delay before failing. Move the sleep after the first successful `rpc()` call.
- **L64-68 — 🟡 risk:** name resolution loops through every outlet sequentially with a 20 s per-call timeout. 24-outlet PDU + flaky network = up to 8 min before failure. Use a shorter per-call timeout (5 s) or fetch all settings in a batch.
- **L75 — 🟡 risk:** JSON-RPC errors are returned as `{"error": {...}}` in the response body. Code prints `result` without checking for `error` key — a failed cycle is silently logged as success. Check `result.get("error")` and raise.
- **L80-82 — 🔵 nit:** bare `except Exception` swallows the traceback. For lab debugging keep it (`logging.exception` or re-raise after logging).
- **L51-56 — 🔵 nit:** top-level module code; wrap in `if __name__ == "__main__": main()` for testability.

### Summary

Two worth fixing before merge:
1. Dead-code/stale comment in `ds5_gmsl_serdes_setup` after the EPROBE_DEFER return.
2. Half-released `dser_control` slot on primary failure path in `ds5_serdes_setup`.

kernel/realsense/d4xx.c
```diff
@@ -3595,12 +3601,18 @@ static int ds5_setup_and_link(struct ds5 *state)
...
 		bool match;
+		bool ready;
 
 		mutex_lock(&ds5_inited[i].lock);
 		match = ds5_inited[i].ds5_primary &&
 			ds5_inited[i].ds5_primary->ser_dev == state->ser_dev;
+		ready = ds5_inited[i].serdes_setup_complete;
 		mutex_unlock(&ds5_inited[i].lock);
 		if (match) { /* Same camera, different stream instance. */
+			if (!ready) {
```
sareluzi, line 3666: 🟡 risk: sibling defers indefinitely while `!ready`. If the primary's probe is itself stuck deferred, both block until `deferred_probe_timeout`. Add a one-shot `dev_info` here when first deferring so the wait is diagnosable from dmesg.

sareluzi (commented):
Inline notes on the SerDes readiness polling and PDU script. Two items worth fixing before merge: dead-code/comment after `EPROBE_DEFER` in `ds5_gmsl_serdes_setup`, and half-released `dser_control` slot on primary failure in `ds5_serdes_setup`.

kernel/realsense/d4xx.c
```diff
@@ -4012,7 +4026,23 @@ static int ds5_gmsl_serdes_setup(struct ds5 *state)
 		goto error;
 	}
 	msleep(100);
-	err = max9295_setup_control(state->ser_dev);
+	attempts = (DS5_SERDES_STARTUP_TIMEOUT_MS +
```
sareluzi, line 4083: 🔵 nit: use `DIV_ROUND_UP(DS5_SERDES_STARTUP_TIMEOUT_MS, DS5_SERDES_STARTUP_RETRY_DELAY_MS)` instead of open-coded `(a+b-1)/b`.

kernel/realsense/d4xx.c
```diff
@@ -4012,7 +4026,23 @@ static int ds5_gmsl_serdes_setup(struct ds5 *state)
...
+		DS5_SERDES_STARTUP_RETRY_DELAY_MS - 1) /
+		DS5_SERDES_STARTUP_RETRY_DELAY_MS;
+	for (retry = 0; retry < attempts; retry++) {
+		err = max9295_setup_control(state->ser_dev);
```
sareluzi, line 4087: 🟡 risk: retry loop re-runs `max9295_setup_control` without redoing `power_off`/`power_on`/`setup_link`. If the serializer is wedged (not just slow), 20 × 100 ms are wasted before the defer kicks in. Either break out early on non-`-ETIMEDOUT`/`-ENXIO` errors, or log each attempt at `dev_dbg` so retry behavior is visible when CI flakes again.

kernel/realsense/d4xx.c
```diff
@@ -4012,7 +4026,23 @@ static int ds5_gmsl_serdes_setup(struct ds5 *state)
...
+			break;
+		if (retry < attempts - 1)
+			msleep(DS5_SERDES_STARTUP_RETRY_DELAY_MS);
+	}
+	if (err) {
+		dev_err(dev,
+			"%s(): serializer setup timed out after %d ms, deferring probe\n",
+			__func__, DS5_SERDES_STARTUP_TIMEOUT_MS);
+		err = -EPROBE_DEFER;
+		goto error;
```
sareluzi, line 4098: 🟡 risk: `dser_ops->setup_link` already mutated deserializer state before we return `-EPROBE_DEFER`. Nothing undoes it before `goto error`; relies on the next probe's `power_off`+`power_on` to reset. Either call `dser_ops->power_off` before this goto or add a comment documenting the assumption.

kernel/realsense/d4xx.c
```diff
@@ -4012,7 +4026,23 @@ static int ds5_gmsl_serdes_setup(struct ds5 *state)
...
+	}
 
 	/* proceed even if ser setup failed, to setup deser correctly */
```
sareluzi, line 4101: 🔴 bug: dead code. After the `EPROBE_DEFER` return above, `err == 0` here is the only reachable state, so the `/* proceed even if ser setup failed */` comment + `if (err) dev_err(...)` block can never fire. Either delete this stale fallback or drop the defer and keep the original behavior. Mixed intent will confuse future readers.

kernel/realsense/d4xx.c
```diff
@@ -4089,6 +4119,19 @@ static int ds5_serdes_setup(struct ds5 *state)
 	if (ret) {
 		max9295_sdev_unpair(state->ser_dev, state->g_ctx.s_dev);
 		state->dser_ops->sdev_unregister(state->dser_dev, state->g_ctx.s_dev);
+		if (state->serdes_primary) {
+			mutex_lock(&state->ds5_dev->lock);
+			if (state->ds5_dev->ds5_primary == state) {
+				state->ds5_dev->ds5_primary = NULL;
```
sareluzi: 🟡 risk: clearing `ds5_primary=NULL` here leaves `dser_inited[j].dser_dev` still pointing at this state's deserializer (assigned in `ds5_setup_and_link`). Slot is half-released — next probe sees stale state. Mirror `ds5_remove`'s full slot cleanup (also release the `dser_control` entry) or push the cleanup into a shared helper.

kernel/realsense/d4xx.c
```diff
@@ -4089,6 +4119,19 @@ static int ds5_serdes_setup(struct ds5 *state)
...
+				state->ds5_dev->serdes_setup_complete = false;
+			}
+			mutex_unlock(&state->ds5_dev->lock);
+		}
+	} else if (state->serdes_primary) {
+		mutex_lock(&state->ds5_dev->lock);
+		if (state->ds5_dev->ds5_primary == state)
+			state->ds5_dev->serdes_setup_complete = true;
```
sareluzi, line 4182: ❓ q: success branch sets `serdes_setup_complete = true` before `max9295_init_settings` / `dser_ops->init_settings` run (called later in `ds5_serdes_setup`). If either fails and we `goto serdes_setup_end`, `ret` is set so we re-enter the error branch and clear the flag — OK. Worth a one-line comment noting the flag is set and cleared from the same exit gate.

scripts/power_cycle_pdu.py
```diff
@@ -0,0 +1,83 @@
...
+    if len(sys.argv) != 6:
+        print(
+            "Usage: power_cycle_pdu.py <host> <user> <pass> <node_name> <delay_sec>",
+            file=sys.stderr,
+        )
+        sys.exit(2)
+
+    host = sys.argv[1]
+    user = sys.argv[2]
+    pw = sys.argv[3]
```
sareluzi, line 22: 🟡 risk: password passed as `argv[3]` is visible in `ps`, `/proc/<pid>/cmdline`, and shell history. Read from an env var (e.g. `PDU_PASS`) or stdin instead.

scripts/power_cycle_pdu.py
```diff
@@ -0,0 +1,83 @@
...
+
+host, user, pw, agent_name, agent_short_name, delay_sec = parse_args()
+
+print(f"Raritan host: {host}")
+print(f"Agent name: {agent_name}")
+print(f"Agent short name: {agent_short_name}")
+print(f"Delay before power-cycle: {delay_sec}s")
+
+# Wait before cycling
+time.sleep(delay_sec)
```
sareluzi, line 47: 🔵 nit: `time.sleep(delay_sec)` runs before any auth/network check. A typo in args forces you to wait the full delay before failing. Move the sleep until after the first successful `rpc()` call.

scripts/power_cycle_pdu.py
```diff
@@ -0,0 +1,83 @@
...
+
+ctx = ssl._create_unverified_context()
```
sareluzi, line 49: 🟡 risk: `ssl._create_unverified_context()` disables certificate validation unconditionally. Acceptable for a lab PDU, but expose it as an explicit `--insecure` flag rather than hard-coding so a misuse on a non-lab host is visible.

scripts/power_cycle_pdu.py
```diff
@@ -0,0 +1,83 @@
...
+    return json.loads(u.urlopen(req, context=ctx, timeout=20).read().decode())
+
+try:
+    outs = rpc("/model/pdu/0", "getOutlets")["result"]["_ret_"]
+    candidates = {agent_name, agent_short_name}
+
+    idx = None
+    seen_names = []
+    for outlet in outs:
+        outlet_name = rpc(outlet["rid"], "getSettings")["result"]["_ret_"].get("name")
```
sareluzi, line 69: 🟡 risk: outlet-name resolution loops sequentially with the 20 s per-call timeout from L57. A 24-outlet PDU on a flaky network can hang up to 8 min before failure. Drop the per-call timeout (e.g. 5 s) or batch the settings fetch.

scripts/power_cycle_pdu.py
```diff
@@ -0,0 +1,83 @@
...
+            idx = int(outlet["rid"].rsplit(".", 1)[1])
+            break
+
+    if idx is None:
+        raise RuntimeError(
+            f"No outlet found for {agent_name} (short {agent_short_name}). "
+            f"Available names: {seen_names}"
+        )
+
+    result = rpc(f"/model/pdu/0/outlet/{idx}", "cyclePowerState")
```
sareluzi: 🟡 risk: JSON-RPC failures are returned as `{"error": {...}}` in the response body. This prints `result` without checking for an `error` key — a failed `cyclePowerState` is silently reported as success. Check `result.get("error")` and raise.

scripts/power_cycle_pdu.py
```diff
@@ -0,0 +1,83 @@
...
+    print(f"Power-cycle result: {result}")
+except Exception as e:
```
sareluzi, line 84: 🔵 nit: bare `except Exception` discards the traceback. For lab debugging keep it — `logging.exception(...)` or re-raise after logging — otherwise root-cause needs another reproduction.

scripts/power_cycle_pdu.py
```diff
@@ -0,0 +1,83 @@
...
+
+    if delay_sec < 0:
+        print("delay_sec must be >= 0", file=sys.stderr)
+        sys.exit(2)
+
+    short_name = node_name.split(".", 1)[0]
+    return host, user, pw, node_name, short_name, delay_sec
+
+
+host, user, pw, agent_name, agent_short_name, delay_sec = parse_args()
```
sareluzi, line 39: 🔵 nit: top-level module code. Wrap as `if __name__ == "__main__": main()` so the file can be imported (e.g. by a test harness) without triggering side effects.

scripts/power_cycle_pdu.py
```diff
@@ -0,0 +1,83 @@
...
+
+ctx = ssl._create_unverified_context()
+auth = "Basic " + base64.b64encode(f"{user}:{pw}".encode()).decode()
+
+def rpc(path, method, params=None):
+    d = {"jsonrpc": "2.0", "method": method, "id": 1}
+    if params is not None: d["params"] = params
+    req = u.Request(f"https://{host}{path}", data=json.dumps(d).encode(),
+                    headers={"Authorization": auth, "Content-Type": "application/json"})
+    return json.loads(u.urlopen(req, context=ctx, timeout=20).read().decode())
```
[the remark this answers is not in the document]
Nikolai-L: @AikidoSec ignore: internal utility, not critical

kernel/realsense/d4xx.c
```diff
@@ -4086,9 +4113,26 @@ static int ds5_serdes_setup(struct ds5 *state)
...
+	 * clears it, success branch sets it. This ensures flag state is
+	 * synchronized with actual setup completion status.
+	 */
 	if (ret) {
 		max9295_sdev_unpair(state->ser_dev, state->g_ctx.s_dev);
 		state->dser_ops->sdev_unregister(state->dser_dev, state->g_ctx.s_dev);
+		if (state->serdes_primary) {
+			mutex_lock(&state->ds5_dev->lock);
+			if (state->ds5_dev->ds5_primary == state) {
+				state->ds5_dev->ds5_primary = NULL;
```
sareluzi: 🔴 **bump — merge-blocking.** Primary-failure cleanup is half-done. Clearing `ds5_primary=NULL` here leaves `dser_inited[j].dser_dev` still pointing at this state's deserializer (assigned in `ds5_setup_and_link`). A subsequent probe of the same camera will see an orphan dser slot and misroute the new primary onto stale state.

Fix: mirror the full slot release that `ds5_remove` performs — release both the `ds5_inited[]` and `dser_inited[]` entries here, or extract a `ds5_release_slot(state)` helper used by both paths. Same comment applies to any future error path that clears `ds5_primary`.

sareluzi (commented):
Bumping the two merge-blocking opens from the prior review. Both are real correctness bugs (not nits) and worth addressing before this lands.

scripts/power_cycle_pdu.py
sareluzi: 🔴 **bump — merge-blocking.** This is the CI-stability fix itself; if `cyclePowerState` fails server-side, JSON-RPC returns `{"error": {...}, "result": null}` with HTTP 200. The current code prints `result` and exits 0, so CI sees a successful power-cycle while the outlet was never cycled. Defeats the purpose of the script.

Fix:
```python
resp = rpc(f"/model/pdu/0/outlet/{idx}", "cyclePowerState")
if resp.get("error"):
    raise RuntimeError(f"cyclePowerState failed: {resp['error']}")
print(f"Power-cycle result: {resp['result']}")
```
Apply the same `error` check inside `rpc()` so all calls (`getOutlets`, `getSettings`) catch server-side failures.

kernel/realsense/d4xx.c
```diff
@@ -4086,9 +4162,20 @@ static int ds5_serdes_setup(struct ds5 *state)
...
 serdes_setup_end:
+	/* Set/clear serdes_setup_complete from the same exit gate: error branch
+	 * clears it, success branch sets it. This ensures flag state is
+	 * synchronized with actual setup completion status.
+	 */
 	if (ret) {
 		max9295_sdev_unpair(state->ser_dev, state->g_ctx.s_dev);
 		state->dser_ops->sdev_unregister(state->dser_dev, state->g_ctx.s_dev);
+		if (state->serdes_primary)
+			ds5_release_slot(state);
```
sareluzi, line 4178: 🔴 **bug: lock contract violation.** `ds5_release_slot()` is documented as `/* Caller must hold serdes_lock__. */` (see L3590). The other call site, `ds5_remove`, takes `mutex_lock(&serdes_lock__)` before calling it. This call path does **not** — `ds5_serdes_setup` only holds `serdes_lock__` transiently inside `ds5_board_setup` / `ds5_gmsl_serdes_setup`, both of which release it before returning. So at `serdes_setup_end:` the lock is dropped.

The `has_other_users` scan inside the helper walks `ds5_inited[i]` and `dser_inited[*]` without `serdes_lock__`, so a concurrent probe can reassign `dser_control` mid-scan and we may either NULL a slot that's still in use or miss a slot we should release.

Fix: take `serdes_lock__` around the call in this branch, e.g.
```c
if (state->serdes_primary) {
    mutex_lock(&serdes_lock__);
    ds5_release_slot(state);
    mutex_unlock(&serdes_lock__);
}
```
Alternatively, move the lock acquisition inside `ds5_release_slot()` and drop it from `ds5_remove`. Pick one ownership model and document it on the helper.

sareluzi (commented):
Round 3.

**Resolved by this commit** (already marked resolved on GitHub):
- Half-released `dser_control` slot → fixed via new `ds5_release_slot()` helper, also adopted in `ds5_remove`.
- JSON-RPC error check → handled inside `rpc()` so all calls raise on server-side errors.
- Dead `if (err) dev_err(...)` block → removed.
- Success-flag ordering → explanatory comment added at `serdes_setup_end`.

**New bug introduced.** See inline comment on `ds5_release_slot()` call from `ds5_serdes_setup` error path.

**Still open from prior review** (no changes — leaving threads open):
- d4xx.c: defer-log on L3661, `DIV_ROUND_UP` on L4078, retry without re-power-cycle on L4082, `setup_link` not undone before `EPROBE_DEFER` on L4093.
- power_cycle_pdu.py: password in argv (L22), sleep-before-validate (L47), `_create_unverified_context` (L49), 20 s per-call timeout (L69), bare `except` (L84), no `__main__` guard (L39).

kernel/realsense/d4xx.c
ymodlin: Fixed in 105b757 — the stale `/* proceed even if ser setup failed */` comment and the unreachable `if (err) dev_err(...)` block are removed.

kernel/realsense/d4xx.c
ymodlin: Fixed in 72f7fee — extracted into `ds5_release_slot()` (d4xx.c L3592–3638) and used from both `ds5_serdes_setup`'s error gate and `ds5_remove`. The lock-contract follow-up flagged on the new call site is being tracked separately.

kernel/realsense/d4xx.c
ymodlin: Fixed in 105b757 — comment added at `serdes_setup_end:` documenting that `serdes_setup_complete` is set in the success branch and cleared in the error branch, both from the same exit gate.

scripts/power_cycle_pdu.py
ymodlin: Fixed in 72f7fee — `rpc()` now checks `resp.get('error')` and raises `RuntimeError`, so `getOutlets`/`getSettings`/`cyclePowerState` all surface server-side failures. Call site reads `resp['result']`.

kernel/realsense/d4xx.c
ymodlin: Confirmed — this is a real lock-contract violation introduced by 72f7fee. `ds5_release_slot()` documents `/* Caller must hold serdes_lock__. */` (L3591), and `ds5_remove` honors that (L6510), but by the time we reach `serdes_setup_end:` both `ds5_gmsl_serdes_setup` and `ds5_board_setup` have already released the lock. The `has_other_users` scan races with concurrent probes. Should be fixed before merge — wrap the call in `mutex_lock(&serdes_lock__) / mutex_unlock(&serdes_lock__)`, or move acquisition inside the helper and drop it from `ds5_remove`.

kernel/realsense/d4xx.c
ymodlin: Open — agree. A one-shot `dev_info(dev, "deferring: waiting for primary serdes setup\n")` here makes the wait diagnosable from dmesg without spamming if the defer loops.

kernel/realsense/d4xx.c
ymodlin: Open — agree. `attempts = DIV_ROUND_UP(DS5_SERDES_STARTUP_TIMEOUT_MS, DS5_SERDES_STARTUP_RETRY_DELAY_MS);` is the idiomatic form.

kernel/realsense/d4xx.c
ymodlin: Open — agree. Two cheap improvements: (1) early-out on non-`-ETIMEDOUT`/`-ENXIO` errors so a permanently-wedged serializer doesn't burn 2 s before deferring; (2) per-attempt `dev_dbg(dev, "%s retry %d: err=%d\n", __func__, retry, err)` so retry patterns are visible when CI flakes.

kernel/realsense/d4xx.c
ymodlin: Open — agree. `dser_ops->setup_link` mutates deserializer state and we currently rely on the next probe's `power_off`+`power_on` to reset it. Either call `state->dser_ops->power_off(state->dser_dev)` before `goto error`, or add a brief comment documenting the next-probe-resets assumption.

scripts/power_cycle_pdu.py
ymodlin: Open — agree. Reading password from `PDU_PASS` env var (or stdin) keeps it out of `ps`/`/proc/<pid>/cmdline`/shell history. Argv form can stay as a documented fallback if needed.

scripts/power_cycle_pdu.py
ymodlin: Open — agree. Moving `time.sleep(delay_sec)` to after the first successful `rpc()` call means bad args/network fail fast.

scripts/power_cycle_pdu.py
ymodlin: Open — agree. Make it opt-in via a `--insecure` flag (default to verified context, system CA). Keeps the lab-PDU use case working while flagging misuse on non-lab hosts.

scripts/power_cycle_pdu.py
ymodlin: Open — agree. With 24 outlets the 20 s per-call timeout compounds to ~8 min worst case. Dropping to 5 s, or batching the settings fetch into a single call, would bound the failure time.

scripts/power_cycle_pdu.py
ymodlin: Open — agree. `logging.exception("power-cycle failed")` (or re-raise after logging) preserves the traceback for lab debugging.

scripts/power_cycle_pdu.py
ymodlin: Open — agree. Wrapping the top-level into a `main()` and guarding with `if __name__ == "__main__": main()` lets the file be imported without side effects (useful for a future test harness).

kernel/realsense/d4xx.c
ymodlin: Update: fixed in aeb07cb — `ds5_release_slot()` now acquires `serdes_lock__` internally (with release on both exit paths) and `ds5_remove` no longer wraps the call. The `has_other_users` scan is now race-free.

Tiny follow-up: the kerneldoc-style comment `/* Caller must hold serdes_lock__. */` at L3591 is now stale — should be inverted, e.g. `/* Acquires serdes_lock__ internally; caller must not hold it. */` to avoid a future deadlock if someone wraps the call thinking the old contract applies.

ymodlin:
## Review summary

### Status
- **CI**: all 10 builds green (`v4l2-test` pending). Aikido: passed.
- **Merge-blocker**: ✅ resolved by `aeb07cb` — `ds5_release_slot()` now self-acquires `serdes_lock__`, `ds5_remove` updated to match.
- **PR is correctness-clean** at `aeb07cb` and ready to merge from a review standpoint.

### Fixed during review
| Issue | Fixed in |
|---|---|
| Dead code after `EPROBE_DEFER` (`d4xx.c`) | `105b757` |
| Success-flag ordering comment at `serdes_setup_end:` | `105b757` |
| Half-released `dser_control` slot on primary failure | `72f7fee` (extracted `ds5_release_slot()`) |
| JSON-RPC server-side errors silently swallowed (`power_cycle_pdu.py`) | `72f7fee` (raise in `rpc()`) |
| Improved timeout log to include error code | `6b9a4da` |
| Lock-contract violation on new `ds5_release_slot()` call site | `aeb07cb` |

### Open follow-up items (non-blocking)

`kernel/realsense/d4xx.c`:
- One-shot `dev_info` on first defer in `ds5_setup_and_link` (L3661)
- Use `DIV_ROUND_UP` macro for `attempts` calc (L4078)
- Early-out on non-transient errors + per-attempt `dev_dbg` in retry loop (L4082)
- `setup_link` mutation not undone before `-EPROBE_DEFER` (L4093)
- Stale kerneldoc on `ds5_release_slot()` — `/* Caller must hold serdes_lock__. */` is now inverted (L3591)

`scripts/power_cycle_pdu.py`:
- Read password from `PDU_PASS` env var instead of `argv[3]` (L22)
- Wrap top-level in `if __name__ == "__main__": main()` (L39)
- Move `time.sleep(delay_sec)` to after first `rpc()` call (L47)
- Make unverified TLS context opt-in via `--insecure` flag (L49)
- Shorter per-call timeout (5 s) or batched settings fetch — current 20 s × 24 outlets = ~8 min worst case (L69)
- `logging.exception(...)` instead of bare `except Exception` (L84)

### Recommendation
LGTM for merge. The 10 open items + stale-comment fix are all polish; suggest landing them in a follow-up PR so the CI-stability fix can ship now.

ymodlin (approved):

PR #441, opened 2026-05-17
RSDEV-7848  - Serdes and I2c less driver (for D500 bringup)

ejgoldik:


kernel/realsense/d4xx.c
```diff
@@ -79,6 +79,8 @@ struct dser_interface {
 #define GMSL_CSI_DT_EMBED 0x12
 #endif
 
+#define I2C_LESS 1
```
ymodlin, line 82: **BUG if merged** — `I2C_LESS` is hardcoded on, so every build of `d4xx.c` loses all FW communication (stream start becomes a no-op, formats/device type are faked). Same for the `#subdir-ccflags-y += -DCONFIG_VIDEO_D4XX_SERDES` change in `nvidia-oot/6.0` and `6.2` `0001-*.patch`, which disables the SerDes path for all JP6 consumers. This needs to be an opt-in build flag (Kconfig or a dedicated `subdir-ccflags` in a bringup-only patch), never a `#define` in the shared source.

kernel/realsense/d4xx.c
```diff
@@ -6259,16 +6285,15 @@ static int ds5_probe(struct i2c_client *c
 	state->reset_ref_dser = atomic_read(dser_get_reset_gen(state));
 #else
 	ds5_init_global_slots_once();
-	mutex_lock(&ds5_inited[0].lock);
 	if (NULL == ds5_inited[0].ds5_primary) {
```
ymodlin, line 6288: **BUG** — dropping `mutex_lock(&ds5_inited[0].lock)` around the primary-slot test-and-set makes two concurrent probes race to claim the same slot. CLAUDE.md is explicit: per-camera mutable slot state (`ds5_primary`, the streaming flags) must be held under `struct ds5_dev::lock`. This is unrelated to the I2C-less goal — please restore it.

kernel/realsense/d4xx.c
```diff
@@ -6353,7 +6383,9 @@ static int ds5_probe(struct i2c_client *c
 		goto e_chardev;
 	}
 
+#ifndef I2C_LESS
```
ymodlin, line 6386: **BUG** — with `I2C_LESS` the read is skipped but `rec_state` is not reset, so it still holds the device type written by `ds5_wait_device_type()` at line 6378, and `ret` is that call's status. If the retained value happens to equal `DS5_DFU_MAGIC_LSW`, probe takes the DFU-recovery early return and no sensors register. Set `rec_state = 0` in the `#else` branch.

ymodlin (commented):
Automated review. Understood that this is a D500 bringup aid rather than a shipping feature — comments are about what would break if it landed as-is on `dev`.

PR #443, opened 2026-05-24
[RSDSO-21522] Advantech Thor 6 cameras

ejgoldik:


ymodlin (commented):
Review of the multi-camera-per-dser changes in this PR (`ser_primary` rename + new `dser_primary` field and gating). Four inline comments below. Two larger concerns (teardown asymmetry around the existing `ds5_release_slot` / `power_off` path; the 2-second retry loop holding `serdes_lock__`) apply to code that's already on `dev` (not in this PR's current diff), so I'll raise those separately rather than as inline blockers here.

Recommendation: small follow-ups before merge — fix the typo / step rendering, and please consider the `atomic_inc(reset_gen)` and defensive-reset suggestions below.

kernel/realsense/d4xx.c
```diff
@@ -4063,23 +4065,26 @@ static int ds5_gmsl_serdes_setup(struct ds5 *state)
...
-	if (err) {
-		dev_err(dev, "gmsl deserializer link config failed\n");
-		goto error;
+		dev_dbg(dev, "Setup SERDES addressing and control pipeline\n");
+		/* setup serdes addressing and control pipeline */
+		err = state->dser_ops->setup_link(state->dser_dev, &state->client->dev);
+		if (err) {
+			dev_err(dev, "gmsl deserializer link config failed\n");
+			goto error;
+		}
```
ymodlin, line 4071: **Missing `atomic_inc(&state->ds5_dev->dser_control->reset_gen);`** — this PR adds the `dser_primary` gate around a full dser power-cycle (`power_off → power_on → setup_link` here at 4069-4085), but nothing in the tree ever bumps `dser_control->reset_gen`. The streaming-start path at `d4xx.c:4902-4919` reads exactly that counter to detect dser re-init and invalidate per-sensor streaming/config state — so the detection logic is dead code today.

After `setup_link` here (and symmetric to `ds5_get_reset_gen()` which IS incremented at `d4xx.c:2561`), suggest:
```c
msleep(100);
atomic_inc(&state->ds5_dev->dser_control->reset_gen);
```
This closes the loop with the existing `reset_ref_dser` invalidation logic and matters even more once multiple cameras share a dser.

kernel/realsense/d4xx.c
```diff
@@ -3701,6 +3702,7 @@ static int ds5_setup_and_link(struct ds5 *state)
 						if (NULL == dser_inited[j].dser_dev) {
 							dser_inited[j].dser_dev = state->dser_dev;
 							state->ds5_dev->dser_control = &dser_inited[j];
+							state->dser_primary = true;
```
ymodlin, line 3692: **`state->dser_primary` is set here but never reset.** On an EPROBE_DEFER retry (e.g. peer not ready at `d4xx.c:3669`, or `max9295_setup_control` timing out further down), the same `struct ds5 *state` re-enters `ds5_setup_and_link()` carrying the previous attempt's flags. `state->ds5_dev` is defensively cleared at function entry (around `d4xx.c:3655`), but `state->ser_primary` / `state->dser_primary` are not.

Current code paths happen to overwrite them correctly, but this branch only sets `dser_primary = true` — it never clears it. A future refactor that re-enters this function without hitting the `if (NULL == dser_inited[j].dser_dev)` arm could leave a stale `dser_primary=true`, which would then re-run the heavy dser bring-up (power_off / power_on / setup_link) at `d4xx.c:4068-4086` on a dser someone else is actively using.

Suggest defensive reset at the top of `ds5_setup_and_link()`:
```c
state->ds5_dev = NULL;
state->ser_primary = false;
state->dser_primary = false;
```

README_JP7.md
```diff
@@ -99,7 +99,21 @@ sudo sed -i 's/search updates/search extra updates kernel/g' /etc/depmod.d/ubunt
...
+4. Select the correct overlay for your HW:
+
+    Currently supported overlays are -
+
+    | Overlay file | Description |
+    |---|---|
+    | `tegra264-camera-d4xx-overlay.dtbo` | max9296 deserializer board w/ one camera |
+    | `tegra264-camera-d4xx-overlay-max96712-EVB.dtbo` | max96712 evaluation board w/ one camera |
+    | `tegra264-camera-d4xx-overlay-advantech.dtbo` | Advantech board w/ one camera on the bottom right of the left port (i2c9) |
+    | `tegra264-camera-d4xx-overlay-advantech-cams-0-1-2-3.dtbo` | Advantech board w/ four cameras on the left port (2c9) |
```
ymodlin: Typo: `(2c9)` → `(i2c9)`. Other rows in the same table say `(i2c9)` / `(i2c12)`.

README_JP7.md
```diff
@@ -99,7 +99,21 @@ sudo sed -i 's/search updates/search extra updates kernel/g' /etc/depmod.d/ubunt
...
+    | `tegra264-camera-d4xx-overlay-advantech-cams-4-5.dtbo` | Advantech board w/ two cameras on the bottom right and top right of the right port (i2c12) |
+    | `tegra264-camera-d4xx-overlay-advantech-cams-4-5-6-7.dtbo` | Advantech board w/ four cameras on the right port (i2c12) |
+    | `tegra264-camera-d4xx-overlay-advantech-cams-0-1-2-3-4-5.dtbo` | Advantech board w/ six cameras - four on the left port (i2c9) + two on the bottom right and top right of the right port (i2c12) |
+
+5.
```
ymodlin: Markdown nit (pre-existing — was `4.` before): the step number is on its own line and renders awkwardly. Since this PR is editing the exact step boundary, easy to fix inline:
```
5. Verify bootloader configuration
```

kernel/realsense/d4xx.c
ejgoldik: To be honest, I have no idea why we have the reset_gen logic at all. As far as I'm aware it's AI slop and should be completely removed.

kernel/realsense/d4xx.c
ejgoldik: Done

kernel/realsense/d4xx.c
ejgoldik: Removed

README_JP7.md
ejgoldik: Fixed

README_JP7.md
ejgoldik: Fixed

PR #452, opened 2026-05-28
RSDSO-20101: fix BACKEND_TIMESTAMP > TIME_OF_ARRIVAL on GMSL

ymodlin:
## Summary

- VI5's `vi5_capture_dequeue` stamped `v4l2_buffer.timestamp` with a hardware SOF stamp (`descr->status.sof_timestamp`) while the queue advertised `V4L2_BUF_FLAG_TSTAMP_SRC_EOF`. On D457/D401 GMSL the SOF value, after librealsense's `monotonic_to_realtime()`, could land later than the user callback's `system_clock::now()` — violating `BACKEND_TIMESTAMP <= TIME_OF_ARRIVAL`. Not reproducible on USB (uvcvideo stamps at URB completion).
- Replace the SOF stamp with `ktime_get_ns()` at frame-done. Guarantees CLOCK_MONOTONIC + strictly precedes the userspace callback's wall-clock sample, so `BACKEND <= TOA` holds. Existing `V4L2_BUF_FLAG_TSTAMP_SRC_EOF` advertisement is now semantically correct. Hardware SOF time remains available via the per-frame FW `hwTimestamp` (parsed by librealsense into `RS2_FRAME_METADATA_FRAME_TIMESTAMP`).

## Scope: all JP5.x, JP6.x and JP7.x

The base NVIDIA `vi5_capture_dequeue` sets `descr->status.sof_timestamp` on **every** supported generation, independent of the RealSense `0001-*` patch, so the bug mechanism exists on JP5.x too. This PR therefore fixes JP5.x as well as JP6.x/7.x. (This corrects an earlier claim that the fix was JP6.x/7.x-only.)

### Patch layout

**JP5.x — `kernel/nvidia/`, standalone `RSDSO-20101-backend-timestamp-gmsl.patch` (real files):**
- `5.0.2/0022-RSDSO-20101-backend-timestamp-gmsl.patch`
- `5.1.2/0003-RSDSO-20101-backend-timestamp-gmsl.patch` (identical to 5.0.2)
- `5.1.6/0003-RSDSO-20101-backend-timestamp-gmsl.patch` (different surrounding context — `if (frame_err)` block, no Y12I hunk)

**JP6.x/7.x — `nvidia-oot/`, currently folded into `Fix-y12i-calibration-stream.patch`:**
- `6.0/0004-Fix-y12i-calibration-stream.patch` — real file (source of truth for the 6.x tree)
- `6.1/0004`, `6.2/0004` — symlinks → `6.0/0004`
- `6.2.1`, `6.2.2` — directory symlinks → `6.2`
- `7.0/0003-Fix-y12i-calibration-stream.patch` — **real file**, byte-identical to 6.0 (a duplicate, not a symlink; can drift from 6.0)
- `7.1/0003`, `7.2/0003` — symlinks → `7.0/0003`

> **Follow-up (tracked):** on JP6.x/7.x the timestamp change currently rides inside the unrelated `Fix-y12i-calibration-stream.patch`. Once the Y12I calibration-stream fix is validated it will be split out into its own standalone `RSDSO-20101-*` patch, matching the JP5.x layout. Making `7.0` a symlink to `6.0` (they are byte-identical) is also on the list to avoid future drift.

## Test plan

- [x] Built `./build_all.sh 6.0` on a clean tree — passes.
- [x] Deployed to `fw-orin-2` (Jetson AGX Orin Dev Kit, JP6.0, L4T R36.3.0, kernel 5.15.136-tegra).
- [x] Streamed depth + color at 848x480@30fps via pyrealsense2 (`.triage/RSDSO-20101/check_timestamps.py`). Result: 290 frames, **0 violations** of `BACKEND_TIMESTAMP <= TIME_OF_ARRIVAL`. Average `TOA - BACKEND` ~0.3 ms (depth) / ~0.15 ms (color); max 7 ms / 1 ms; min 0 ms (expected ms-rounding boundary).
- [x] Validated on JP5.x — both Orin and Xavier (the ticket's original JP5.1.2 / Xavier target).


sysrsbuild-gh:
Can one of the admins verify this patch?

kernel/nvidia/5.1.2/0004-backend-timestamp-gmsl.patch
```diff
@@ -0,0 +1,21 @@
...
+-	vb->vb2_buf.timestamp = descr->status.sof_timestamp;
++	/* RSDSO-20101: stamp at frame-done with CLOCK_MONOTONIC so
++	 * librealsense BACKEND_TIMESTAMP (= monotonic_to_realtime(buf.timestamp))
++	 * is by construction <= TIME_OF_ARRIVAL (= system_clock::now() at callback).
++	 * The buffer queue flag advertises V4L2_BUF_FLAG_TSTAMP_SRC_EOF, which
++	 * now matches the value semantics. Hardware SOF time remains
++	 * available to userspace via the per-frame FW hwTimestamp in the MIPI
++	 * metadata payload (-> RS2_FRAME_METADATA_FRAME_TIMESTAMP).
++	 */
++	vb->vb2_buf.timestamp = ktime_get_ns();
```
ipilchin1, line 18: **Scope contradiction with the PR description — must reconcile.**

The description says *"this fix is JP6.x/7.x-only"* and *"JP5.x reproduction needs a separate analysis"* — but this file (plus the 5.0.2 and 5.1.6 ones) changes the JP5.x timestamp to `ktime_get_ns()`.

I checked the base source: JP5.1.2's `vi5_capture_dequeue` already stamps `descr->status.sof_timestamp` (`vi5_fops.c:702`), same on 5.0.2 (`:674`). So the bug mechanism **does** exist on JP5.x and these patches are probably correct — which means the description's reasoning ("the `0001` RS patch doesn't touch the timestamp → 5.x unaffected") is the flaw: the *base NVIDIA code* sets `sof_timestamp` regardless of the RS patch.

Please reconcile: either drop the JP5.x patches (if genuinely out of scope) or update the description and validate them. Right now the code and the narrative disagree.

My verdict: **must-fix** (scope / narrative mismatch).

kernel/nvidia/5.1.2/0004-backend-timestamp-gmsl.patch
```diff
@@ -0,0 +1,21 @@
+diff --git a/drivers/media/platform/tegra/camera/vi/vi5_fops.c b/drivers/media/platform/tegra/camera/vi/vi5_fops.c
```
ipilchin1, line 1: **Patch-number collision — should-fix, merge hazard.**

This file is `0003-RSDSO-20101-backend-timestamp-gmsl.patch`, but `kernel/nvidia/5.1.2/0003-vi-channel-verify-s-g-callbacks.patch` **already exists on master** — so merging this creates two `0003-*` patches in the same directory. The 5.1.6 copy has the same problem against **PR #490**, which independently adds `5.1.6/0003-vi-channel-verify-s-g-callbacks.patch`.

They happen to apply (the two `0003`s touch different files — `vi5_fops.c` vs `channel.c`), so it's not a build break, but it violates the sequential-numbering convention and is a rebase/merge hazard. 5.0.2 correctly got a unique number (`0022`); 5.1.2 and 5.1.6 should likewise take the next free number (e.g. `0011`).

My verdict: **should-fix** (renumber before merge).

nvidia-oot/6.0/0004-Fix-y12i-calibration-stream.patch
```diff
@@ -2,9 +2,18 @@ diff --git a/drivers/media/platform/tegra/camera/vi/vi5_fops.c b/drivers/media/p
 index 7d07ccd7..c64dcc38 100644
 --- a/drivers/media/platform/tegra/camera/vi/vi5_fops.c
 +++ b/drivers/media/platform/tegra/camera/vi/vi5_fops.c
-@@ -582,7 +641,10 @@ static void vi5_capture_dequeue(struct tegra_channel *chan,
+@@ -582,7 +641,18 @@ static void vi5_capture_dequeue(struct tegra_channel *chan,
  	trace_tegra_channel_capture_frame("sof", &ts);
- 	vb->vb2_buf.timestamp = descr->status.sof_timestamp;
+-	vb->vb2_buf.timestamp = descr->status.sof_timestamp;
++	/* RSDSO-20101: stamp at frame-done with CLOCK_MONOTONIC so
```
ipilchin1, line 8: **Timestamp fix folded into the unrelated Y12I patch — should-fix, traceability.**

On JP6.x/7.x the RSDSO-20101 change is merged into `Fix-y12i-calibration-stream.patch` — this single hunk now does both the timestamp change *and* the Y12I `frame_err` workaround. On JP5.x it's a clean standalone `RSDSO-20101-backend-timestamp-gmsl.patch`.

Folding two unrelated fixes into one patch hurts `git blame` / revert / bisect, and it contradicts the description's claim that this is a separate symlinked patch. Please split it into its own `RSDSO-20101-…` patch for 6.x/7.x too, matching the 5.x layout.

My verdict: **should-fix** (separate the concerns).

kernel/nvidia/5.1.6/0004-backend-timestamp-gmsl.patch
```diff
@@ -0,0 +1,21 @@
...
+-	vb->vb2_buf.timestamp = descr->status.sof_timestamp;
++	/* RSDSO-20101: stamp at frame-done with CLOCK_MONOTONIC so
++	 * librealsense BACKEND_TIMESTAMP (= monotonic_to_realtime(buf.timestamp))
++	 * is by construction <= TIME_OF_ARRIVAL (= system_clock::now() at callback).
++	 * The buffer queue flag advertises V4L2_BUF_FLAG_TSTAMP_SRC_EOF, which
++	 * now matches the value semantics. Hardware SOF time remains
++	 * available to userspace via the per-frame FW hwTimestamp in the MIPI
++	 * metadata payload (-> RS2_FRAME_METADATA_FRAME_TIMESTAMP).
++	 */
++	vb->vb2_buf.timestamp = ktime_get_ns();
```
ipilchin1, line 18: **Untested variant — should-fix, coverage.**

The test plan validated only `build_all.sh 6.0` + fw-orin-2 (JP6.0). This 5.1.6 patch — which has a *different* surrounding context (`if (frame_err)` below, no Y12I block) — isn't built or run anywhere, and neither are 5.0.2 / 5.1.2 / 7.0 / 7.1 / 7.2. The ticket's original target (Xavier / JP5.1.2) in particular is unvalidated.

Please at least build-test each shipped variant; ideally reproduce the original Xavier `BACKEND > TOA` violation and confirm it's gone.

My verdict: **should-fix** (validate the shipped variants).

ipilchin1 (commented):
Full review of RSDSO-20101. Each finding is a **separate inline comment** below, ordered must-fix → should-fix → nit, each tagged with my verdict.

The core `descr->status.sof_timestamp` → `ktime_get_ns()` change is technically sound: `sof_timestamp` isn't in the `CLOCK_MONOTONIC` domain librealsense's `monotonic_to_realtime()` assumes, so it could convert to a wall-clock after the callback; `ktime_get_ns()` is `CLOCK_MONOTONIC` and matches the advertised `V4L2_BUF_FLAG_TSTAMP_SRC_EOF`, so `BACKEND <= TOA` holds. Hunk arithmetic and symlink targets check out. The notes below are about scope, patch packaging, numbering, and test coverage.

nvidia-oot/7.0/0003-Fix-y12i-calibration-stream.patch
```diff
@@ -2,9 +2,18 @@ diff --git a/drivers/media/platform/tegra/camera/vi/vi5_fops.c b/drivers/media/p
 index 7d07ccd7..c64dcc38 100644
 --- a/drivers/media/platform/tegra/camera/vi/vi5_fops.c
 +++ b/drivers/media/platform/tegra/camera/vi/vi5_fops.c
-@@ -582,7 +641,10 @@ static void vi5_capture_dequeue(struct tegra_channel *chan,
+@@ -582,7 +641,18 @@ static void vi5_capture_dequeue(struct tegra_channel *chan,
  	trace_tegra_channel_capture_frame("sof", &ts);
- 	vb->vb2_buf.timestamp = descr->status.sof_timestamp;
+-	vb->vb2_buf.timestamp = descr->status.sof_timestamp;
++	/* RSDSO-20101: stamp at frame-done with CLOCK_MONOTONIC so
```
ipilchin1, line 8: **Description's symlink list is inaccurate + 7.0 is a real duplicate — nit.**

The description says the patch is *"symlinked from 6.1/, 6.2/, 6.2.1/, 6.2.2/, 7.0/, 7.1/"*. But on the branch **7.0 is a real file** (mode 100644), not a symlink — it duplicates 6.0's content byte-for-byte (both blob `5c8620a7`), so it had to be hand-edited and can silently drift from 6.0 later. The list also omits **7.2** (a symlink → 7.0).

Fix the description; and since 6.0 and 7.0 are identical, consider making 7.0 a symlink to 6.0 to avoid future drift.

My verdict: **nit** (accuracy / maintainability).

nvidia-oot/6.0/0004-Fix-y12i-calibration-stream.patch
```diff
@@ -2,9 +2,18 @@ diff --git a/drivers/media/platform/tegra/camera/vi/vi5_fops.c b/drivers/media/p
...
++	 * librealsense BACKEND_TIMESTAMP (= monotonic_to_realtime(buf.timestamp))
++	 * is by construction <= TIME_OF_ARRIVAL (= system_clock::now() at callback).
++	 * The buffer queue flag advertises V4L2_BUF_FLAG_TSTAMP_SRC_EOF, which
++	 * now matches the value's semantics. Hardware SOF time remains
```
ipilchin1, line 12: **Comment wording differs across trees — nit, cosmetic.**

Here the comment reads *"now matches the value's semantics"*; the 5.x patches say *"matches the value semantics"*. Harmless, but worth unifying since these are meant to be the identical fix.

My verdict: **nit** (cosmetic).

kernel/nvidia/5.1.2/0004-backend-timestamp-gmsl.patch
ymodlin: I validated also on JP5.x

nvidia-oot/6.0/0004-Fix-y12i-calibration-stream.patch
ymodlin: the patch name is wrong and the fix of y12 should be removed anyway. this will be fixed once we finish to validate that y12 is fixed.

kernel/nvidia/5.1.6/0004-backend-timestamp-gmsl.patch
ymodlin: Orin and xavier were validated.

kernel/nvidia/5.1.2/0004-backend-timestamp-gmsl.patch
ymodlin: will fix

ipilchin1 (approved):
Blocking items resolved — approving.

- #1 (scope): description corrected to "all JP5.x/6.x/7.x", now matching the code. ✅
- #2 (numbering collision): `5.1.2`/`5.1.6` renumbered `0003 → 0004`; verified `0004` is free on master and doesn't clash with PR #490's `0003`. ✅
- #4: the ticket's original Xavier / JP5.1.2 target is now validated. ✅
- Core `sof_timestamp → ktime_get_ns()` fix is sound.

Non-blocking follow-ups (fine to handle later):
- #3: timestamp change still rides inside `Fix-y12i-calibration-stream.patch` on 6.x/7.x — tracked for split-out once Y12I is validated.
- JP7.x (7.0/7.1/7.2) not yet listed as validated.
- The Scope section still cites the old `0003-RSDSO-20101-*` names; files are now `0004-backend-timestamp-gmsl.patch` — worth syncing the description.

LGTM.

kernel/nvidia/5.1.6/0004-backend-timestamp-gmsl.patch
sareluzi, line 18: @ymodlin - the path filename is not really corresponding to this fix..

sareluzi: NVM; I see it was already dicussed. resolving

PR #453, opened 2026-05-28
ext_sync_gen.py - Allow enable after enable

ejgoldik:
Allow changing ext_sync_gen pwm without disabling in between so this input is now legal:
python ext_sync_gen.py --enable --fps 30
python ext_sync_gen.py --enable --fps 60
(Previously the second command would have failed on 'resource busy')


ymodlin (approved):
external script you can merge this

PR #458, opened 2026-06-07
nvidia-oot: reorganize JP6.x patches by subject + fix embedded-metadata blocker bugs

ymodlin:
## Summary

Two related changes to the JetPack 6.x `nvidia-oot/` patch set:

1. **Reorganize JP6.x patches by subject** (`118bcb3`) — split the old layout (one mega "Porting" patch + scattered SerDes follow-ups) into subject-grouped patches:
   - `0001-embedded-metadata` — Tegra VI/camera-core changes + `d4xx.o` build wiring
   - `0002-max9295-max9296-SerDes` — GMSL serializer+deserializer pair (squashes old 0004/0005/0007/0008/0009)
   - `0003-Adding-max96712`
   - `0004-Fix-y12i-calibration-stream` (unchanged)
   - `0005-Runtime-tsc-rate-config` (unchanged)

   The split is a pure regrouping of existing hunks: applying the new set reproduces a **byte-identical source tree** to the old set on every JP6.x baseline (verified against L4T 36.3/36.4/36.4.3/36.4.4). Preserves the repo's per-file symlink dedup (6.0 canonical; 6.1/6.2 symlink shared patches; 6.2.1/6.2.2 dir-symlink 6.2). JP7 (7.0/7.1) left unchanged.

2. **Fix embedded-metadata upstream-review blocker bugs** (`a4e01fc`) in the `0001-embedded-metadata` patch:
   - **1.1** single, idempotent owner of embedded teardown (no double-free / use-after-free)
   - **1.2/1.3** NULL-guard + cleanup of the embedded node on graph error paths (no NULL-deref / leak)
   - **1.4** return ring-full metadata buffers to vb2 with `VB2_BUF_STATE_ERROR` (no buffer leak / DQBUF hang)
   - **1.5** real `S_FMT`/`TRY_FMT` for the metadata node (V4L2 API compliance)
   - proper `init_video_embedded()` error ladder; removed commented-out dead code

## Verification

- **Byte-identical-tree proof** for the refactor on all 4 available L4T baselines, plus independent multi-agent review (equivalence / integration / kernel-safety).
- **Builds clean** for JP6.0 and JP6.2 (d4xx + max9295/96/712 modules).
- **Blocker-fix kernel-safety review: SAFE** — no double-free/UAF/NULL-deref/deadlock/leak; steady-state streaming path unchanged.
- **Deployed to fw-orin-2** (AGX Orin Devkit, JP6.0, D457 via max96712) and regression-tested:
  - Camera discovered (DEPTH/RGB/Y8/IMU, FW 5.17.3.7), driver v1.0.4.1
  - V4L2 streaming: depth/IR **30.02 fps**, color **29.99 fps**
  - Metadata node streams (30 × 68-byte payloads); `G_FMT` correct
  - `rmmod`/`modprobe` teardown cycle clean — **0 crash signatures** in dmesg
  - No regression


ejgoldik (approved):
Looks great

ejgoldik (changes requested):
JP7 patching broken because of the patch links

nvidia-oot/6.0/0001-embedded-metadata-support-for-D4xx.patch
```diff
@@ -830,12 +486,19 @@ index 91f8d5f8..d9be2677 100644
...
 +				chan->emb_buf_addr, chan->emb_buf);
 +		chan->emb_buf_size = 0;
-+		vb2_queue_release(queue);
++	}
++
++	/*
++	 * The vb2 queue and DMA alloc ctx are initialized unconditionally in
++	 * tegra_channel_init_video_embedded(), so release them based on the node
++	 * having been created (video != NULL above), independent of emb_buf_size.
++	 * This function is the single owner of embedded teardown.
```
sareluzi, line 493: **Latent coupling — `emb_buf` free now gated on `video != NULL`.** The `if (!video) return -EINVAL` guard at the top of this function sits *before* the `emb_buf` free block. `tegra_channel_cleanup` no longer frees `emb_buf` directly — it delegates here. Safe today because `emb_buf` is only ever allocated on the embedded path (`vi5_release_buffer` gates on `embedded_data_height == 1`, which implies an embedded node exists, so `video != NULL`). But `emb_buf` lifetime is now implicitly tied to `embedded.video`. If a future path ever allocates `emb_buf` without an embedded node, it leaks silently.

Cheap hardening: free `emb_buf` *before* the `if (!video)` guard, or add a one-line comment at the `emb_buf` alloc site documenting the invariant.

nvidia-oot/6.0/0001-embedded-metadata-support-for-D4xx.patch
```diff
@@ -603,15 +236,35 @@ index 91f8d5f8..d9be2677 100644
...
-+   return 0;
++	struct v4l2_fh *vfh = file->private_data;
++	struct v4l2_meta_format *fmt = &format->fmt.meta;
++
++	if (format->type != vfh->vdev->queue->type)
++		return -EINVAL;
++
++	memset(fmt, 0, sizeof(*fmt));
++	fmt->dataformat = V4L2_META_FMT_D4XX;
++	fmt->buffersize = 255;
```
sareluzi: Nit: `255` (buffersize) is now hardcoded across `get_format`/`set_format`/`try_format`/`queue_setup`/`buf_prepare`/the `vi5` `memcpy`, and `68` (payload) in `vi5_release_metadata_buffer`. Worth two `#define`s (e.g. `D4XX_META_BUFSIZE 255`, `D4XX_META_PAYLOAD 68`) so buffer size and payload can't drift apart across the call sites.

Also: `set_format` and `try_format` are byte-identical and ignore the incoming format (always force `V4L2_META_FMT_D4XX`/255) — acceptable for a fixed-format meta node, but they could share one helper.

sareluzi (commented):
Reviewed the embedded-metadata blocker fixes (the real code change) plus the patch-refactor mechanics.

**The blocker fixes are correct** — ring-full → `vb2_buffer_done(VB2_BUF_STATE_ERROR)` (lock released on both branches), the `init_video_embedded` error ladder (labels ordered, `embedded.video = NULL` on every error exit), the idempotent single-owner `cleanup_video_embedded` (`if (!video) return -EINVAL` guards the double-call), and the graph.c error-path routing all check out.

**Refactor plumbing verified:** `0001` owns `drivers/media/i2c/Makefile`, `0002` owns max9295/96 — no file-ownership overlap under `apply_patches.sh`'s `git apply <dir>/*`. Symlink dedup intact (`6.0/` real, `6.1`/`6.2` symlink byte-identical, `6.2.1`/`6.2.2` → `6.2`). No stale patch-name refs in scripts/docs; `external-sync-fg12-16ch.md` renumbered `0010`→`0005` for 6.x correctly.

_Not independently re-run: the byte-identical-tree claim across the 4 L4T baselines — trusting the PR verification + CI there._

Two inline notes below. Two more nits that fall on non-diffable lines:
- **Double semicolon** `chan->embedded.tail = 0;;` in `tegra_metadata_stop_streaming`.
- **`memcpy(frm_buffer, chan->emb_buf_addr, 255)`** in `vi5_release_metadata_buffer`: destination is bounded (`queue_setup`/`buf_prepare` reject planes `< 255`), so no plane overflow. Source assumes `emb_buf_addr` ≥ 255 bytes — fine for D4XX line width. The `255` here vs `68` payload on the next line is the magic-number drift flagged inline.

Recommend **approve** once the `emb_buf` coupling note is fixed or acknowledged.

nvidia-oot/6.0/0001-embedded-metadata-support-for-D4xx.patch
ymodlin: Done in c8f751c — moved the `emb_buf` free **above** the `if (!video)` guard in `tegra_channel_cleanup_video_embedded()`, so emb_buf release no longer depends on `embedded.video` (with a comment documenting the decoupling). `tegra_channel_cleanup()` continues to delegate here. Thanks for the catch.

nvidia-oot/6.0/0001-embedded-metadata-support-for-D4xx.patch
ymodlin: Done in c8f751c — added `D4XX_META_BUFSIZE` (255) and `D4XX_META_PAYLOAD` (68) in `mc_common.h` and used them across `get/set/try_format`, `queue_setup`, `buf_prepare` and the vi5 `memcpy`/payload, so the two can't drift apart. Also collapsed the byte-identical G/S/TRY_FMT handlers into a shared `tegra_metadata_fill_format()`. (And dropped the stray `;;` in `tegra_metadata_stop_streaming()` you noted.)

sareluzi (commented):
## Re-review of follow-up commit `c8f751c`

Verified all three actionable findings are addressed correctly, in **both** the `6.0` and `6.2` variants of `0001-embedded-metadata`:

- **#1 emb_buf decoupling** ✅ — `tegra_channel_cleanup_video_embedded()` now frees `emb_buf` *before* the `if (!video) return -EINVAL` guard. `vi_unit_dev` is resolved at function top, so the early free is valid. Idempotency preserved: second call sees `emb_buf_size == 0` (skip) and `video == NULL` (return) — no double-free, no double `dma_free_coherent`.
- **#2 magic numbers** ✅ — `D4XX_META_BUFSIZE` (255) / `D4XX_META_PAYLOAD` (68) defined in `mc_common.h` with a rationale comment, applied across `queue_setup`/`buf_prepare`/the format helper/the vi5 per-frame copy. `vi5_fops.c` and `channel.c` both pull in `mc_common.h`, so the defines resolve. G/S/TRY_FMT collapsed into the shared `tegra_metadata_fill_format()` — clean, behavior identical (still validates `format->type`).
- **#3 double semicolon** ✅ — `chan->embedded.tail = 0;` fixed.

The vi5 `memcpy` bound note from the prior review was informational only (destination still guarded `>= D4XX_META_BUFSIZE` by `queue_setup`/`buf_prepare`); now self-documenting via the define.

**LGTM.** No new issues introduced by the follow-up. Same standing caveat as before: I did not independently re-run the byte-identical-tree proof or CI — confirm the JP6.0/6.2 build workflow is green before merge.

nvidia-oot/7.0/0007-Allow-external-address-assignment-on-max9295.patch
sareluzi, line 1: why duplicate the patch instead of link like before

ymodlin: I didn't change 7.0/7.1 they still have the old patches

ymodlin: I need the HW to check for 7.0 before we merge. I think this should be a different PR

ejgoldik (dismissed):

sareluzi (dismissed):

ymodlin:
✅ **Validated on `fw-orin-nano-1`** (rebased HEAD `865aff5`)

**Target:** Jetson Orin Nano (Seeed reComputer), JetPack 6.2 / L4T 36.4.3, dual-camera (`dev_dual` / seeed-cams-0-1 overlay).

| Stage | Result |
|---|---|
| Rebase onto `dev` | Clean (1 doc conflict in `CLAUDE.md` resolved) |
| `apply_patches.sh 6.2` | Reorganized nvidia-oot patches apply cleanly |
| `build_all.sh 6.2` | Clean — `d4xx.ko` 1.0.4.6 + `max9295/max9296/max96712.ko`, vermagic `5.15.148-tegra` |
| Deploy + reboot | OK — driver upgraded 1.0.4.4 → 1.0.4.6 |
| Probe | Clean: DEPTH / RGB / Y8 / IMU on both cameras (FW 5.17.3.23), no dmesg errors |
| Streaming | 30 fps on depth-0, depth-1, ir-0, color-0 |

No regressions observed; the JP6.x patch reorganization builds and runs correctly on hardware.

ejgoldik (approved):

PR #463, opened 2026-06-09
[RSDEV-11964]: d4xx: re-probe camera after in-recovery DFU to re-enumerate without reboot

ymodlin:
## RSDEV-11964 — D457 GMSL fails to enumerate after a successful in-recovery FW update

**Bug (P1):** On GMSL/MIPI, if a FW update is interrupted (power cut) the camera enters recovery, and after a host reboot it is bound by `d4xx` in `DS5_DFU_RECOVERY` (chardev only, no V4L2 nodes). A subsequent **successful** in-recovery update reports success, but the camera **never re-enumerates** — it stays pinned in the recovery binding until `rmmod d4xx && modprobe d4xx` or a host reboot. There is no GMSL hotplug, so nothing re-probes it.

**Root cause:** after `DS5_DFU_DONE` the driver never transitioned the instance out of recovery — the `ds5_v4l_init()` re-init was left commented out (`init_v4l_f` "need refactoring"), and even if enabled it only covered the depth instance (the RGB/Y8/IMU peers have no chardev/release hook) and didn't wait for the FW reboot.

## Fix

Schedule a `delayed_work` at `DS5_DFU_DONE` (in the DFU chardev release) that:
1. waits (bounded, polling the DFU magic reg) for the camera to reboot into application FW,
2. discovers the camera's four sibling i2c clients via `driver_for_each_device` (filtered by shared `ds5_dev`),
3. re-probes them — `device_release_driver` (peers → primary) then `device_attach` (primary → peers, the order GMSL SERDES requires) — so they leave recovery and recreate their V4L2 nodes.

This automates the proven `unbind`/`bind` recovery in-driver. Gated to the recovery case only (`!mux.sd.subdev.v4l2_dev`), so normal in-place updates are untouched. Lifecycle: static work slots, `INIT_DELAYED_WORK`/`cancel_delayed_work_sync` via new `module_init`/`module_exit` (converted from `module_i2c_driver`); device refs held across the deferral.

## Validation (fw-orin-nano-1, JP6.2 / L4T R36.4.3, d4xx 1.0.4.0)

A/B on a real D457 GMSL:
- **Baseline (stock):** after a successful in-recovery update the camera stays a ghost — `D4XX recovery state`, no nodes, absent from `rs-enumerate-devices` — until manual rebind/reboot.
- **Fixed:** the camera **auto-re-enumerates** with no manual rebind and no host reboot. dmesg:
  ```
  d4xx 2-002a: D4XX remove DEPTH
  d4xx 2-002a: Probing driver for D4xx
  d4xx 2-002a: D4XX Sensor: DEPTH, firmware build: 5.17.3.7
  ... 2-002b RGB / 2-002c Y8 / 2-002d IMU likewise
  ```
  `rs-enumerate-devices` lists the D457 again; video nodes recreated. The fix was also independently exercised on a second camera (D401) on the same rig.

## Notes / follow-ups
- A companion **librealsense** hardening is recommended (the MIPI recovery-update path reports success and prints a "reboot / reload d4xx" hint instead of waiting for reconnect) — separate change; the SDK cannot re-enumerate MIPI on its own, so this driver change is the load-bearing fix.
- Separate robustness observation: entering MIPI DFU on an **operational** GMSL camera reboots the host — worth its own investigation, unrelated to this fix.

### DIFF
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -44,6 +44,7 @@
 #include <media/max9296.h>
 #include <media/max96712.h>
 
+
 /* Deserializer interface structure for abstraction */
 struct dser_interface {
 	/* Pipeline management */
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -17,6 +17,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+
 #include <linux/delay.h>
 #include <linux/gpio.h>
 #include <linux/i2c.h>
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -31,6 +31,7 @@
 #include <linux/videodev2.h>
 #include <linux/version.h>
 #include <linux/mutex.h>
+#include <linux/workqueue.h>
 #include <media/media-entity.h>
 #include <media/v4l2-ctrls.h>
 #include <media/v4l2-device.h>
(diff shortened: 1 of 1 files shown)


sareluzi (commented):
Code review — design is solid (gating, ref balance on the normal path, and the peers→primary / primary→peers unbind-bind order all check out). Three findings inline: one convention must-fix, one latent refcount leak, one edge-case note.

kernel/realsense/d4xx.c
```diff
@@ -6221,6 +6224,213 @@ static int ds5_v4l_init(struct i2c_client *c, struct ds5 *state)
...
+
+	/* 1. Wait for the camera to leave the bootloader and serve app FW again.
+	 *    Mirror probe's recovery test: operational == magic reg no longer
+	 *    returns the DFU magic.  Bounded so a dead device degrades to the
+	 *    pre-fix behaviour (a warning telling the user to reload/reboot).
+	 */
+	deadline = jiffies + msecs_to_jiffies(DS5_DFU_REBIND_TIMEOUT_MS);
+	do {
+		if (depth_state) {
+			ret = ds5_read(depth_state, DS5_DFU_MAGIC_REG, &magic);
```
sareluzi, line 6323: **Convention must-fix:** this poll loop must use `ds5_read_poll()`, not `ds5_read()`.

Per CLAUDE.md: *"For polling loops expecting transient I2C failures (HWMC status checks, reset readiness polls, DFU timeout checks), use `ds5_read_poll()`... Reserve `ds5_read()` for normal I2C operations where retry semantics are desired."*

This loop polls the magic reg while the camera is rebooting into app FW — transient I2C failures are expected on every iteration. `ds5_read()` retries + logs each failure, so a recovery spams the log for up to `DS5_DFU_REBIND_TIMEOUT_MS` (30s).

The one-shot recovery check in `ds5_probe()` (~line 6740) correctly stays `ds5_read` because it is not a loop — only this loop needs the swap.

```c
			ret = ds5_read_poll(depth_state, DS5_DFU_MAGIC_REG, &magic);
```

kernel/realsense/d4xx.c
```diff
@@ -6221,6 +6224,213 @@ static int ds5_v4l_init(struct i2c_client *c, struct ds5 *state)
...
+		mutex_unlock(&ds5_dfu_rebind_lock);
+		dev_warn(&state->client->dev,
+			"%s(): no free re-probe slot; reload d4xx or reboot to enumerate\n",
+			__func__);
+		return;
+	}
+	ctx->pending = true;
+	ctx->ds5_dev = state->ds5_dev;
+	ctx->depth_client = state->client;
+	get_device(&state->client->dev);	/* released in ds5_dfu_rebind_work_fn */
```
sareluzi, line 6425: **Latent refcount leak on slot reuse.** When a slot is already `pending` for this `ds5_dev`, `ctx` is reused but execution still falls through to `get_device()` below — while `schedule_delayed_work()` on an already-queued delayed_work is a no-op (it will not requeue). The single work run does exactly one `put_device()` for the schedule ref, so the second `get_device()` leaks one device reference per double-schedule.

Low probability (needs two DFU completions inside the `DS5_DFU_REBIND_BOOT_DELAY_MS` 2s window) but real. Fix: early-return when reusing an already-pending slot, e.g.

```c
	for (i = 0; i < DS5_DFU_REBIND_SLOTS; i++)
		if (ds5_dfu_rebind_slots[i].pending &&
		    ds5_dfu_rebind_slots[i].ds5_dev == state->ds5_dev) {
			mutex_unlock(&ds5_dfu_rebind_lock);
			return;	/* already scheduled for this camera */
		}
```

or only `get_device()` on the take-a-free-slot branch.

kernel/realsense/d4xx.c
```diff
@@ -6221,6 +6224,213 @@ static int ds5_v4l_init(struct i2c_client *c, struct ds5 *state)
...
+	}
+	return 0;
+}
+
+static void ds5_dfu_rebind_work_fn(struct work_struct *w)
+{
+	struct ds5_dfu_rebind_ctx *ctx =
+		container_of(to_delayed_work(w), struct ds5_dfu_rebind_ctx, work);
+	struct i2c_client *depth = ctx->depth_client;
+	struct ds5 *depth_state = ds5_state_from_client(depth);
```
sareluzi, line 6308: **Edge-case note (non-blocking):** only a device ref on the i2c_client is held across the deferral — not a bound-driver guarantee. A manual sysfs unbind between schedule and work-run frees the devm `ds5` state while `clientdata` still points at it, so `ds5_state_from_client()` here would deref freed memory.

`module_exit` covers `rmmod` via `cancel_delayed_work_sync`, but not a sysfs `unbind`. Worth a guard or at least a comment noting the assumption.

PR #468, opened 2026-06-11
[RSDSO-21614] DFU/recovery robustness: finalize on close, route unsupported device types to DFU-only

Nikolai-L:
**Tracking:** RSDSO-21614 — https://rsjira.realsenseai.com/browse/RSDSO-21614

## Summary
Hardens the D4XX DFU / recovery paths so firmware can always be flashed when the device is reachable over I2C.

### DFU download finalization (the main fix)
The bootloader follows standard size-agnostic USB DFU: a single zero-length `DFU_DNLOAD` (`0x4a04=0`) is the **sole** end-of-transfer marker that drives `DOWNLOAD_IDLE → MANIFEST → flash-commit → reboot`. The old write handler emitted it via a heuristic (only when a `write()` carried a partial `< DFU_BLOCK_SIZE` block), which was wrong in both directions:
- **Stall:** a 1024-aligned image (e.g. 2048×1024 = 2,097,152 B) never produced a partial block → terminator never sent → camera stuck in `DOWNLOAD_IDLE`. `rs-fw-update` and raw `cat`/`dd` all hit this.
- **Premature commit:** userspace buffering (e.g. `std::ofstream`) can split the image so an *intermediate* `write()` is non-aligned → mid-image commit of a truncated image.

Fix: `ds5_dfu_device_write()` is now **transfer-only** (streams blocks, per-block `dfuDNLOAD_IDLE` sync, stays `DS5_DFU_IN_PROGRESS`); `ds5_dfu_device_release()` is the **sole** finalizer, sending the one `0x4a04=0` + manifest at `close()` — the only reliable end-of-image signal. Wire protocol is unchanged; only *when* the terminator fires moves to `close()`.

### Device-type readiness / DFU-only routing
- `ds5_wait_device_type()` gains `require_supported`: reset path still requires an operationally-supported type (confirms GMSL link recovery); probe accepts any FW-populated (non-zero) type.
- `ds5_probe()` routes alive-but-unsupported types (e.g. legacy D430 GS reporting type 3, which reports type 5 once updated) to a DFU-only path instead of failing probe and tearing down `/dev/d4xx-dfu` — so old firmware can still be flashed. The DFU flow needs only I2C (absolute HWMC registers), so the skipped `ds5_v4l_init()` doesn't matter.

CLAUDE.md updated with the finalize-at-close contract, the `ds5_wait_device_type()` contract, and the DFU-only routing rule.

## Backward compatibility
- FW/bootloader: identical traffic, fully compatible.
- Flashers (`rs-fw-update`, `cat`, `dd`): write→close, so they finalize at close; previously-broken aligned-image cases now work.
- No ABI change (no ioctl/struct/device-node changes).

## Testing
- [ ] Build + deploy (done locally by author)
- [ ] Flash a **1024-aligned** image (2048×1024) via `rs-fw-update` and raw `cat` → commits + reboots into new FW (was: stuck in `DOWNLOAD_IDLE`)
- [ ] Non-aligned-chunk writer (`dd bs=999`) → no early/truncated commit
- [ ] `dmesg | grep -i dfu` shows no `DFU finalize failed`
- [ ] Round-trip re-flash back to original FW

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -56,6 +56,8 @@ pytest -vs -m d457 test/                   # Direct pytest invocation
 
 Pytest marker: `d457` (defined in `test/pytest.ini`). Test timeout: 200 seconds.
 
+CI also runs a V4L2 test workflow on a self-hosted Jetson runner when `kernel/realsense/**` or `test/v4l2_test/**` paths change.
+
 ## Architecture
 
 ### Driver stack (top to bottom)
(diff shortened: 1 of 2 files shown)


kernel/realsense/d4xx.c
```diff
@@ -2376,7 +2424,17 @@ static int ds5_set_calibration_data(struct ds5 *state,
 #define DS5_DFU_MAGIC_REG	0x5020
 #define DS5_DFU_MAGIC_LSW		0x0201  /* Lower 16 bits of 0x04030201 */
 
-static int ds5_wait_device_type(struct ds5 *state, u16 *dev_type)
+/*
```
ymodlin, line 2427: This comment is too long. please summeraize.

CLAUDE.md
```diff
@@ -113,6 +115,32 @@ The build system cross-compiles for ARM64. Toolchains vary by JetPack:
 | 6.2     | 36.4.3      | kernel/kernel-jammy-src |
 | 6.2.1   | 36.4.4      | kernel/kernel-jammy-src |
 
+## Coding conventions
+
+### Kernel driver (C — `kernel/realsense/d4xx.c`)
```
ymodlin, line 120: I think this instruction should go into a coding agent

kernel/realsense/d4xx.c
```diff
@@ -6045,14 +6110,15 @@ static ssize_t ds5_dfu_device_write(struct file *flip,
...
 				ret = ds5_dfu_wait_for_get_dfu_status(state, dfuDNLOAD_IDLE);
-			if (!ret)
-				ret = ds5_write(state, 0x4a04, 0x00); /*Download complete */
-			if (!ret)
-				ret = ds5_dfu_wait_for_get_dfu_status(state, dfuMANIFEST);
 			if (ret < 0)
 				goto dfu_write_error;
-			state->dfu_dev.dfu_state_flag = DS5_DFU_DONE;
 		}
+		/* Do not finalize here: a write() length is not a reliable
```
ymodlin, line 6116: Same here

ymodlin (changes requested):

kernel/realsense/d4xx.c
```diff
@@ -6330,6 +6344,22 @@ static int ds5_dfu_device_release(struct inode *inode, struct file *file)
...
+	 * marks end-of-image, so issue the single zero-length DFU_DNLOAD
+	 * (0x4a04=0) to drive DOWNLOAD_IDLE -> MANIFEST -> flash-commit -> reboot.
+	 */
+	if (state->dfu_dev.dfu_state_flag == DS5_DFU_IN_PROGRESS) {
+		ret = ds5_write(state, 0x4a04, 0x00); /* zero-length DFU_DNLOAD */
+		if (!ret)
+			ret = ds5_dfu_wait_for_get_dfu_status(state, dfuMANIFEST);
+		if (ret < 0)
+			dev_err(&state->client->dev,
+				"%s(): DFU finalize failed (%d)\n", __func__, ret);
```
ymodlin, line 6361: **Potential BUG:** moving the finalize here drops the terminal state bookkeeping that the write path used to do.

- `DS5_DFU_DONE` is no longer assigned anywhere in the driver (it was set in `ds5_dfu_device_write()` before this change), so the enum value is now dead.
- A failed finalize only logs; it does not set `DS5_DFU_ERROR`. Execution falls into the next line, which sets `DS5_DFU_IDLE` — so a failed FW update is indistinguishable from a successful one to any later state check.

Worth noting #564 gates its post-DFU auto re-probe on `dfu_state_flag == DS5_DFU_DONE`; with this PR merged that condition can never be true.

Suggest setting `DS5_DFU_DONE` on success and `DS5_DFU_ERROR` on failure, and guarding the `IDLE` assignment accordingly.

kernel/realsense/d4xx.c
```diff
@@ -6097,14 +6110,15 @@ static ssize_t ds5_dfu_device_write(struct file *flip,
...
 				ret = ds5_dfu_wait_for_get_dfu_status(state, dfuDNLOAD_IDLE);
-			if (!ret)
-				ret = ds5_write(state, 0x4a04, 0x00); /*Download complete */
-			if (!ret)
-				ret = ds5_dfu_wait_for_get_dfu_status(state, dfuMANIFEST);
 			if (ret < 0)
 				goto dfu_write_error;
-			state->dfu_dev.dfu_state_flag = DS5_DFU_DONE;
 		}
+		/* Do not finalize here: a write() length is not a reliable
```
ymodlin, line 6116: **CLAUDE.md guideline violation** (Comment style):

- *"No comments that describe an absence or a removal."* This 6-line block explains what the function deliberately no longer does — exactly the pattern the rule asks to delete.
- *"In-code comments: 2-3 lines maximum"* — long rationale belongs in the commit message / PR description / CLAUDE.md.

The same rationale is already written in all three places (this comment, the one in `ds5_dfu_device_release()`, and the new CLAUDE.md paragraph). Suggest dropping this one and keeping the short `why` at the finalize site.

PR #472, opened 2026-06-15
[RSDEV-12089] d4xx: bound ds5_hwmc_wait by wall-clock deadline (fix mid-stream preset at low fps)

ymodlin:
## Problem
On D457 over MIPI/GMSL, changing a depth **visual preset mid-stream** fails at **low fps** (5 fps reliably, some 15 fps) with librealsense `error code=-110`. Works at 30/60 fps, and works if the stream starts with the preset preselected. No issue over USB.

## Root cause
A preset apply is a burst of ~35 HWMC round-trips + a few V4L2 control writes (laser/AE/gain). One of those control writes makes the driver **re-arm the stream** (`ds5_mux_s_stream`), whose confirm-poll waits on the FW `DS5_STATUS_STREAMING` bit until a **frame boundary** — 0.6–1.4 s at 5 fps — while **hogging the single shared I2C bus**. Meanwhile `ds5_hwmc_wait()` had only a ~100 ms budget (`retries=100` × `msleep_range(1)`), so a concurrent HWMC in the burst couldn't get enough bus turns before the re-arm finished → false `-ETIMEDOUT` → `-110`. It's fps-gated because the confirm bit flips on a frame boundary (instant at 30/60 fps, 200 ms at 5 fps). librealsense and the firmware were both confirmed correct; the short driver-side wait budget is the cause.

## Fix
`ds5_hwmc_wait()`: replace the fixed retry count with a **2 s jiffies deadline** (`DS5_HWMC_MAX_TIME`), matching the existing `s_stream` (`DS5_START_MAX_TIME`) and hw-reset deadline idiom in this file. The happy path still returns as soon as status leaves WIP; it only waits longer when the FW genuinely needs more time (e.g. behind a low-fps stream re-arm). No locking change — `ds5_hwmc_wait` runs under the depth subdev `state->lock` while the re-arm uses `ds5_dev->lock`, so the longer wait cannot deadlock the re-arm (deliberately **no** shared mutex added).

## Validation (fw-orin-nano-1, D457/MIPI, FW 5.17.3.22 held constant)
A/B with only the d4xx module changed:
- **Baseline** (stock d4xx 1.0.4.4): 5 fps preset-mid-stream **FAILS -110**.
- **This fix**: **ALL fps PASS** — 848×480 @ 5/15/30/60/90, 1280×720 @ 5/15/30, 640×360 @ 5/90, 424×240@5, 480×270@5 — every preset (Medium/High Accuracy/Default/High Density). **dmesg shows zero `ds5_hwmc_wait` timeouts.**

Ref: RSDEV-12089. Related: RSDEV-12096 (add a low-fps preset-switch-during-stream regression test — 30 fps masks this), RSDEV-12109 (rs-fw-update GMSL no-progress, separate).

### DIFF
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -217,6 +217,13 @@ enum ds5_mux_pad {
 #define DS5_START_POLL_TIME	10
 #define DS5_START_MAX_TIME	2000
 #define DS5_START_MAX_COUNT	(DS5_START_MAX_TIME / DS5_START_POLL_TIME)
+/*
+ * RSDEV-12089: max wall-clock to wait for an HWMC command to complete. Must cover
+ * a worst-case low-fps stream re-arm (ds5_mux_s_stream, bounded by
+ * DS5_START_MAX_TIME) that monopolises the shared I2C bus while an HWMC is in
+ * flight; a shorter budget makes the HWMC falsely time out (-110 to the host).
+ */
+#define DS5_HWMC_MAX_TIME	2000
 #define MAX_DS5_CONFIG_RETRIES	5
 
 /* I2C retry configuration */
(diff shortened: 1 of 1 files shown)


ymodlin:
## Hardware A/B re-validation — PASS ✅

Re-validated on **fw-orin-nano-1** (Jetson Orin Nano, JP6.2 / L4T R36.4.3, kernel `5.15.148-tegra`) with a **D457 over GMSL** (s/n 220422301535). Clean A/B isolating only this change: the two `d4xx.ko` builds are this commit (`a995ffa`) vs its **parent** — identical source except `ds5_hwmc_wait`.

| Build | srcversion | Result |
|---|---|---|
| **Baseline** (parent, old `retries=100` ~100 ms budget) | `8843653B1EDBC0919468652` | ❌ **FAIL** |
| **Fixed** (this PR, 2 s jiffies deadline) | `B8912E00398977B559B905C` | ✅ **PASS** |

**Repro:** start depth stream, then change `VISUAL_PRESET` mid-stream (advanced-mode HWMC burst).

**Baseline — reproduces the bug** at `848x480@5`:
```
preset HIGH_ACCURACY: FAIL -> RuntimeError: Operation failed with error code=-110
dmesg:
  d4xx 2-002a: stream 0 toggle ok to 1 in 696ms, retries 10
  d4xx 2-002a: ds5_hwmc_wait(): HWMC command timed out   (x3)
```

**Fixed — passes at every fps**, all four presets each, **zero `ds5_hwmc_wait` timeouts** in dmesg:

| Profile | HIGH_ACCURACY / HIGH_DENSITY / MEDIUM_DENSITY / DEFAULT |
|---|---|
| 848x480@5  | PASS (~1.5 s each) |
| 848x480@15 | PASS |
| 848x480@30 | PASS |
| 848x480@60 | PASS |
| 848x480@90 | PASS |
| 1280x720@5 | PASS |
| 640x360@5  | PASS |

The mid-stream preset apply now completes in ~0.9–1.5 s (the HWMC rides out the low-fps stream re-arm window that hogs the shared I2C bus) instead of falsely returning `-110` at the old ~100 ms budget. Happy-path latency for normal HWMC commands is unchanged (loop exits as soon as status leaves WIP).

**Review:** code + security reviewed — one-line, kernel-idiomatic (`time_before`/jiffies), no locking change, no memory-safety impact, brick-risk none.

_FW held constant during the A/B (only the `d4xx.ko` module swapped). The fix is FW-independent — its 2 s budget tolerates any FW-side stall._

ejgoldik (approved):

ymodlin:
@ejgoldik / @ymodlin did online code review and approved the change.

PR #478, opened 2026-06-18
install_to_kernel.sh: accept any rootfs.tar.* compression for JP6/7

Nir-Az:
**Overview:** `install_to_kernel.sh` hardcodes `rootfs.tar.gz` in the JP6/7 install branch, so a release packaged as `rootfs.tar.bz2` (or `.xz`) fails to extract and the script aborts without updating the kernel modules.

### Problem
On a Jetson Orin running JetPack 6.2, with the rootfs packaged as `rootfs.tar.bz2`:

```
Extracting rootfs.tar.gz...
tar: rootfs.tar.gz: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
Error: Failed to extract rootfs.tar.gz; not modifying kernel modules.
```

The JP6/7 branch only ever tries `tar xf rootfs.tar.gz`, ignoring any other compression.

### Fix
Resolve the actual `rootfs.tar.*` file and let `tar` autodetect the compression (gz/bz2/xz/...):

```bash
ROOTFS_TARBALL=$(ls rootfs.tar.* 2>/dev/null | head -n1)
echo "Extracting ${ROOTFS_TARBALL:-rootfs.tar.*}..."
if [ -z "${ROOTFS_TARBALL}" ] || ! tar xf "${ROOTFS_TARBALL}"; then
    echo "Error: Failed to extract rootfs tarball (rootfs.tar.*); not modifying kernel modules."
    exit 1
fi
```

This resolves the single matching tarball first (so a stray second match isn't misread as a tar member) and still errors clearly when none exists. Scope is limited to the JP6/7 branch where the hard failure occurs.

🤖 Generated by AI


ejgoldik (approved):
Great!

PR #479, opened 2026-06-21
D585 max96712 bringup (pixel mode)

ejgoldik:


kernel/nvidia/max96717.c
```diff
@@ -0,0 +1,416 @@
...
+#define MAX96717_PWDN_ESYNC_EXT 0x24
+
+#define MAX96717_2C0_ESYNC 0x57
+#define MAX96717_2C1_ESYNC 0x05
+#define MAX96717_2C2_ESYNC 0x1F
+#define MAX96717_2C3_ESYNC 0x57
+
+struct max96717_client_ctx {
+	struct gmsl_link_ctx *g_ctx;
+	bool st_done;
```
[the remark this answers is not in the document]
ejgoldik: @AikidoSec ignore: priv = dev_get_drvdata(dev) is invariantly non-NULL here - it is set in max96717_probe() before the device is usable, and these EXPORT_SYMBOL ops are only ever called by d4xx.c after the serializer has probed, so drvdata cannot be NULL. This matches the NVIDIA max9295/max9296 SerDes driver idiom, which intentionally omits the guard; adding it would be unreachable defensive code.

kernel/nvidia/max96717.c
```diff
@@ -0,0 +1,416 @@
...
+	struct mutex lock;
+};
+
+static int max96717_write_reg(struct device *dev, u16 addr, u8 val)
+{
+	struct max96717 *priv = dev_get_drvdata(dev);
+	int err;
+
+	err = regmap_write(priv->regmap, addr, val);
+	if (err)
```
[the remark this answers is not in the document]
ejgoldik: @AikidoSec ignore: priv = dev_get_drvdata(dev) is invariantly non-NULL here - it is set in max96717_probe() before the device is usable, and these EXPORT_SYMBOL ops are only ever called by d4xx.c after the serializer has probed, so drvdata cannot be NULL. This matches the NVIDIA max9295/max9296 SerDes driver idiom, which intentionally omits the guard; adding it would be unreachable defensive code.

kernel/nvidia/max96717.c
```diff
@@ -0,0 +1,416 @@
...
+	return err;
+}
+EXPORT_SYMBOL(max96717_setup_control);
+
+static int max96717_write_acc(struct device *dev, struct regmap *regmap,
+			     u16 addr, u8 val, int *err)
+{
+	unsigned int readback = 0;
+
+	if (*err)
```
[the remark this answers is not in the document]
ejgoldik: @AikidoSec ignore: priv = dev_get_drvdata(dev) is invariantly non-NULL here - it is set in max96717_probe() before the device is usable, and these EXPORT_SYMBOL ops are only ever called by d4xx.c after the serializer has probed, so drvdata cannot be NULL. This matches the NVIDIA max9295/max9296 SerDes driver idiom, which intentionally omits the guard; adding it would be unreachable defensive code.

kernel/nvidia/max96717.c
```diff
@@ -0,0 +1,416 @@
...
+	struct max96717 *priv = dev_get_drvdata(dev);
+	int err = 0;
+
+	mutex_lock(&priv->lock);
+
+	/* Configure ESYNC registers for MAX96712A deserializer */
+	max96717_write_acc(dev, priv->regmap, MAX96717_SRC_PWDN_ADDR, MAX96717_PWDN_ESYNC_EXT, &err);
+	max96717_write_acc(dev, priv->regmap, MAX96717_SRC_CTRL_ADDR, MAX96717_RESET_ESYNC, &err);
+	max96717_write_acc(dev, priv->regmap, MAX96717_2C0_ADDR, MAX96717_2C0_ESYNC, &err);
+	max96717_write_acc(dev, priv->regmap, MAX96717_2C1_ADDR, MAX96717_2C1_ESYNC, &err);
```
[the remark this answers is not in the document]
ejgoldik: @AikidoSec ignore: priv = dev_get_drvdata(dev) is invariantly non-NULL here - it is set in max96717_probe() before the device is usable, and these EXPORT_SYMBOL ops are only ever called by d4xx.c after the serializer has probed, so drvdata cannot be NULL. This matches the NVIDIA max9295/max9296 SerDes driver idiom, which intentionally omits the guard; adding it would be unreachable defensive code.

kernel/nvidia/max96717.c
```diff
@@ -0,0 +1,416 @@
...
+	struct max96717 *priv = dev_get_drvdata(dev);
+	int err = 0;
+
+	mutex_lock(&priv->lock);
+
+	/* Restore default SRC pin function — plain GPIO, no ESYNC tunneling */
+	max96717_write_acc(dev, priv->regmap, MAX96717_SRC_PWDN_ADDR, MAX96717_PWDN_GPIO, &err);
+	max96717_write_acc(dev, priv->regmap, MAX96717_SRC_CTRL_ADDR, MAX96717_RESET_SRC, &err);
+	max96717_write_acc(dev, priv->regmap, MAX96717_2C0_ADDR, 0x00, &err);
+	max96717_write_acc(dev, priv->regmap, MAX96717_2C1_ADDR, 0x00, &err);
```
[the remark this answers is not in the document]
ejgoldik: @AikidoSec ignore: priv = dev_get_drvdata(dev) is invariantly non-NULL here - it is set in max96717_probe() before the device is usable, and these EXPORT_SYMBOL ops are only ever called by d4xx.c after the serializer has probed, so drvdata cannot be NULL. This matches the NVIDIA max9295/max9296 SerDes driver idiom, which intentionally omits the guard; adding it would be unreachable defensive code.

kernel/nvidia/max96717.c
```diff
@@ -0,0 +1,416 @@
...
+{
+	struct max96717 *priv = dev_get_drvdata(dev);
+	int err = 0;
+
+	mutex_lock(&priv->lock);
+	if (!priv->g_client.g_ctx) {
+		dev_err(dev, "%s: no sdev client found\n", __func__);
+		err = -EINVAL;
+		goto error;
+	}
```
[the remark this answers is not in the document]
ejgoldik: @AikidoSec ignore: priv = dev_get_drvdata(dev) is invariantly non-NULL here - it is set in max96717_probe() before the device is usable, and these EXPORT_SYMBOL ops are only ever called by d4xx.c after the serializer has probed, so drvdata cannot be NULL. This matches the NVIDIA max9295/max9296 SerDes driver idiom, which intentionally omits the guard; adding it would be unreachable defensive code.

kernel/nvidia/max96717.c
```diff
@@ -0,0 +1,416 @@
...
+
+	priv = dev_get_drvdata(dev);
+	mutex_lock(&priv->lock);
+	if (priv->g_client.g_ctx) {
+		dev_err(dev, "%s: device already paired\n", __func__);
+		err = -EINVAL;
+		goto error;
+	}
+
+	priv->g_client.st_done = false;
```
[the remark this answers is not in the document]
ejgoldik: @AikidoSec ignore: priv = dev_get_drvdata(dev) is invariantly non-NULL here - it is set in max96717_probe() before the device is usable, and these EXPORT_SYMBOL ops are only ever called by d4xx.c after the serializer has probed, so drvdata cannot be NULL. This matches the NVIDIA max9295/max9296 SerDes driver idiom, which intentionally omits the guard; adding it would be unreachable defensive code.

kernel/nvidia/max96717.c
```diff
@@ -0,0 +1,416 @@
...
+	u8 val;
+};
+
+static int max96717_set_registers(struct device *dev, struct reg_pair *map,
+				 u32 count)
+{
+	int err = 0;
+	u32 j = 0;
+
+	for (j = 0; j < count; j++) {
```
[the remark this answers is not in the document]
ejgoldik: @AikidoSec ignore: priv = dev_get_drvdata(dev) is invariantly non-NULL here - it is set in max96717_probe() before the device is usable, and these EXPORT_SYMBOL ops are only ever called by d4xx.c after the serializer has probed, so drvdata cannot be NULL. This matches the NVIDIA max9295/max9296 SerDes driver idiom, which intentionally omits the guard; adding it would be unreachable defensive code.

nvidia-oot/max96717.c
```diff
@@ -0,0 +1,403 @@
...
+		goto error;
+	}
+
+	priv->g_client.st_done = false;
+
+	priv->g_client.g_ctx = g_ctx;
+
+error:
+	mutex_unlock(&priv->lock);
+	return 0;
```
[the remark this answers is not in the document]
ejgoldik: Fixed

kernel/nvidia/max96717.c
```diff
@@ -0,0 +1,416 @@
...
+		goto error;
+	}
+
+	priv->g_client.st_done = false;
+
+	priv->g_client.g_ctx = g_ctx;
+
+error:
+	mutex_unlock(&priv->lock);
+	return 0;
```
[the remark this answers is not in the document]
ejgoldik: Fixed

utilities/depth_capture/ir_capture.py
```diff
@@ -0,0 +1,261 @@
...
+                frame = bytes(maps[b.index][:ln])
+            fcntl.ioctl(fd, VIDIOC_QBUF, b)
+            if got_data:
+                break
+        fcntl.ioctl(fd, VIDIOC_STREAMOFF, C.c_int(BUF_TYPE_VIDEO_CAPTURE))
+        if not got_data:
+            print("WARNING: no frame contained data in %d frames -- is the IR stream running "
+                  "during the capture? Keeping last (empty) frame." % nframes)
+        if raw_out and frame is not None:
+            with open(raw_out, "wb") as o:
```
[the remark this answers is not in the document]
ejgoldik: @AikidoSec ignore: local developer diagnostic CLI run manually by the operator; raw_out is the user's own --raw argument, not attacker-controlled input, and this is not a network service. The suggested ".." check is ineffective (ignores absolute paths/symlinks) and would break legitimate relative paths.

nvidia-oot/max96717.c
```diff
@@ -0,0 +1,412 @@
...
+EXPORT_SYMBOL(max96717_init_settings);
+
+int max96717_set_pipe(struct device *dev, int pipe_id,
+		     u8 data_type1, u8 data_type2, u32 vc_id)
+{
+	/* No runtime config needed in pixel mode - Only reset rx and oneshot */
+	struct max96717 *priv = dev_get_drvdata(dev);
+	int err = 0;
+
+	mutex_lock(&priv->lock);
```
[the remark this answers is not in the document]
ejgoldik: @AikidoSec ignore: priv = dev_get_drvdata(dev) is invariantly non-NULL here - it is set in max96717_probe() before the device is usable, and these EXPORT_SYMBOL ops are only ever called by d4xx.c after the serializer has probed, so drvdata cannot be NULL. This matches the NVIDIA max9295/max9296 SerDes driver idiom, which intentionally omits the guard; adding it would be unreachable defensive code.

kernel/realsense/d4xx.c
```diff
@@ -41,12 +41,12 @@
 #include <media/max9295.h>
 #include <media/max9296.h>
 #include <media/max96712.h>
+#include <media/max96717.h>
 
 /* Deserializer interface structure for abstraction */
 struct dser_interface {
 	/* Pipeline management */
 	int (*get_available_pipe_id)(struct device *dev, int vc_id);
-	int (*get_ser_pipe_id)(struct device *dev, int dser_pipe_id, int vc_id);
```
ymodlin, line 49: Why you remove this?

kernel/realsense/d4xx.c
```diff
@@ -665,6 +687,32 @@ static const struct dser_interface max96712_interface = {
 	.name = "max96712",
 };
 
+/* MAX9295 serializer interface implementation */
```
ymodlin, line 690: How did it work before this interface?

kernel/realsense/d4xx.c
```diff
@@ -1423,6 +1472,73 @@ static const struct ds5_format ds5_onsemi_rgb_format = {
 };
 #define DS5_ONSEMI_RGB_N_FORMATS 1
 
+static const struct ds5_resolution d58x_depth_sizes[] = {
```
ymodlin, line 1477: I hope this is the last camera we are adding like this

kernel/realsense/d4xx.c
```diff
@@ -1423,6 +1472,73 @@ static const struct ds5_format ds5_onsemi_rgb_format = {
...
+	DS5_RES(640, 360, ds5_framerate_to_60)
+	DS5_RES(1280, 720, ds5_framerate_to_60)
```
ymodlin, line 1479: maybe we should consider to offload to header file per cameras

kernel/realsense/d4xx.c
```diff
@@ -1795,24 +1911,47 @@ static int ds5_sensor_set_fmt(struct v4l2_subdev *sd,
...
+ * The ser_pipe_id mapping depends both on the serializer nad deserializer identity:
+ *   - MAX96717 serializer:       ser_pipe_id = 2 (fixed)
+ *   - MAX9295 + MAX96712 dser:   ser_pipe_id = ser_vc_id
+ *   - MAX9295 + MAX9296 dser:    ser_pipe_id = dser_pipe_id
+ */
+static int serdes_get_ser_pipe_id(struct ds5 *state, int dser_pipe_id,
+				  int ser_vc_id)
+{
+	if (state->ser_ops == &max96717_interface)
+		return 2;
```
ymodlin: Why magic numbers?

kernel/realsense/d4xx.c
```diff
@@ -3935,6 +4073,22 @@ static int ds5_board_setup(struct ds5 *state)
 	}
 
 	state->ser_dev = &ser_i2c->dev;
+	/* Initialize serializer interface. Match by name prefix so device-tree
```
ymodlin, line 4079: **[blocker]** OF node refcount leak. The original code called `of_node_put(ser_node)` immediately after `of_find_i2c_device_by_node()` (line ~4064); this PR removed that and moved the put down to the end of this serializer-selection block. But three `goto error` paths now sit between the `of_parse_phandle` and this put and never release the node:

1. `of_property_read_u32(ser_node, "reg", ...)` failure ("serializer reg not found")
2. `if (ser_i2c == NULL)` → `-EPROBE_DEFER` — **this fires on every probe retry** until the MAX96717 driver loads, so the leak accumulates
3. `if (ser_i2c->dev.driver == NULL)`

Simplest fix: restore the unconditional `of_node_put(ser_node)` right after `of_find_i2c_device_by_node()` as before — `ser_node->name` is only read (not held) by the strncmp() calls below, so it's safe to put first and then read the name.

kernel/realsense/d4xx.c
```diff
@@ -1795,24 +1911,47 @@ static int ds5_sensor_set_fmt(struct v4l2_subdev *sd,
 }
 
 #ifdef CONFIG_VIDEO_D4XX_SERDES
+/*
+ * Resolve the serializer pipe id for the current SerDes topology.
+ *
+ * The ser_pipe_id mapping depends both on the serializer nad deserializer identity:
```
ymodlin: **[code-style]** Typo: "the serializer nad deserializer" → "and".

ymodlin (commented):
Code review of the D585 / MAX96717 bringup. Findings are posted as inline comments, each prefixed with a severity tag:

- **[blocker]** — must fix before merge: a correctness bug, leak, or behavior regression that affects function or stability.
- **[nit]** — minor defect, optional but worth fixing: a latent bug with low trigger probability, a wrong-but-harmless errno, dead code, etc.
- **[code-style]** — formatting / convention / consistency only; no behavioral impact.

The serializer-vtable refactor, the D58X format tables, and the new `tegra234-camera-d4xx-overlay-fg12-16ch-d5xx.dts` overlay all check out — the DT file faithfully matches the proven `fg12-16ch` overlay convention (per-stream `port-index`/`tegra_sinterface`, depth node without `override_reg`). Two blockers below are worth fixing before merge.

Note (not inline, lives in nested .patch files): the JP5 Kconfig entries read `tristate "MAX92717 Serializer"` — typo, should be `MAX96717` (appears in the 5.0.2 / 5.1.2 / 5.1.6 porting patches). [code-style]

kernel/realsense/d4xx.c
```diff
@@ -3935,6 +4073,22 @@ static int ds5_board_setup(struct ds5 *state)
...
 	state->ser_dev = &ser_i2c->dev;
+	/* Initialize serializer interface. Match by name prefix so device-tree
+	 * nodes with link suffixes (e.g. max9295_a, max9295_b) are recognized. */
+	if (!strncmp(ser_node->name, "max9295", strlen("max9295"))) {
+		state->ser_ops = &max9295_interface;
+	} else if (!strncmp(ser_node->name, "max96717", strlen("max96717"))) {
+		state->ser_ops = &max96717_interface;
+	} else {
+		dev_err(dev, "%s: Unsupported serializer = %s\n", __func__, ser_node->name);
+		/* Should not be used, this is just to make sure we don't have NULL pointers */
```
ymodlin: **[nit]** This `state->ser_ops = &max9295_interface;` assignment is dead — the branch immediately does `err = -ENODEV; goto error;`, and no caller inspects `ser_ops` on an error return. The "so we don't have NULL pointers" comment is misleading: it reads like a real fallback to MAX9295 when in fact probe fails. Suggest dropping the assignment (leave `ser_ops` unset on the error path).

nvidia-oot/max96717.c
```diff
@@ -0,0 +1,403 @@
...
+
+	if (!priv->g_client.g_ctx) {
+		dev_err(dev, "%s: no sensor dev client found\n", __func__);
+		err = -EINVAL;
+		goto error;
+	}
+
+	g_ctx = priv->g_client.g_ctx;
+
+	max96717_write_reg(dev, MAX96717_I2C4_ADDR, (g_ctx->sdev_reg << 1));
```
ymodlin: **[blocker]** `max96717_setup_control()` drops the return values of both proxy-address writes and sets `g_ctx->serdev_found = true` unconditionally, then returns 0. Compare `max9295_setup_control()`, which writes `CTRL0` and checks the result ("ser device not found") as a link-presence gate. Because this always returns 0, the retry loop in `ds5_gmsl_serdes_setup()` — whose whole purpose is to retry until the GMSL link is up — exits on the first attempt even when the I2C writes failed, and the device proceeds with un-programmed proxy addresses. Capture `err` from both writes and `goto error` on failure before setting `serdev_found`.

Same issue in `kernel/nvidia/max96717.c:100`.

nvidia-oot/max96717.c
```diff
@@ -0,0 +1,403 @@
...
+static int max96717_probe(struct i2c_client *client,
+				const struct i2c_device_id *id)
+#endif
+{
+	struct max96717 *priv;
+	int err = 0;
+
+	dev_info(&client->dev, "[MAX96717]: probing GMSL Serializer\n");
+
+	priv = devm_kzalloc(&client->dev, sizeof(*priv), GFP_KERNEL);
```
ymodlin, line 341: **[nit]** `devm_kzalloc()` can return NULL; the next line dereferences `priv->i2c_client = client` with no check → NULL deref under memory pressure. Add `if (!priv) return -ENOMEM;`. (This mirrors the pre-existing max9295 pattern, but since this is a fresh file it's worth getting right. Same in `kernel/nvidia/max96717.c:344`.)

nvidia-oot/max96717.c
```diff
@@ -0,0 +1,403 @@
...
+		return -EINVAL;
+	}
+
+	priv = dev_get_drvdata(dev);
+
+	mutex_lock(&priv->lock);
+
+	if (!priv->g_client.g_ctx) {
+		dev_err(dev, "%s: device is not paired\n", __func__);
+		err = -ENOMEM;
```
ymodlin: **[nit]** "device is not paired" returns `-ENOMEM`; should be `-EINVAL` (copy-paste from max9295). No runtime impact today since the d4xx unpair callers only log the value, but the errno is semantically wrong and will mislead debugging. Same in `kernel/nvidia/max96717.c:232`.

nvidia-oot/max96717.c
```diff
@@ -0,0 +1,403 @@
...
+};
+
+static const struct of_device_id max96717_of_match[] = {
+	{ .compatible = "maxim,max96717", },
+	{ },
+};
+MODULE_DEVICE_TABLE(of, max96717_of_match);
+MODULE_DEVICE_TABLE(i2c, max96717_id);
+
+static struct i2c_driver max96717_i2c_driver = {
```
ymodlin, line 396: **[code-style]** `max96717_of_match[]` is defined and exported via `MODULE_DEVICE_TABLE(of, ...)` but never wired into the driver: `.driver` has no `.of_match_table = of_match_ptr(max96717_of_match)`. The sibling `max96712` driver sets it (`.of_match_table = of_match_ptr(max96712_of_match)`). DT probe still works here via the `i2c_device_id` fallback (`{ "max96717", 0 }`), so this isn't a blocker, but the of_match table is currently dead code and diverges from the repo convention — please wire it in.

kernel/nvidia/max96717.c
```diff
@@ -0,0 +1,416 @@
...
+{
+	return i2c_add_driver(&max96717_i2c_driver);
+}
+
+static void __exit max96717_exit(void)
+{
+	i2c_del_driver(&max96717_i2c_driver);
+}
+
+module_init(max96717_init);
```
ymodlin: **[code-style]** This file open-codes `module_init`/`module_exit` + `i2c_add_driver`/`i2c_del_driver`, whereas `nvidia-oot/max96717.c` uses the idiomatic `module_i2c_driver(max96717_i2c_driver)`. Prefer `module_i2c_driver()` here too for consistency. (Same `.of_match_table` omission as the OOT file — see note on `nvidia-oot/max96717.c:389`.)

kernel/nvidia/max96717.c
```diff
@@ -0,0 +1,416 @@
...
+
+	if (!priv->g_client.g_ctx) {
+		dev_err(dev, "%s: no sensor dev client found\n", __func__);
+		err = -EINVAL;
+		goto error;
+	}
+
+	g_ctx = priv->g_client.g_ctx;
+
+	max96717_write_reg(dev, MAX96717_I2C4_ADDR, (g_ctx->sdev_reg << 1));
```
ymodlin: **[blocker]** Same as `nvidia-oot/max96717.c:90` — `setup_control()` ignores the proxy-write return codes and sets `serdev_found = true` / returns 0 unconditionally, defeating the GMSL link-up retry loop in d4xx. Propagate `err` and `goto error` on write failure.

utilities/v4l2_capture/v4l2_capture.py
```diff
@@ -0,0 +1,253 @@
...
+
+        rb = v4l2_requestbuffers(); rb.count = 4
+        rb.type = BUF_TYPE_VIDEO_CAPTURE; rb.memory = MEMORY_MMAP
+        fcntl.ioctl(fd, VIDIOC_REQBUFS, rb)
+
+        maps = []
+        for i in range(rb.count):
+            b = v4l2_buffer(); b.type = BUF_TYPE_VIDEO_CAPTURE; b.memory = MEMORY_MMAP; b.index = i
+            fcntl.ioctl(fd, VIDIOC_QUERYBUF, b)
+            mm = mmap.mmap(fd, b.length, mmap.MAP_SHARED, mmap.PROT_READ, offset=b.m.offset)
```
ymodlin, line 109: **[nit]** The `mmap` buffers in `maps` are never `.close()`d — the `finally` only does `os.close(fd)`. Harmless for a one-shot CLI run (GC reclaims at exit), but a genuine leak if `capture()` is ever looped or imported. Add `for mm in maps: mm.close()` before closing the fd. Same in `ir_capture.py` and `nv12_capture.py`.

utilities/depth_capture/depth_capture.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+def _has_data(buf, ln):
+    """Cheap all-zero test: sample ~4K bytes spread across the frame.
+    Distinguishes a real frame from a zero-filled (no-stream) error buffer."""
+    step = max(1, ln // 4096)
+    return any(buf[i] for i in range(0, ln, step))
+
+
+def capture(dev, W, H, nframes, save_idx, raw_out):
```
ymodlin: **[code-style]** `depth_capture.py`, `ir_capture.py`, and `nv12_capture.py` share ~540 lines that are byte-identical (the ioctl encoding, all V4L2 ctypes structs, VIDIOC constants, `_has_data()`, the capture loop, `_show()`). Only the fourcc and `render()` differ. Consider factoring the shared parts into a `v4l2_capture.py` module so an ABI/ioctl fix only has to be made once.

kernel/realsense/d4xx.c
ejgoldik: Because the deserializer is no longer the entity that determines which pipe the serializer will use. Now you have to know both the deserializer and the serializer identity, so this logic was moved to the d4xx driver itself

kernel/realsense/d4xx.c
ejgoldik: Before this, d4xx called max9295 functions directly, not using an interface abstraction. Now that we have another serializer, I added this.

kernel/realsense/d4xx.c
ejgoldik: That is indeed the goal

kernel/realsense/d4xx.c
ejgoldik: We can do it in a follow-up, but depends if we are going for the refactor soon, in which case probably shouldn't bother

kernel/realsense/d4xx.c
ejgoldik: Max96717 only has pipe 2. I can add a define

kernel/realsense/d4xx.c
ejgoldik: Fixed, although it makes me nervous

kernel/realsense/d4xx.c
ejgoldik: Fixed although claude suggested not to do put and then access the name

kernel/realsense/d4xx.c
ejgoldik: Added define

kernel/realsense/d4xx.c
ejgoldik: Fixed

nvidia-oot/max96717.c
ejgoldik: Fixed, although max9295 also does not check the retval of these two writes.

nvidia-oot/max96717.c
ejgoldik: Fixed

nvidia-oot/max96717.c
ejgoldik: Fixed

kernel/nvidia/max96717.c
ejgoldik: Fixed

kernel/nvidia/max96717.c
ejgoldik: fixed

utilities/v4l2_capture/v4l2_capture.py
ejgoldik: Fixed

utilities/depth_capture/depth_capture.py
ejgoldik: Fixed

nvidia-oot/max96717.c
ejgoldik: Fixed

utilities/v4l2_capture/v4l2_capture.py
```diff
@@ -0,0 +1,206 @@
...
+                frame = bytes(maps[b.index][:ln])
+            fcntl.ioctl(fd, VIDIOC_QBUF, b)
+            if got_data:
+                break
+        fcntl.ioctl(fd, VIDIOC_STREAMOFF, C.c_int(BUF_TYPE_VIDEO_CAPTURE))
+        if not got_data:
+            print("WARNING: no frame contained data in %d frames -- is the stream running "
+                  "during the capture? Keeping last (empty) frame." % nframes)
+        if raw_out and frame is not None:
+            with open(raw_out, "wb") as o:
```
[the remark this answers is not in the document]
ejgoldik: @SikidoSec ignore: already ignored before file was renamed. This is irrelevant as it is a local testing utility.

nvidia-oot/max96717.c
ejgoldik: Fixed

ymodlin (approved):

PR #488, opened 2026-06-30
apply_patches_ext.sh: add JP7/noble (7.x) support for ext-source builds

ymodlin:
## What
Add a `7.x` branch to `apply_patches_ext.sh` so the **external-source** patch flow works on JetPack 7 (noble / L4T R39.x).

## Why
`apply_patches_ext.sh` is the variant used by the Artifactory-based Jenkins pipeline (it patches an externally-supplied source tree via `patch -p1`, rather than the git-clone tree that `apply_patches.sh` uses with `git apply`). It only handled `5.x`/`6.x`: for `7.x` it copied `d4xx.c` but **skipped all the noble-specific steps**, so a JP7.2 ext build had no `max96712.h` and no D4XX device-tree overlays.

This is the prerequisite for building **JP7.2 in Jenkins via the Artifactory source flow** (companion PR in `librealsense-deploy`: `LRS_Jetson_mipi_driver` → JP7.2).

## Change
Mirrors the existing `7.x` block of `apply_patches.sh`, but `cp`s into the external source tree (`$TARGET`) instead of `ln`-ing into the git-clone tree:
- `max96712.h` → `nvidia-oot/include/media/`
- `tegra264-gpio.h` dt-binding (for Thor overlay compilation), if missing
- tegra264 (Thor) + tegra234 (Orin Fangzhu) `.dtso` overlays → noble kernel `arch/arm64/boot/dts/nvidia/`

## Notes / review asks
- Assumes the R39.2 `public_sources.tbz2` extracts to the same internal layout as the git-clone flow (`kernel/kernel-noble-src`, `nvidia-oot`, `3rdparty/canonical/linux-noble`). Please confirm against the actual Artifactory bundle.
- No effect on 5.x/6.x paths.

### DIFF
--- a/apply_patches_ext.sh
+++ b/apply_patches_ext.sh
@@ -54,6 +54,18 @@ if [[ "$JETPACK_VERSION" == "6.x" ]]; then
     cp hardware/realsense/tegra234-camera-d4xx-overlay*.dts "$TARGET/hardware/nvidia/t23x/nv-public/overlay/"
     # max96712 header
     cp nvidia-oot/max96712.h "$TARGET/nvidia-oot/include/media/"
+elif [[ "$JETPACK_VERSION" == "7.x" ]]; then
+    # max96712 header
+    cp nvidia-oot/max96712.h "$TARGET/nvidia-oot/include/media/"
+    # tegra264-gpio.h for Thor overlay compilation if not already present
+    if [[ ! -f "$TARGET/$KERNEL_DIR/include/dt-bindings/gpio/tegra264-gpio.h" ]]; then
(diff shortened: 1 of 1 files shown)


ymodlin:
Superseded: the JP7.2 source artifact the team uploaded to Artifactory is the pre-synced `sources_7.2` git tree (`tegra-l4t-r39.2.tar.gz`), not the `public_sources.tbz2` layout. The Jenkins job now feeds that tarball into `setup_workspace.sh`'s cache mechanism and uses the normal `apply_patches.sh` (git-apply) flow, so the ext-source 7.x branch added here isn't exercised. Reopen if we ever drive JP7 through the `apply_patches_ext.sh` / public_sources route (e.g. a usb_jp7 pipeline).

PR #489, opened 2026-07-01
Add FG12-4CH D457 support

pinquanw:
## Summary

Add support for running a RealSense D457 on Jetson Orin Nano with the FangZhu FG12-4CH carrier board using the validated full 2-lane path.

- Add `tegra234-camera-d4xx-overlay-fg12-4ch.dts`.
- Add the FG12-4CH overlay dtbo to the JP6 Orin overlay build patch.
- Add independently gated FangZhu FG12 reverse-lane-polarity and POC sequences in the shared MAX96712 nvidia-oot patch.
- Keep the D4XX driver unchanged; lane count stays on the existing 2-lane path through the DTS endpoints and modes.

## Version Coverage

The MAX96712 driver patch is shared by the supported nvidia-oot patch sets:

- `nvidia-oot/6.1/0003...` -> `../6.0/0003...`
- `nvidia-oot/6.2/0003...` -> `../6.0/0003...`
- `nvidia-oot/7.0/0006...` -> `../6.0/0003...`
- `nvidia-oot/7.1/0006...` -> `../7.0/0006...`
- `nvidia-oot/7.2/0006...` -> `../7.0/0006...`

So the single `nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch` update covers JP6 and JP7 variants that consume those symlinks.

## Why This Is Needed

The FG12-4CH board routes the D457 through MAX9295 to MAX96712 at the FG12-4CH deserializer address (`0x29`) and uses CSI port C (`port-index = <2>` / `serial_c`). The overlay configures that board topology and keeps all CSI endpoints and D457 modes at 2 lanes.

The MAX96712 behavior added for this board is split behind two independent DT flags:

- `fangzhu,fg12-reverse-lane-polarity` applies the FG12-4CH lane polarity configuration.
- `fangzhu,fg12-poc-enable` enables the FangZhu board POC power sequence before the serializer/camera path is accessed.

The polarity writes and POC writes are kept separate because lane polarity and power-over-coax are board characteristics that are not inherently tied together. Boards that do not set these DT flags keep the existing generic MAX96712 path.

## Validation

- Hardware validated on Orin Nano + MAX96712 FG12-4CH + MAX9295 / RealSense D457 using `./build_all.sh 6.2`.
- Verified the final PR diff does not modify `kernel/realsense/d4xx.c`.
- Verified the final PR diff only includes the FG12-4CH overlay, Orin overlay dtbo build entry, and shared MAX96712 reverse-lane-polarity / POC hooks.
- Ran `git diff --check origin/dev...HEAD`.

### DIFF
--- a/nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
+++ b/nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
@@ -7,14 +7,14 @@ Add MAX96712 deserializer driver support for D4XX/D5XX cameras.
 
 Signed-off-by: ejgoldik <ehud.joseph.goldik@realsenseai.com>
 ---
- drivers/media/i2c/max96712.c | 607 ++++++++++++++++++++++++++++++++---
- 1 file changed, 570 insertions(+), 37 deletions(-)
+ drivers/media/i2c/max96712.c | 672 ++++++++++++++++++++++++++++++++++++++++---
+ 1 file changed, 635 insertions(+), 37 deletions(-)
 
 diff --git a/drivers/media/i2c/max96712.c b/drivers/media/i2c/max96712.c
-index 8e237eef..3b4376fe 100644
+index 8e237eef..42df25b3 100644
 --- a/drivers/media/i2c/max96712.c
 +++ b/drivers/media/i2c/max96712.c
-@@ -16,32 +16,104 @@
+@@ -16,32 +16,111 @@
  #include <linux/gpio.h>
  #include <linux/of.h>
  #include <linux/of_gpio.h>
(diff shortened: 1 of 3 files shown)


nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
```diff
@@ -358,6 +367,31 @@ index 8e237eef..3b4376fe 100644
 +	return err;
 +}
 +
++static int max96712_enable_fangzhu_fg12_poc(struct max96712 *priv)
++{
++	struct reg_pair poc_enable[] = {
++		{MAX96712_MIPI_PHY5_ADDR, 0x3f},
```
ejgoldik, line 379: Does the FG12-4CH in general use reverse lane polarity? If so better to comment this on these reg writes, it's more fundamental than just providing power over coax.
Correct me if I'm wrong, but it seems currently the only boards we have that require POC IO writes are the ones that are reversed polarity, but it seems that the two aren't necessarily related. Perhaps they should be driven by two distinct flags. One for POC and one for reversed polarity that would explicitly state "reverse polarity".

ejgoldik (changes requested):

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
pinquanw: Yes, FG12-4CH use reverse lane polarity. I will split polarity and POC into two function and add more comments.

ejgoldik (approved):

PR #490, opened 2026-07-02
Fix JetPack 5.1.6 streaming: G_PARM EINVAL + D457 color-path panic/recovery

ymodlin:
## Fix JetPack 5.1.6 (L4T R35.6.x) streaming regression on D457 GMSL

Upgrading a D457 GMSL rig from JetPack 5.0.2 / 5.1.2 to **5.1.6** broke streaming: `VIDIOC_G_PARM` returned `EINVAL`, and starting the **color** stream hard-hung the box (panic → watchdog reboot). Depth/IR alone worked; color and multi-stream did not. These three patches restore full streaming — each stream alone and all combinations in parallel.

### Root cause
JP5.1.6 pulled in two NVIDIA L4T changes that the RealSense D457 color path was never adapted to:
- **VI channel lifecycle rewrite** — L4T 35.6.x *"camera: vi5: fix stream on/off memory leakage"* (bug 4206070): channels are now tracked as `struct file` in `chan->fp[]` and every close path goes through `filp_close()`.
- **CSI error-config change** — L4T 35.5 *"csi5: add error config"* (bug 4499208): the v4l2 path now sends `NVCSI_CONFIG_FLAG_ERROR` with a zeroed `nvcsi_error_config`, replacing the RCE default error handling used on 5.1.2.

Three resulting failures:
1. **G_PARM `EINVAL`** — the 5.1.6 patch set (created from 5.1.2) dropped the direct-op G/S_PARM fallback: `0001` kept the wrapped `v4l2_subdev_call()` but not the fallback. `__tegra_channel_get_parm` passes an on-stack `v4l2_subdev_frame_interval` with an uninitialized `pad`, so kernel 5.10's `check_pad()` rejects the call before the sensor op runs — librealsense can't start any stream (`rs2_pipeline_start()` → `G_PARM` failed). (`S_PARM` set `pad` explicitly, so it worked — the diagnostic tell.)
2. **Color-path panic** — D457 color frame 0 always arrives corrupted (`err_data 0x20000`; expected, also on 5.1.2). The RS `vi5` error recovery still used the old `vi_channel_close_ex`/`open_ex` sequence, which bypasses `chan->fp[]`; after recovery `chan->fp[]` was left pointing at freed `private_data`, so closing the stream NULL-deref'd in `vi_capture_shutdown` → panic. *(This mechanism is the working diagnosis from the crash signature plus on-HW behavior; the 35.6.x `vi_capture_shutdown` internals were not independently source-reviewed — see Notes.)*
3. **No recovery** — with the new `NVCSI_CONFIG_FLAG_ERROR` override the same corruption is reported as `err_data 0x100`, which the D457 recovery mask in `vi5_fops.c` does not cover → corr_err flood, no recovery, no frames.

### Fix (3 patches under `kernel/nvidia/5.1.6/`)
- **`0003-vi-channel-verify-s-g-callbacks.patch`** — restore the direct-op `g_frame_interval` fallback lost from 5.1.2, so `G_PARM` succeeds.
- **`0012-csi5-do-not-override-rce-error-config-on-v4l2-path.patch`** — drop `NVCSI_CONFIG_FLAG_ERROR` so the RCE defaults apply and the corruption is reported as `err_data 0x20000` (covered by the existing recovery mask) instead of `0x100`.
- **`0013-vi5-port-d457-error-recovery-to-filp-lifecycle.patch`** — port the D457 recovery to the filp lifecycle:
  - `vi_capture_release(CAPTURE_CHANNEL_RESET_FLAG_IMMEDIATE)` **before** `filp_close()` (this frees the RCE channel-match configuration; without it the reopen fails with *"match configuration is already in use"*), then reopen via `vi5_channel_open()` so `chan->fp[]` stays consistent.
  - On a **failed** `vi_capture_release`, **bail (`goto done`)** rather than force-closing — force-closing a channel whose release just failed drives `vi_capture_shutdown` against inconsistent state (the NULL-deref this patch avoids), and it matches NVIDIA's `vi5_channel_error_recover` on the JP6.2 base.
  - **NULL `chan->fp[vi_port]` right after `filp_close()`** (alongside the `tegra_vi_channel` reset), so a partial recovery can't leave a freed pointer for the real stream-stop to `filp_close()` a second time.

### Validation — hardware (D457 GMSL, Jetson AGX Xavier, L4T R35.6.4, FW 5.17.3.21)
Kernel with all three patches, via librealsense (callback-based), no panic, `pstore` clean:

| Permutation | Result |
|---|---|
| depth / color / ir (each alone) | ✅ 30 fps |
| depth+color, depth+ir, color+ir | ✅ 30 + 30 fps |
| depth+color+ir (all parallel) | ✅ 30 + 30 + 30 fps |

Color frame-0 recovery fires as designed (channel reset + re-arm) and streaming continues through it; also consistent with the LibCI fps test (depth+color, 5–90 fps).

### Notes
- Raw `v4l2-ctl` color-standalone still needs the SDK/driver-coordinated frame-0 recovery-retry (depth/IR work under raw v4l2); use librealsense to exercise color. Out of scope here.
- **Verification method:** the patches were reviewed against the JP6.2 filp-model reference (`vi5_channel_open` / `vi5_channel_error_recover`) plus on-HW behavior; a 35.6.x kernel source tree was not available for line-level confirmation of the `vi_capture_shutdown` path, so the frame-0-panic mechanism above is the working diagnosis rather than a source-verified claim.

### DIFF
--- a/kernel/nvidia/5.1.6/0013-vi5-port-d457-error-recovery-to-filp-lifecycle.patch
+++ b/kernel/nvidia/5.1.6/0013-vi5-port-d457-error-recovery-to-filp-lifecycle.patch
@@ -1,0 +1,80 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Yogev Modlin <yogev.modlin@realsenseai.com>
+Date: Wed, 2 Jul 2026 16:11:00 +0300
+Subject: [PATCH] vi5: port D457 channel error recovery to filp-based lifecycle
+
+L4T 35.6.x ("camera: vi5: fix stream on/off memory leakage", bug
+4206070) rewrote the VI channel lifecycle: channels are tracked as
+struct file in chan->fp[] and all close paths use filp_close().
+vi5_channel_error_recover_internal() still used the old sequence
+(vi_channel_close_ex + vi_channel_open_ex), which bypasses chan->fp[]
+bookkeeping. On the first D457 color-frame corruption (err_data
+0x20000) the recovery left chan->fp[] pointing at freed private_data;
+closing the stream then oopsed in vi_capture_shutdown ->
+tegra_camrtc_reboot (NULL deref) and panicked the whole system, with a
+secondary Oops in tegra_capture_ivc_worker during the RTCPU teardown
+race.
+
+Port the recovery to the new lifecycle: reset the RCE-side channel
+first (vi_capture_release with CAPTURE_CHANNEL_RESET_FLAG_IMMEDIATE,
+while the handle is still valid - this is what actually frees the CSI
+channel-match configuration in the RCE; without it the reopen fails
+with "match configuration is already in use"), then filp_close(), then
+reopen through vi5_channel_open() so chan->fp[] stays consistent.
+
+Validated on Jetson AGX Xavier + D457 GMSL (L4T 35.6.4): color streams
+at 5-90 fps with the first-frame corruption recovered transparently; no
+panic on stream close.
+
+Signed-off-by: Yogev Modlin <yogev.modlin@realsenseai.com>
+---
+ drivers/media/platform/tegra/camera/vi/vi5_fops.c | 25 ++++++++++-----------
+ 1 file changed, 12 insertions(+), 13 deletions(-)
+
+diff --git a/drivers/media/platform/tegra/camera/vi/vi5_fops.c b/drivers/media/platform/tegra/camera/vi/vi5_fops.c
+index 845d0cedf..a1c6e5bc5 100644
+--- a/drivers/media/platform/tegra/camera/vi/vi5_fops.c
++++ b/drivers/media/platform/tegra/camera/vi/vi5_fops.c
+@@ -539,15 +539,16 @@ static int vi5_channel_error_recover_internal(struct tegra_channel *chan)
+ 	struct v4l2_subdev *csi_subdev;
+ 	dev_dbg(chan->vi->dev, "%s() vc: %d\n", __func__, chan->virtual_channel);
+ 
+-	/* stop vi channel */
++	/* stop vi channel (filp-based lifecycle on 35.6.x base).
++	 * Reset the RCE-side channel FIRST (frees the CSI match config)
++	 * while the VI channel handle is still valid, then close the filp.
++	 */
+ 	for(vi_port = 0; vi_port < chan->valid_ports; vi_port++) {
+ 		err = vi_capture_release(chan->tegra_vi_channel[vi_port],
+ 			CAPTURE_CHANNEL_RESET_FLAG_IMMEDIATE);
+-		if (err) {
+-			dev_err(&chan->video->dev, "vi capture release failed\n");
+-			goto done;
+-		}
+-		vi_channel_close_ex(chan->id, chan->tegra_vi_channel[vi_port]);
++		if (err)
++			dev_err(chan->vi->dev, "vi capture release failed\n");
++		filp_close(chan->fp[vi_port], NULL);
+ 		chan->tegra_vi_channel[vi_port] = NULL;
+ 	}
+ 
+@@ -583,14 +584,11 @@ static int vi5_channel_error_recover_internal(struct tegra_channel *chan)
+ 	v4l2_subdev_call(csi_subdev, core, sync,
+ 		V4L2_SYNC_EVENT_SUBDEV_ERROR_RECOVER);
+ 
+-	/* restart vi channel */
++	/* restart vi channel (filp-based lifecycle on 35.6.x base) */
+ 	for(vi_port = 0; vi_port < chan->valid_ports; vi_port++) {
+-		chan->tegra_vi_channel[vi_port] = vi_channel_open_ex(chan->id + vi_port, false);
+-		if (IS_ERR(chan->tegra_vi_channel[vi_port])) {
+-			err = PTR_ERR(chan->tegra_vi_channel[vi_port]);
+-			chan->tegra_vi_channel[vi_port] = NULL;
++		err = vi5_channel_open(chan, vi_port);
++		if (err < 0)
+ 			goto done;
+-		}
+ 		err = tegra_channel_capture_setup(chan, vi_port);
+ 		if (err < 0)
+ 			goto done;
+-- 
+2.34.1
--- a/kernel/nvidia/5.1.6/0013-vi5-port-d457-error-recovery-to-filp-lifecycle.patch
+++ b/kernel/nvidia/5.1.6/0013-vi5-port-d457-error-recovery-to-filp-lifecycle.patch
@@ -22,23 +22,33 @@ channel-match configuration in the RCE; without it the reopen fails
 with "match configuration is already in use"), then filp_close(), then
 reopen through vi5_channel_open() so chan->fp[] stays consistent.
 
+Keep the bail-out (goto done) on a failed vi_capture_release instead of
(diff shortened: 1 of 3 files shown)


kernel/nvidia/5.1.6/0013-vi5-port-d457-error-recovery-to-filp-lifecycle.patch
```diff
@@ -0,0 +1,80 @@
...
+ 		err = vi_capture_release(chan->tegra_vi_channel[vi_port],
+ 			CAPTURE_CHANNEL_RESET_FLAG_IMMEDIATE);
+-		if (err) {
+-			dev_err(&chan->video->dev, "vi capture release failed\n");
+-			goto done;
+-		}
+-		vi_channel_close_ex(chan->id, chan->tegra_vi_channel[vi_port]);
++		if (err)
++			dev_err(chan->vi->dev, "vi capture release failed\n");
++		filp_close(chan->fp[vi_port], NULL);
```
ipilchin1, line 66: **Latent use-after-free on the reopen-failure path — `chan->fp[vi_port]` isn't nulled after `filp_close`.**

`filp_close()` frees the file, but `chan->fp[vi_port]` still points at it (only `tegra_vi_channel` is nulled on the next line). On the happy path the reopen loop overwrites `chan->fp[]` via `vi5_channel_open()`, so it's a transient. But if the reopen loop `goto done`s early (`vi5_channel_open` / `tegra_channel_capture_setup` fails) on a `valid_ports > 1` config (D457 dual-port), any port closed here but not yet reopened keeps a freed `chan->fp[]` — and the next real stream-stop will `filp_close()` a freed pointer.

Suggest nulling it alongside `tegra_vi_channel`:

```c
filp_close(chan->fp[vi_port], NULL);
chan->fp[vi_port] = NULL;
chan->tegra_vi_channel[vi_port] = NULL;
```

My verdict: **should-fix** — cheap, and it closes a real error-path UAF.

kernel/nvidia/5.1.6/0013-vi5-port-d457-error-recovery-to-filp-lifecycle.patch
```diff
@@ -0,0 +1,80 @@
...
+ 	for(vi_port = 0; vi_port < chan->valid_ports; vi_port++) {
+ 		err = vi_capture_release(chan->tegra_vi_channel[vi_port],
+ 			CAPTURE_CHANNEL_RESET_FLAG_IMMEDIATE);
+-		if (err) {
+-			dev_err(&chan->video->dev, "vi capture release failed\n");
+-			goto done;
+-		}
+-		vi_channel_close_ex(chan->id, chan->tegra_vi_channel[vi_port]);
++		if (err)
++			dev_err(chan->vi->dev, "vi capture release failed\n");
```
ipilchin1, line 62: **Question: is `filp_close` safe here after a _failed_ `vi_capture_release`?**

This changed from bail-out to log-and-continue. For comparison, NVIDIA's mature recovery on the JP6.2 base (`vi5_channel_error_recover` in `vi5_fops.c`) does `goto done` on `vi_capture_release` failure — it does **not** force a close afterward. Here we fall through into `filp_close()`, which runs the char-dev `.release` fop → `vi_capture_shutdown`: the exact path this patch's commit message says NULL-derefs / panics when the channel state is inconsistent. Force-closing a channel whose release just failed is precisely that inconsistent-state case.

@ymodlin — can you confirm the 35.6.x `vi_capture_shutdown` path is hardened against a close after a failed release (i.e. it can't re-trigger the panic 0013 is fixing)? If not, keeping the `goto done` on the `err` branch (matching NVIDIA) may be safer. I couldn't verify the shutdown internals — no 35.6.x source tree was reachable to me; this is checked against the 6.2 filp-model reference.

My verdict: **not a blocker**, but worth confirming before merge.

ipilchin1 (commented):
Reviewed all three patches. Note on method: no L4T 35.6.x source tree was reachable to me, so I verified the recovery against the JP6.2 filp-model implementation of `vi5_channel_open`/`vi5_channel_error_recover` plus the "normal streaming works" fact — a couple of points below are inference, called out inline.

- **0003 (G_PARM fallback)** and **0012 (csi5 error-config revert)** look sound.
- **0013 reopen path** checks out: `vi5_channel_open()` is the shared opener used on normal stream-start, and since normal streaming + filp_close-based stop works on 5.1.6, it necessarily repopulates both `chan->fp[]` and `chan->tegra_vi_channel[]`. The `if (err < 0)` return check is correct for that signature.

Two notes on 0013's stop loop below — one should-fix, one question.

kernel/nvidia/5.1.6/0013-vi5-port-d457-error-recovery-to-filp-lifecycle.patch
ymodlin: Fixed in `de6e4b3` — `chan->fp[vi_port] = NULL;` now runs right after `filp_close()`, alongside `chan->tegra_vi_channel[vi_port] = NULL`. So even on an early `goto done` out of the reopen loop (including the `valid_ports > 1` D457 dual-port case), no port is left holding a freed `chan->fp[]` for the next stream-stop to `filp_close()` a second time. Thanks for catching the error-path UAF.

kernel/nvidia/5.1.6/0013-vi5-port-d457-error-recovery-to-filp-lifecycle.patch
ymodlin: Good point — took the conservative route in `de6e4b3`: on a failed `vi_capture_release` the recovery now bails with `goto done` instead of falling through to `filp_close()`, matching NVIDIA's `vi5_channel_error_recover` on the JP6.2 base.

To answer directly: I did **not** assume the 35.6.x `vi_capture_shutdown` is hardened against a close-after-failed-release — I couldn't source-review it either (no 35.6.x tree available to me). So rather than depend on that, the fix simply avoids the force-close on the failed-release path. In the normal D457 frame-0 recovery `vi_capture_release` succeeds, so this is purely the defensive branch now.

ymodlin:
Thanks @ipilchin1 — both `0013` notes are addressed in `de6e4b3` (null `chan->fp[vi_port]` after `filp_close()`, and bail via `goto done` on a failed `vi_capture_release` instead of force-closing); details in the threads.

I've also rewritten the PR description for accuracy per your review:
- dropped the unverifiable commit hashes — now referencing the L4T changes by title + **bug 4206070** (vi5 stream on/off leak / filp lifecycle) and **bug 4499208** (csi5 error config);
- corrected the reopen-failure string to the actual *"match configuration is already in use"*;
- updated the `0013` summary to the current bail + null-`fp` behavior;
- removed the specific recovery-mask hex (kept it as "the D457 recovery mask in `vi5_fops.c`");
- added a **Verification method** note that the frame-0-panic mechanism is the working diagnosis from the crash signature + on-HW behavior, since a 35.6.x source tree wasn't available for line-level confirmation of the `vi_capture_shutdown` path.

ipilchin1 (approved):
Both `0013` review items are addressed in `de6e4b3` and verified against the diff:
- `chan->fp[vi_port] = NULL;` now runs right after `filp_close()` — closes the error-path UAF.
- On a failed `vi_capture_release` the recovery bails via `goto done` instead of force-closing, matching NVIDIA's JP6.2 `vi5_channel_error_recover`.

Both hunks re-count correctly so the patch applies. `0003`/`0012` unchanged. LGTM.

PR #492, opened 2026-07-03
[Host][GMSL]: Support D5xx CSI metadata

polejoe:
## Summary

Report D5xx CSI metadata sidecar payload sizes from the runtime `HKRM` header instead of stream-specific fixed sizes.

## Changes

- Use a generic CSI metadata capture capacity instead of Depth/IR and RGB fixed word-count values.
- Parse and validate the runtime `HKRM` CSI metadata header before deriving sidecar `bytesused`.
- Validate header magic, version, stream type, payload size bounds, and captured byte bounds.
- Derive valid metadata `bytesused` as `sizeof(csi_metadata_header) + csiHdr->metadata_size`.
- Keep invalid, truncated, or legacy metadata on the existing 68-byte fallback path so old firmware remains usable.
- Keep unrelated device-tree profile movement out of this PR.

## Runtime Size Semantics

- The metadata node buffer size is capacity only.
- The host driver does not assume Depth, IR, or RGB metadata payload sizes.
- The app-visible V4L2 metadata payload length comes from the validated per-frame header.
- Legacy firmware without a valid `HKRM` header falls back to 68 bytes, matching the previous sidecar behavior.

## Related PRs

- Device manager metadata buffer sizing: https://github.com/RealSenseAI-Internal/soc-rs-soc-device-mgr/pull/385
- IO payload support: https://github.com/RealSenseAI-Internal/soc-rs-soc-io-drivers/pull/425
- Preview metadata display: https://github.com/RealSenseAI-Internal/soc-rs-hkr-test/pull/268

## Verification

- `git diff --check -- kernel/realsense/d5xx.c nvidia-oot/6.0/0001-Porting-nvidia-driver-patches-to-jetpack-6.0.patch nvidia-oot/6.2/0001-Porting-nvidia-driver-patches-to-jetpack-6.0.patch nvidia-oot/7.0/0001-Porting-nvidia-driver-patches-to-jetpack-6.0.patch nvidia-oot/7.1/0001-Porting-nvidia-driver-patches-to-jetpack-6.0.patch`
- Header validation logic check for valid Depth/RGB metadata, bad magic, bad version, bad stream type, zero payload, oversized payload, truncated capture, and tiny legacy capture.
- `./build_all.sh --d5xx --fg 6.2.2`

### DIFF
--- a/kernel/realsense/d5xx.c
+++ b/kernel/realsense/d5xx.c
@@ -192,8 +192,7 @@ struct dser_interface {
 #define DS5_STATUS_UNAVAILABLE		0xffff
 
 #define MIPI_LANE_RATE				1300
-#define DS5_CSI_METADATA_DEPTH_IR_WC	256	/* csi header 20 + depth/IR metadata 236 */
-#define DS5_CSI_METADATA_RGB_WC		263	/* csi header 20 + RGB metadata 243 */
+#define DS5_CSI_METADATA_MAX_WC		4096
 
 #define MAX_DEPTH_EXP				200000
 #define MAX_RGB_EXP					10000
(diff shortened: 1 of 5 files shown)


lemon6630 (approved):

polejoe:
Replaced by #508, which uses the upstream `realsenseai:zhous/gmsl_e2e_meta` branch and is rebased onto `d500_gmsl_dev`.

PR #505, opened 2026-07-16
[RSDEV-12608] d4xx: gate GMSL one-shot link reset to cold bring-up

Nikolai-L:
### Root cause (driver-side, confirmed on HW)
`max96712.c __max96712_set_pipe()` issued a GMSL **one-shot link reset** (`CTRL1` + `msleep(100)`) on **every** pipe (re)alloc. All of a D457's streams share one physical GMSL link (`vc_id_msb = vc_id >> 2`). d4xx releases the SERDES pipe on every stream stop and re-runs `set_pipe` on restart, so re-arming **color** on a toggle re-trained the whole link and **truncated the depth/IR frames in flight** → Tegra-VI `PIXEL_INCOMPLETE` → 2500 ms capture timeout → channel reset → RGB dies. `max9296`'s `set_pipe` writes the same class of registers with **no** reset, which is why max9296-only rigs never reproduced.

The reset's *granularity* was already correct (one link); the bug was the *trigger* — per-alloc instead of per-cold-bring-up.

### Fix — `arm_link_reset` (additive deserializer op)
Move only the **decision** to d4xx; keep the reset at its exact existing site inside `__max96712_set_pipe`.

- d4xx computes "is this the first active stream on this camera's link?" from the **reset-cleared** `ds5_dev->{depth,rgb,ir,imu}_streaming` flags, captured under `ds5_dev->lock` **before** writing its own flag (`ds5_mux_s_stream`).
- It passes the decision to the deserializer via a **new additive** `dser_interface` op `arm_link_reset(dev, bool)` (not a `set_pipe` signature change → ABI-additive, preserves the validated reset position), called under `serdes_lock__` immediately before `set_pipe` in `ds5_setup_pipeline`.
- `max96712` stores the bool (`pending_link_reset`, default `true`) and consumes it at the `CTRL1` site; it re-arms `true` afterwards as a fail-safe. `max9296` is left as a no-op (the d4xx call is guarded `if (ops->arm_link_reset)`).
- A DT property `maxim,oneshot-on-alloc` forces the reset on for any camera that needs per-alloc re-init (unused by default).

**Why flag-driven, not a deser-side pipe refcount:** `ds5_hw_reset_with_recovery` clears the streaming flags but deliberately does **not** release SERDES pipes. A counter in `struct max96712` would stay ≥2 through recovery → skip the re-lock the recovery path depends on → both streams come up on an un-retrained link (0 frames). Keying off the reset-cleared flags keeps recovery correct — proven below.

### Files
- `sources_6.2/nvidia-oot/drivers/media/i2c/max96712.c` — via regenerated patch `nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch` (shared by all JP6.x **and** JP7.x via symlinks). Adds `pending_link_reset`, `max96712_arm_link_reset()` + `EXPORT_SYMBOL`, and the gated `CTRL1` call.
- `nvidia-oot/max96712.h` — prototype + `MAX96712_HAS_ARM_LINK_RESET` feature macro.
- `kernel/realsense/d4xx.c` — the op, the two `struct ds5` flags, DT parse, the `ds5_mux_s_stream` capture, the `ds5_setup_pipeline` call, and the macro-guarded interface wiring.

### Cross-version / CI
- **JP6.x + JP7.x**: `d4xx.c`, `max96712.h`, and patch `0003`/`0006` are shared (symlinks), so all get the fix from the same source.
- **JP5 (Xavier)**: uses independent `kernel/nvidia/` patches and a different header. The `MAX96712_HAS_ARM_LINK_RESET` macro (defined only in the shared 6.x/7.x header) gates the d4xx wiring, so JP5 **builds unchanged** (op stays NULL → old behavior). JP5 was not in scope and has no rig here; a follow-up can port the same change to the JP5 patches if desired.
- New symbol is module-to-module (`max96712.ko` → `d4xx.ko`, both rebuilt together); no kernel public-header change, so the BSP display-module ABI is unaffected.

### Diagnostics
Two low-noise `dev_info` lines were added at the reset sites (per-link fire in `__max96712_set_pipe`, coarse `0x0F` in `max96712_reset_oneshot`). They fire only on rare events (cold bring-up / recovery / calib), never per-frame, and provide field visibility into exactly this class of GMSL link event. They were also the only way to directly count `CTRL1` on the rig kernel (no dynamic_debug / kprobes / regmap tracepoint).

### Validation (fw-orin-nano-1, D457 s/n 220422301535, FW 5.17.3.25, JP6.2)
| metric | stock 1.0.5.3 | **this fix** |
|---|---|---|
| PIXEL_INCOMPLETE (`err_data 131072`), 15 toggles | 22 | **0** |
| VI errors total | 27 | **4** (all at final teardown, none on a toggle) |
| color NOFRAME restarts | hard-break | **0/15** |
| CTRL1 per link on cold start | — | **exactly 1** |
| CTRL1 during the 15 toggles | ~15 | **0** |

**Recovery sweep (the case the refcount design fails):** with depth+color streaming, forced `ds5_hw_reset_with_recovery` via the HW Reset control → recovery orchestration ran clean (`Device ready after 712 ms`, `GMSL link recovered (device type 0x0006)`) → the **first restart fired the CTRL1** (`GMSL link 0 one-shot reset fired`) → **both streams recovered** (depth 203→241, color 57→72 frames), `PIXEL_INCOMPLETE = 0`.

Zero `PIXEL_INCOMPLETE` across the entire session (cold + 15 toggles + forced recovery + smoke). Post-test D457 smoke healthy (depth 162 / color 148 in 5 s).

### Follow-up (separate ticket)
The Y12I/`is_calib` coarse `reset_oneshot(0x0F)` (all-links) is **unchanged** by this PR and still cross-glitches other cameras on the same deserializer — to be filed separately.

### Draft status
Validated on JP6.2 only. JP7 shares the exact source but was not run on JP7 hardware. Marking draft pending CI (all JP versions) and any JP7 bring-up check.


Nikolai-L:
Superseded and split into two stacked PRs off current `dev`:
- #506 — `.claude` skills sync (tooling, unrelated to the fix)
- #507 — the RSDEV-12608 driver fix, stacked on #506

Both are rebased onto current dev; the fix's patch 0003 was regenerated against dev's OOT and confirmed via `apply_patches reset`/`apply_patches` + a clean JP6.2 build. Closing this 1.0.5.3-based combined PR in favor of those.

PR #506, opened 2026-07-16
chore(.claude): sync driver-workflow skills (soak-watch, mipi-bandwidth-check)

Nikolai-L:
## `.claude` driver-workflow skills sync (not related to any driver fix)

Repo tooling housekeeping for on-host driver work — independent of the RSDEV-12608 driver fix (that lands in the stacked PR that targets this branch).

- Remove 6 legacy `*.skill` zip duplicates (`build`/`deploy`/`streaming-test`/`v4l2-test`/`verify-deploy` + a misplaced `workspace-setup.skill`).
- Add two skills:
  - `soak-watch` — escalate-on-event watcher for long hardware soaks (no per-turn polling).
  - `mipi-bandwidth-check` — GMSL/CSI SOF/EOF bandwidth + coexistence-headroom measurement.
- `.gitignore` — also ignore the underscore `v4l2_test_results/` output dir (the existing rule only matched the dashed variant, so test output was being tracked).

No driver / kernel / device-tree changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #507, opened 2026-07-16
[RSDEV-12608] move the GMSL one-shot flush from set_pipe to HW-reset recovery

Nikolai-L:
## RSDEV-12608 (P1) — D457 GMSL: toggling one stream off/on truncates the siblings

**Root cause (HW-confirmed).** `max96712` fired a GMSL one-shot (`CTRL1` + `msleep(100)`) on **every** pipe (re)alloc in `__max96712_set_pipe` / `__max96712_set_multi_vc_pipe`. That one-shot only **flushes the deserializer's pixel line buffer** — it does *not* re-train the link. All of a camera's streams share one physical GMSL link, so flushing on a plain stream toggle **truncated the sibling streams' in-flight frames** → Tegra-VI `PIXEL_INCOMPLETE` → 2500 ms timeout → stream dies. (`max9296`'s `set_pipe` leaves no stale buffer, so max9296-only rigs never reproduced.)

**Fix.**
- Remove the flush from `set_pipe` (both sites) — a pipe (re)alloc must never touch the shared link.
- The flush *is* needed after a **camera HW reset**: the camera resets mid-frame but the serializer keeps its config + link lock, leaving a stale partial frame nothing else clears. Issue it **once, per-link, from `ds5_hw_reset_with_recovery()`** via a new additive deserializer op `reset_oneshot_link(dev, vc_id)` — max96712 flushes `1 << (vc_id / MAX9295_MAX_STREAMS)`; `max9296` wires no op (NULL, call is guarded).

**Scope.** All JetPacks: `d4xx.c` (symlinked) + `max96712.{c,h}` via patch `0003` (JP6/7) and `0007`/`0017` (JP5). Module-to-module symbol → ABI-safe, no kernel-header change.

**Validation.**
- *Falsification:* a no-flush build **kernel-panicked** during recovery (stale frame → depth never completes → VI 2500 ms cascade → `err_rec` NULL-deref) — proving the recovery flush is load-bearing and can't live only at probe time.
- *HW recovery:* nano-1 max96712 (D457 + D401) **4/4** and D401 **4/4**; fw-orin-3 max9296 D401 **3/3** (flush correctly NOT fired); **0 panic**.
- *On-device v4l2 pytest suite* green on both rigs — streaming / controls / error-handling / HW-reset-recovery all pass; the few remaining failures are test-harness or camera-config (metadata-CRC modeling, a 90 fps consumer-throughput limit, D401 laser), not this change.
- *Build CI* green across all JP versions.

**Notes.**
- Stacked on #506 (`.claude` skills sync); retarget to `dev` once #506 merges.
- Out of scope: the coarse all-links calib `reset_oneshot(0x0F)` (→ RSDEV-13028) and the transient VI 2500 ms residual that self-recovers (→ RSDEV-13047).


kernel/realsense/d4xx.c
```diff
@@ -4230,6 +4254,12 @@ static int ds5_board_setup(struct ds5 *state)
 	}
 	state->g_ctx.dst_vc = value;
 
+	/* RSDEV-12608: opt-in for cameras that need the deserializer one-shot
+	 * link reset on every pipe alloc (not just cold bring-up). Absent by
+	 * default, so normal cameras rely purely on link_cold_bringup. */
+	state->needs_oneshot_on_alloc =
+		of_property_read_bool(gmsl, "maxim,oneshot-on-alloc");
```
ejgoldik: Device tree is not aware of which camera is connected (d401/457/430 etc) so this can't be a device tree flag.

Nikolai-L: Yep, will fix that

kernel/realsense/d4xx.c
```diff
@@ -5427,6 +5448,18 @@ static int ds5_mux_s_stream(struct v4l2_subdev *sd, int on)
 
 	restore_val = sensor->streaming;
 	mutex_lock(&state->ds5_dev->lock);
+	/* RSDEV-12608: before marking our own stream active, capture whether any
```
ymodlin: Shorter comments please

sareluzi:
@Nikolai-L  - did you mean PR to dev or chore/claude-skills-sync?

nvidia-oot/6.0/0002-max9295-max9296-SerDes-support-for-D4xx.patch
sareluzi, line 1: Did you mean to change this file?  I thought change affects only 96712.
if so - it's pretty hard to review as too many things moved... can you rearrange it for easier review?

Nikolai-L:
> @Nikolai-L - did you mean PR to dev or chore/claude-skills-sync?

chore/claude-skills-sync

nvidia-oot/6.0/0002-max9295-max9296-SerDes-support-for-D4xx.patch
Nikolai-L: Yep, we redesigned one shot reset approach globally.
I can try but it's almost impossible with the patches approach and we know that :-(

sareluzi:
> > @Nikolai-L - did you mean PR to dev or chore/claude-skills-sync?
> 
> chore/claude-skills-sync

@Nikolai-L  - can you explain? what is this branch? why not to dev?

ejgoldik (changes requested):
Commenting here what we talked about offline:
In general need to understand when we really need oneshot reset when in max96712.
As far as we currently know, d457 never needs it, and d401 needs it after set pipe.
If that is the case, we can expose oneshot_reset_link_mask functions, for d4xx to call after set_pipe if the camera is d401 and remove the call to it from set_pipe completely.

But we first need to be sure when it is really needed

d4xx already calls the full oneshot reset when toggling calibration mode, just need to add the selective link oneshot reset for d401 if the deserializer exposed this function

sareluzi:
> Commenting here what we talked about offline: In general need to understand when we really need oneshot reset when in max96712. As far as we currently know, d457 never needs it, and d401 needs it after set pipe. If that is the case, we can expose oneshot_reset_link_mask functions, for d4xx to call after set_pipe if the camera is d401 and remove the call to it from set_pipe completely.
> 
> But we first need to be sure when it is really needed
> 
> d4xx already calls the full oneshot reset when toggling calibration mode, just need to add the selective link oneshot reset for d401 if the deserializer exposed this function

note that we have other GMSL cameras - D430 GMSL, D415 GMSL. Need to check impact on those if we think there's camera specific behavior to tolerate.

Nikolai-L:
> > > @Nikolai-L - did you mean PR to dev or chore/claude-skills-sync?
> > 
> > 
> > chore/claude-skills-sync
> 
> @Nikolai-L - can you explain? what is this branch? why not to dev?

FW team agreed to separate skills/config patches from functional ones, so chore/claude-skills-sync is about skills related to the driver/CI/Jetson...

Nikolai-L:
> > Commenting here what we talked about offline: In general need to understand when we really need oneshot reset when in max96712. As far as we currently know, d457 never needs it, and d401 needs it after set pipe. If that is the case, we can expose oneshot_reset_link_mask functions, for d4xx to call after set_pipe if the camera is d401 and remove the call to it from set_pipe completely.
> > But we first need to be sure when it is really needed
> > d4xx already calls the full oneshot reset when toggling calibration mode, just need to add the selective link oneshot reset for d401 if the deserializer exposed this function
> 
> note that we have other GMSL cameras - D430 GMSL, D415 GMSL. Need to check impact on those if we think there's camera specific behavior to tolerate.

I hope CI covers it :-)

ejgoldik (approved):

PR #508, opened 2026-07-16
[Host][GMSL]: Support D5xx CSI metadata

polejoe:
## Summary

This PR replaces #492 with an upstream branch based directly on `d500_gmsl_dev`.

It adds host-side D5xx CSI metadata support by reporting metadata sidecar `bytesused` from the runtime HKRM CSI metadata header instead of relying on a fixed 68-byte legacy payload size. The runtime path validates the metadata header magic, version, stream type, and payload size before using it, and keeps invalid or legacy metadata on the existing 68-byte fallback path for compatibility with older firmware.

The NVIDIA OOT patch hunk ranges are also corrected for the expanded metadata helper so the JetPack 6.x/7.x patch stack applies cleanly.

## Validation

- `git diff --check origin/d500_gmsl_dev..HEAD`
- `git apply --check --verbose` for the JetPack 6.2 NVIDIA OOT patch against the JetPack 6.2.2 source tree
- Full Orin GMSL build and deploy to Jetson 67 was previously completed for this change set; `orin-gmsl verify --require-loaded` passed with `d5xx`, `max96717`, and `max96724` loaded and deployed artifacts matching by SHA256.

### DIFF
--- a/kernel/realsense/d5xx.c
+++ b/kernel/realsense/d5xx.c
@@ -192,6 +192,8 @@ struct dser_interface {
 #define DS5_STATUS_UNAVAILABLE		0xffff
 
 #define MIPI_LANE_RATE				1300
+#define DS5_CSI_METADATA_DEPTH_IR_WC	256	/* csi header 20 + depth/IR metadata 236 */
+#define DS5_CSI_METADATA_RGB_WC		263	/* csi header 20 + RGB metadata 243 */
 
 #define MAX_DEPTH_EXP				200000
 #define MAX_RGB_EXP					10000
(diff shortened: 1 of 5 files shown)


polejoe:
Replaced by #510. The new PR keeps the same validated D5xx CSI metadata sidecar function, but moves runtime metadata sizing behind per-sensor metadata ops and keeps generic VI code free of D5xx/HKRM policy.

polejoe:
Closing in favor of #510. The replacement keeps the validated D5xx CSI metadata sidecar behavior, removes the generic VI HKRM/D5xx policy, and moves runtime metadata sizing behind sensor-provided metadata ops.

PR #513, opened 2026-07-20
Fix D5xx GMSL probe reload cleanup

polejoe:
## Status

Ready for review.

Review page: http://zsy-nuc13.lan:3000/pr513-d5xx-probe-reload-202607201546

NVCSI debugfs diagnostics are out of scope and remain isolated in PR #514.

## Summary

Fix D5xx GMSL probe stability across repeated `d5xx` module reloads, make the
MAX96724 source registry safe across probe teardown, and isolate D5xx internal
symbols from the D4xx `ds5` namespace.

## Root Cause

The FG24 topology exposes Depth, RGB, IR, and IMU as logical D5xx instances
that share one physical GMSL link. Probe failure and remove paths did not
consistently release the primary instance, serializer state, and MAX96724
source registration. A later reload could retain stale registration or
reference state, reject the primary instance, and leave peer instances unable
to access the camera through the I2C alias.

The MAX96724 registry also used a borrowed `gmsl_link_ctx` pointer for identity
checks. That context is embedded in devm-managed D5xx probe state and cannot be
dereferenced after teardown. Control references and detected-source accounting
were global rather than balanced per registered source.

`d4xx.c` and `d5xx.c` retained overlapping global `ds5_*` and `DS5_*`
identifiers from their common origin. They currently build as separate
modules, but the overlap makes diagnostics ambiguous and prevents safe future
linkage into the same kernel object.

## Changes

- Add one primary-instance SERDES cleanup path for setup failure, later probe
  failure, remove, and module reload.
- Track serializer and deserializer control setup independently so each
  successful stage is reset exactly once.
- Clear the shared D5xx primary pointer under lock before unpairing and
  unregistering the source.
- Store MAX96724 source identity and compatibility fields by value; retain the
  borrowed context only for live operations under the registry mutex.
- Reject active duplicate device or GMSL-link registrations instead of
  replacing their contexts.
- Track control setup and serializer discovery per source, with balanced
  `sdev_ref` and `num_src_found` updates on reset and unregister.
- Commit setup accounting only after control configuration succeeds and make
  repeated setup/reset calls idempotent.
- Compact released source slots and enforce configured and array capacities.
- Rename D5xx internal functions, types, globals, and macros to `d5x_*` and
  `D5X_*`.
- Rename D5xx diagnostic attributes to `d5x_fw_ver`, `d5x_read_reg`, and
  `d5x_write_reg`.
- Document the D4xx/D5xx naming boundary in the repository development
  guidance.

## Compatibility

The following externally consumed names remain unchanged:

- I2C driver and device-tree matching: `d5xx`, `realsense,d5xx`
- Media graph entities: `DS5 mux`, `D4XX`
- DFU class and character device: `d4xx-class`, `d4xx-dfu`
- Existing V4L2 formats, controls, media topology, and metadata behavior

Only the three D5xx diagnostic sysfs filenames change from inherited `ds5_*`
names to `d5x_*`. No consumer was found in the host application, tests,
librealsense, or installed enumeration rules.

## Validation

Validated on Jetson 67 with JetPack 6.2.2 and kernel `5.15.185-tegra`.

- JP6.2.2 modules-only build: PASS.
- Build-log scan for warnings, compiler errors, undefined references, and
  duplicate definitions: PASS.
- Package deployment, reboot, module vermagic, and installed hash checks: PASS.
- Three consecutive `rmmod d5xx; modprobe d5xx` cycles while MAX96717 and
  MAX96724 remained loaded: PASS.
- All four logical D5xx instances rebound after every reload: PASS.
- Seven V4L2 nodes were recreated and device register `0x030c` remained
  readable after every reload: PASS.
- Kernel scan found no duplicate registration, failed probe, refcount,
  WARNING, BUG, KASAN, or use-after-free diagnostics.
- Depth, IR, and RGB 1280x720 headless decode: PASS with zero sequence drops
  and zero out-of-order frames.
- Depth, IR, and RGB metadata sidecars: PASS; all buffers parsed and validated
  with zero missing, truncated, mismatched, or backwards metadata.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -18,6 +18,7 @@ Linux kernel driver and userspace utilities for Intel RealSense D4XX series 3D d
 - **Function naming**: prefix all functions with `ds5_`. Mux-related functions use `ds5_mux_`. Examples: `ds5_read()`, `ds5_write()`, `ds5_probe()`, `ds5_mux_s_stream()`.
 - **Struct naming**: prefix with `ds5_`. Examples: `struct ds5`, `struct ds5_sensor`, `struct ds5_ctrls`, `struct ds5_format`.
 - **Macro naming**: prefix with `DS5_`. Register addresses: `DS5_FW_VERSION`, `DS5_START_STOP_STREAM`, `DS5_DEPTH_STREAM_DT`.
+- **D5xx naming**: the rules above apply to `d4xx.c`; use `d5x_*` and `D5X_*` in `d5xx.c`, including D5xx diagnostic sysfs attributes. Keep shared ABI names required by existing character devices, media graph discovery, and DT matching unchanged unless an ABI migration is explicitly required.
 - **Driver name**: `DS5_DRIVER_NAME` = `"d4xx"`, with variants `-awg`, `-asr`, `-class`, `-dfu`.
 - **I2C access**: use `ds5_read()` / `ds5_write()` wrappers around `regmap_raw_read()` / `regmap_raw_write()` with built-in retry logic (`DS5_I2C_RETRY_COUNT=5`, `DS5_I2C_RETRY_DELAY_US=5000`).
 - **Polling vs. normal I2C semantics**: for transient-failure-expected polling loops (HWMC status checks, reset readiness polls, DFU timeout checks), use `ds5_read_poll()` instead of `ds5_read()`. `ds5_read_poll()` performs a single direct `regmap_raw_read()` call without retry or logging, preventing false warnings and excessive log spam on expected transients. Reserve `ds5_read()` for normal I2C operations where retry logic and verbose failure logging are desired.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -76,6 +76,7 @@ RealSense D4XX camera module
 ### Key directories
 
 - **`kernel/realsense/d4xx.c`** — The main driver. Single-file V4L2 subdevice driver handling I2C communication, MIPI CSI-2 stream config, firmware control (DFU), calibration data, metadata capture, and V4L2 controls (exposure, laser power, AE ROI, etc.). Registers four sensor subdevices per camera: Depth, RGB, IR (Y8/Y8I/Y12I), and IMU.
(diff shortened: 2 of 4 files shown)


polejoe:
Frozen for review. The PR description now reflects the final code scope and compatibility boundary. Chinese review page: http://zsy-nuc13.lan:3000/pr513-d5xx-probe-reload-202607201546. NVCSI debugfs work is isolated in draft PR #514.

lemon6630 (approved):

polejoe:
Addressed the three lifecycle findings in `c0bd109`.

- **Stale MAX96724 source context:** confirmed. Registry identity and
  compatibility fields are now copied into the registry entry. Lookup,
  duplicate detection, and unregister no longer dereference a previous
  sensor-owned `gmsl_link_ctx`. The borrowed context is used only by live
  operations under the registry mutex. An exact repeated registration is
  idempotent; a different context for an active device or GMSL link returns
  `-EBUSY`.
- **Partial serializer/deserializer control setup:** confirmed. D5xx now tracks
  serializer and deserializer setup independently. If serializer setup
  succeeds and deserializer setup fails, cleanup resets the serializer stage
  and balances `pst2_ref` before unpair/unregister.
- **Stale `num_src_found`:** confirmed. MAX96724 now tracks `control_setup` and
  `serdev_found` per source. `sdev_ref` and `num_src_found` are committed once
  after successful setup and are decremented once by reset or defensive
  unregister before slot compaction.

These findings do not involve debugfs, so all three are fixed in this PR.

Validation on Jetson 67 / JP6.2.2:

- modules-only build and build-log error scan: PASS
- three `rmmod d5xx; modprobe d5xx` cycles with MAX96717/MAX96724 retained:
  PASS
- four D5xx instances, seven V4L2 nodes, and `0x030c` I2C read after each
  cycle: PASS
- Depth/IR/RGB headless frame decode and metadata sidecars: PASS, with no
  sequence drops, backwards frames/metadata, parse errors, truncation, or
  stream mismatch
- kernel WARNING/BUG/KASAN/refcount/use-after-free scan: clean

ghu3 (approved):

PR #514, opened 2026-07-20
Add runtime NVCSI debugfs diagnostics

polejoe:
## Summary

Add read-only NVCSI debugfs diagnostics that distinguish runtime VI state, device-tree routing, and the NVCSI driver's initialized format cache.

## Problem

The existing `/sys/kernel/debug/csi/summary` reports values from the initialized `tegra_csi_port` cache. Before a negotiated format is propagated into that cache, every channel can appear as VC0, 1920x1080, and `MEDIA_BUS_FMT_SRGGB10_1X10`. The `sensor: unavailable` line only means `camera_common_data` is not attached to the NVCSI subdevice; it does not prove that the source subdevice is absent.

This makes the current output unsuitable for distinguishing four GMSL virtual channels or diagnosing the VI capture state.

## Changes

- Keep `summary` as a compatibility entry and back it with the expanded status output.
- Add `status` for controller state, NVCSI and VI streaming state, source binding, video node, capture and descriptor sequences, queued requests, queue errors, and metadata queue state.
- Add `routes` for device-tree endpoint routing and the bound VI channel's runtime stream ID, virtual channel, CSI port, and lane count.
- Add `formats` for configured and active VI formats, metadata node format and capacity, embedded-data dimensions, and the NVCSI port format cache.
- Label the NVCSI port format cache as `driver-default` when it has not been negotiated, instead of presenting it as an active stream format.
- Validate the sensor-mode pointer and index before dereferencing them.

All entries are read-only. The change does not write hardware registers or invent cumulative error counters; hardware and RTCPU errors remain available through kernel logs and tracing.

## Compatibility

- The existing `csi/summary` path remains available.
- No V4L2, media-controller, device-tree, or userspace ABI is changed.
- The diagnostics use the existing channel and VI runtime structures without changing stream setup or teardown.

## Validation

- JetPack 6.2.2 full driver build: PASS.
- `tegra-camera.ko` build with the new debugfs implementation: PASS.
- Build-log scan for compiler errors, undefined references, and duplicate definitions: PASS.
- Jetson 67 runtime debugfs and V4L2 regression validation: pending.

### DIFF
--- a/nvidia-oot/6.2/0012-Enable-Tegra-CSI-debugfs.patch
+++ b/nvidia-oot/6.2/0012-Enable-Tegra-CSI-debugfs.patch
@@ -1,5 +1,5 @@
 diff --git a/drivers/media/platform/tegra/camera/csi/csi.c b/drivers/media/platform/tegra/camera/csi/csi.c
-index 7e3dfb3a..fea546c5 100644
+index 7e3dfb3a..b3023a15 100644
 --- a/drivers/media/platform/tegra/camera/csi/csi.c
 +++ b/drivers/media/platform/tegra/camera/csi/csi.c
 @@ -7,7 +7,9 @@
@@ -21,7 +21,7 @@ index 7e3dfb3a..fea546c5 100644
  
  #include <media/media-entity.h>
  #include <media/v4l2-async.h>
-@@ -155,6 +159,191 @@ u64 read_mipi_clk_from_dt(struct tegra_csi_channel *chan)
+@@ -155,6 +159,390 @@ u64 read_mipi_clk_from_dt(struct tegra_csi_channel *chan)
  	return mipi_clk;
  }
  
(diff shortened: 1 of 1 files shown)


nvidia-oot/6.2/0012-Enable-Tegra-CSI-debugfs.patch
```diff
@@ -137,39 +157,213 @@ index 7e3dfb3a..fea546c5 100644
...
++	struct tegra_csi_device *csi = s->private;
++	struct tegra_csi_channel *chan;
++
++	if (!csi)
++		return 0;
++
++	seq_printf(s, "device: %s\n", dev_name(csi->dev));
++	seq_puts(s, "route fields marked runtime come from the bound VI channel\n");
++
++	list_for_each_entry(chan, &csi->csi_chans, list) {
```
ymodlin, line 224: The three new show handlers walk `csi->csi_chans` and dereference the bound `tegra_channel` without holding any lock, so a channel unregistering while a `cat` is in flight can be a use-after-free. `csi->source_update` already guards this list elsewhere — take it around the iteration (it is safe to sleep in a seq_file handler).

ymodlin (commented):
Automated review of the NVCSI debugfs diagnostics. Read-only dumps look sound (teardown via `tegra_csi_debugfs_remove()`, `of_node_put()` paths balanced, `0444` files with no user input to validate). One concurrency note.

PR #516, opened 2026-07-21
[Host][GMSL] Support D5xx flat NV12 over RAW8

polejoe:
## Summary

Add the D5xx product path for a flat NV12 surface carried over CSI-2 RAW8 without changing the existing stream-control register protocol.

- Keep the existing RGB `YUYV` format.
- Add `NV12` as the second RGB V4L2 format.
- Selecting NV12 uses the existing RGB stream DT field with packet DT `0x2A`.
- Capture `width x (height * 3 / 2)` RAW8 byte rows into one standard NV12 V4L2 buffer without app-side repacking.

No input-format I2C field or new control register is added.

## Layer contract

| Layer | Contract |
| --- | --- |
| HKR RGB surface | NV12 1280x960@60 |
| HIF MIPI packet | CSI2_DT `0x2A` RAW8, VC1 |
| Wire geometry | 1280 x 1440 byte rows, WC=1280 |
| Payload order | 960 Y rows followed by 480 interleaved-UV rows |
| Tegra VI | Match RAW8 and write all 1440 rows without byte reordering |
| Host V4L2 API | `V4L2_PIX_FMT_NV12` 1280x960, bytesperline=1280, sizeimage=1,843,200 |

For the separate CSI2_DT `0x1E` color path, HKR MC-FCVT and the Tegra receive layout use UYVY bytes, while the Host RGB video node continues to expose `YUYV`. The Host node therefore remains fixed to `YUYV / NV12`; this PR does not expose UYVY as a V4L2 format.

## Scope

- Update only the D5xx RGB format selection needed for YUYV/NV12.
- Add flat NV12 RAW8 capture patches for JetPack 6.0, 6.1, 6.2, 7.0, and 7.1.
- Do not change RAW10/BA10, Y12I, depth, IR, IMU, serializer, deserializer, device tree, deskew, or deployment scripts.

## Validation

- JetPack 6.2.2 full build: PASS.
- Jetson67 loaded `d5xx.ko` srcversion `574F511E6885E27482BEB02`.
- RGB V4L2 inventory: `YUYV` and `NV12`; no UYVY or BA10 is added by this PR.
- Complete enabled D5xx GMSL regression: PASS.
- Existing Depth Z16, IR Y8I, RGB YUYV, IMU 38x1, three-stream, and four-stream cases passed strict counters.
- RGB NV12 RAW8 1280x960@60 passed strict counters.

Artifacts: `/tmp/d5xx_full_nv12_only_pr516`

## Dependencies

- HKR HIF: https://github.com/RealSenseAI-Internal/soc-rs-soc-io-drivers/pull/460
- HKR tests/app: https://github.com/RealSenseAI-Internal/soc-rs-hkr-test/pull/325

### DIFF
--- a/kernel/realsense/d5xx.c
+++ b/kernel/realsense/d5xx.c
@@ -1315,7 +1315,7 @@ static const struct d5x_format d5x_y_formats_d58x[] = {
 
 static const struct d5x_format d5x_rgb_formats_d58x[] = {
 	{
-		/* MC-FCVT uses UYVY bytes on DT 0x1E; V4L2 exposes YUYV. */
+		/* CSI2_DT 0x1E uses MC-FCVT UYVY bytes; V4L2 exposes YUYV. */
 		.data_type = GMSL_CSI_DT_YUV422_8,
 		.mbus_code = MEDIA_BUS_FMT_YUYV8_1X16,
 		.n_resolutions = ARRAY_SIZE(d58x_rgb_sizes),
--- a/kernel/realsense/d5xx.c
+++ b/kernel/realsense/d5xx.c
@@ -1313,13 +1313,21 @@ static const struct d5x_format d5x_y_formats_d58x[] = {
 	},
 };
 
-static const struct d5x_format d5x_d58x_rgb_format = {
-	.data_type = GMSL_CSI_DT_YUV422_8,	/* UYVY */
-	.mbus_code = MEDIA_BUS_FMT_YUYV8_1X16,
-	.n_resolutions = ARRAY_SIZE(d58x_rgb_sizes),
-	.resolutions = d58x_rgb_sizes,
+static const struct d5x_format d5x_rgb_formats_d58x[] = {
+	{
+		/* MC-FCVT uses UYVY bytes on DT 0x1E; V4L2 exposes YUYV. */
+		.data_type = GMSL_CSI_DT_YUV422_8,
+		.mbus_code = MEDIA_BUS_FMT_YUYV8_1X16,
+		.n_resolutions = ARRAY_SIZE(d58x_rgb_sizes),
+		.resolutions = d58x_rgb_sizes,
+	}, {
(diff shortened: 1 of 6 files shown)


pinquanw (approved):

lemon6630 (approved):

kernel/realsense/d5xx.c
```diff
@@ -81,8 +81,11 @@ struct dser_interface {
 #else
 #include <media/gmsl-link.h>
 #define GMSL_CSI_DT_YUV422_8        0x1E
+#define GMSL_CSI_DT_YUV420_8        0x18
 #define GMSL_CSI_DT_RGB_888         0x24
 #define GMSL_CSI_DT_RAW_8           0x2A
+#define GMSL_CSI_DT_RAW_10          0x2B
```
ghu3: These fallback definitions are inside the non-SERDES branch, so they are not visible when CONFIG_VIDEO_D5XX_SERDES is enabled. Only the 6.0 and 6.2 patch sets currently add the new constants to gmsl-link.h; JP 6.1, 7.0, and 7.1 therefore fail to compile d5xx.c with GMSL_CSI_DT_RAW_16, GMSL_CSI_DT_RAW_10, and GMSL_CSI_DT_YUV420_8 undeclared. Please add equivalent GMSL header and VI capture-format patches for every supported patch set (or otherwise guard the common driver changes); adding only the constants would compile but would still leave RAW16/NV12 capture unsupported on those releases.

polejoe: Valid point. The updated branch now carries the matching GMSL DT definitions and RAW16/NV12 VI capture changes in all five supported patch sets: 6.0, 6.1, 6.2, 7.0, and 7.1. The NV12 product mapping was also corrected to use a RAW8 carrier instead of native YUV420 semantics. Patch application passes for all five sets, and the full JP6.2.2 module build and Jetson67 60 fps frame-integrity test pass.

PR #522, opened 2026-07-23
[RSDEV-13118] enlarge_GVD_control_size

ejgoldik:
- D4xx gvd size is actually 276B
- D5xx gvd size is actually 606B

This is a breaking change as SDK must allocate enough memory

### DIFF
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -132,6 +132,10 @@ struct ser_interface {
 #define DS5_DEVICE_TYPE_D43X		5
 #define DS5_DEVICE_TYPE_UNKNOWN		0
 
+/* GVD response payload size per product line */
+#define DS5_GVD_LEN_D4XX		276
+#define DS5_GVD_LEN_D5XX		606
+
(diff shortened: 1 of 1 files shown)


kernel/realsense/d4xx.c
```diff
@@ -3308,6 +3312,18 @@ static int ds5_get_calibration_data(struct ds5 *state, enum table_id id,
 	// get table length from fw
 	ret = ds5_raw_read(state, DS5_HWMC_RESP_LEN,
 			&table_length, sizeof(table_length)); /* Read response length */
+	if (ret) {
+		devm_kfree(&state->client->dev, cmd);
+		return ret;
+	}
+
+	if (table_length > length + 4) {
```
ymodlin, line 3320: Why magic numbers?

ymodlin (approved):

kernel/realsense/d4xx.c
ejgoldik: We will fix all magic numbers in a specific follow-up. In this case it is the HWMC header size

PR #525, opened 2026-07-23
[D585][GMSL][RGB Calib] Expose RAW16 Bayer10 surfaces

polejoe:
## Summary

- Keep private `GR0P` for byte-identical HKR `FMT_SGRBG10P` over UD30.
- Expose HKR `FMT_SGRBG10` over RAW16 as standard unpacked `BA10`.
- Optionally expose the same RAW16 receive path as standard packed `pgAA`.
- Stage each matching D5xx RAW16 frame in a cached scratch buffer, then perform a fused byte-order restore or row pack.
- Return postprocess failures to the generic VI adapter, which marks the V4L2 buffer erroneous instead of exposing mislabeled data.
- Keep existing NV12 and D4xx paths outside the D5xx-only callback.

## Format layers

- GR0P: HKR `FMT_SGRBG10P` -> VC1 / DT `0x30` / WC 2048 -> VI `T_R8` -> private `GR0P`, 2048 BPL
- BA10: HKR `FMT_SGRBG10` -> VC1 / DT `0x2E` / WC 3200 -> VI `T_R16` -> D5xx byte-order restore -> standard `BA10`, 3200 BPL
- pgAA: HKR `FMT_SGRBG10` -> VC1 / DT `0x2E` / WC 3200 -> VI `T_R16` -> D5xx row pack -> standard `pgAA`, 3200 BPL with 2000 active packed bytes per row

The optional VI frame-finish adapter remains generic. Only D5xx registers the callback, and the D5xx implementation returns immediately unless the complete private RAW16/Bayer media-bus tuple matches.

## Frame conversion

Direct repeated reads from write-combined VI memory reduced BA10 capture to approximately 3.5-6 fps. The cached scratch stage converts from normal cached memory and restores capture to 25 fps. The scratch allocation is lazy, reused by subsequent frames, and owned by the D5xx sensor instance.

## E2E dependencies

- https://github.com/RealSenseAI-Internal/soc-rs-hkr-apps/pull/229
- https://github.com/RealSenseAI-Internal/soc-rs-soc-device-mgr/pull/406
- https://github.com/RealSenseAI-Internal/soc-rs-soc-io-drivers/pull/470
- https://github.com/RealSenseAI-Internal/soc-rs-hkr-test/pull/332

## Validation

- JP6.0, JP6.1, JP6.2, JP7.0, and JP7.1 patch chains: clean application PASS
- JP6.2.2 driver build, package, deployment, loaded-module hash, and GMSL link verification: PASS
- Three retained-SerDes module reload cycles: PASS
- GR0P/UD30, BA10/RAW16, and pgAA/RAW16: 1600x1300@25 strict-counter E2E PASS
- BA10 capture-only: 25.1 fps, no sequence drops
- BA10 frame 20: 4,160,000 bytes; 2,080,000 little-endian samples; range 64-1023; zero samples above 10 bits
- pgAA: 25.1 fps, 3200 BPL with 2000 active packed bytes per row

The unfiltered suite still stops on a separate strict metadata gate: Depth metadata reports stream-ID mismatches and one RGB metadata frame-counter gap was observed. No tolerance was relaxed for this PR.

## References

- https://rsconf.realsenseai.com/display/RealSense/GMSL_MIPI_CSI/
- https://rsconf.realsenseai.com/display/RealSense/D5x+GMSL+RGB+RAW10+Host+Receiver+Wiki

### DIFF
--- a/nvidia-oot/7.1/0015-Add-D5x-GRBG10P-byte-surface.patch
+++ b/nvidia-oot/7.1/0015-Add-D5x-GRBG10P-byte-surface.patch
@@ -1,34 +1,37 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: zhou <shengyong.zhou@realsenseai.com>
 Date: Fri, 24 Jul 2026 19:30:00 +0800
-Subject: [PATCH] media: tegra: receive D5x GRBG10P over UD30
+Subject: [PATCH] media: tegra: receive D5x RGB Bayer surfaces
 
 D5x RGB ISYS stores FMT_SGRBG10P as 50 little-endian 10-bit pixels
 in each 64-byte cache line. HIF MIPI preserves the complete aligned
 row in a CSI User Defined 1 packet so the receiver treats the payload
 as an opaque byte stream.
 
-Add a private mbus/fourcc tuple for this product surface. VI matches
-wire DT 0x30, maps the User Defined byte stream to RAW8/T_R8, and writes
-the payload unchanged. The visible V4L2 geometry remains 1600x1300 while
-bytesperline and capture frame_x use the 2048-byte HKR stride.
+The optional unpacked path stores one 10-bit sample in each 16-bit
+container and sends those containers unchanged using CSI RAW16.
+
+Add separate internal mbus tuples for both transports. The packed path
+keeps its private GR0P ABI. The unpacked RAW16 path uses T_R16 and exposes
+the standard V4L2 SGRBG10/BA10 surface. Prefer formats advertised by the
+active subdevice when two transport tuples share a standard FourCC.
 
 Signed-off-by: zhou <shengyong.zhou@realsenseai.com>
 ---
 diff --git a/drivers/media/platform/tegra/camera/vi/vi5_fops.c b/drivers/media/platform/tegra/camera/vi/vi5_fops.c
-index 079b5865..fa0acd21 100644
+index 117a3040..067185b3 100644
 --- a/drivers/media/platform/tegra/camera/vi/vi5_fops.c
 +++ b/drivers/media/platform/tegra/camera/vi/vi5_fops.c
-@@ -394,6 +394,8 @@ static void vi5_setup_surface(struct tegra_channel *chan,
+@@ -392,6 +392,8 @@ static void vi5_setup_surface(struct tegra_channel *chan,
  	u32 bpl = chan->format.bytesperline;
  	u32 data_type = chan->fmtinfo->img_dt;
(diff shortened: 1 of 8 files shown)


polejoe:
Joint design and E2E verification review: http://zsy-nuc13.lan:3000/pr525-pr332-d5x-raw10-e2e

polejoe:
Joint product-flow and E2E review: http://zsy-nuc13.lan:3000/pr229-pr406-pr470-pr332-pr525-d5x-bayer10-e2e

nvidia-oot/7.1/0015-Add-D5x-GRBG10P-byte-surface.patch
```diff
@@ -0,0 +1,241 @@
...
++		return;
++
++	s_data = to_camera_common_data(chan->subdev_on_csi->dev);
++	if (!s_data || !s_data->frame_postprocess_ops ||
++	    !s_data->frame_postprocess_ops->process)
++		return;
++
++	data = vb2_plane_vaddr(vb, 0);
++	bytesused = vb2_get_plane_payload(vb, 0);
++	if (!data || !bytesused)
```
ghu3: The byte swap is mandatory for the advertised BA10 layout, but this silently returns the buffer unchanged whenever `vb2_plane_vaddr()` is unavailable (for example, an imported DMA-BUF that cannot be vmapped). The callback also has a `void` return, so the D5x postprocessor cannot propagate its scratch-allocation failure. In either case userspace receives CSI-order bytes labeled as valid BA10. Please make the conversion work for every supported VB2 memory type, or reject/mark the buffer as erroneous instead of silently completing it.

ghu3:
The current head (`bebfdd7`) still has a failing non-security check: **D4xx_Kernel_Module_Jetson_JP6** (build 2958), while the GitHub build matrix is green. Since the PR validation states that the JP6 patch chains and driver build pass, please reconcile or re-run this Jenkins result before merge.

nvidia-oot/7.1/0015-Add-D5x-GRBG10P-byte-surface.patch
polejoe: Fixed. The frame postprocess callback now returns an error status. If the VB2 plane has no CPU mapping, allocation fails, or the geometry is invalid, the VI buffer-finish adapter sets V4L2_BUF_FLAG_ERROR instead of exposing unconverted data as BA10. The D5xx conversion now uses a reusable scratch buffer and was verified at 1600x1300@25: 25.1 fps, zero sequence drops, and frame 20 contained 2,080,000 little-endian samples with range 64-1023 and no values above 10 bits.

pinquanw (approved):

lemon6630 (approved):

PR #528, opened 2026-07-27
Port D5xx MAX96724 tunnel mode to unified driver

pinquanw:
## Summary

Supersedes #517 with a narrower implementation rebased on the current `dev` branch.

- Integrates D5xx MAX96717 + MAX96724 tunnel mode through the existing unified `d4xx.c` framework. The sensor-driver diff is limited to 81 additions and 3 deletions.
- Keeps one JP6/JP7 MAX96717 implementation for both topologies: existing device trees default to MAX96712 pixel mode, while the new MAX96724 overlays opt into tunnel mode with `adi,tunnel-mode`.
- Uses the current `dev` MAX96717 driver as the implementation backbone. Its final diff is `+142/-20` (624 lines total versus 502 on `dev`), down from `+830/-191`.
- Preserves the validated MAX96717 + MAX96712 pixel-mode initialization, BPP selection, GPIO tunneling, and last-stream RX re-arm paths in place.
- Configures the MAX96717 tunnel register table, MIPI RX reset pulse, and video transmitter enable directly in the existing `max96717_init_settings()` path, matching the pixel-mode lifecycle. There is no separate `max96717_setup_streaming()` API or MAX96724-to-MAX96717 setup call.
- Removes the unconditional `CONFIG_VIDEO_D5XX_SERDES` pseudo-Kconfig define and the header-presence probe. The JP6/JP7 Kbuild entry applies `HAS_MAX96724` only to `d4xx.o`, alongside the companion `max96724.o` build entry; the JP5 Kconfig path does not set it and retains its existing MAX96712 path.
- Tunnel mode reuses the same pairing, control, GPIO, stream-state, initialization, and probe infrastructure. Only address reassignment, link setup, compact mode-specific register values, GPIO6, and TX retrigger remain mode-specific.
- Because tunnel mode carries the complete CSI-2 byte stream and preserves VC/DT end-to-end, it reuses the existing `set_pipe()` lifecycle without duplicating pixel-mode VC/DT mapping tables.
- Refactors MAX96724 around the existing MAX9296 lifecycle and callback structure. The driver is 1,491 lines, down from 2,643; the remaining size above the 1,156-line MAX9296 reference is primarily four-link lock/6Gbps handling, D5xx first-stream recovery, POC, and FSYNC.
- Uses one shared MAX96724 datapath configurator for initial setup, ONESHOT recovery, and stream retrigger; removes unused pixel-mode branches, debug exports, register definitions, and public APIs. MAX96712 remains the pixel-mode deserializer.
- Uses managed GPIO/regulator resources and NVIDIA conftest compatibility macros, with explicit error propagation and pipe/source bounds validation.
- Reuses the existing framework callbacks for source pairing, link/control setup, pipe allocation/configuration/release, stream state, reset recovery, D58x capabilities, and external-sync control.
- Adds the AGX and FG24 four-channel D5xx overlays plus JP6/JP7 build, patch, packaging, and install wiring.

This deliberately does not add a standalone `d5xx.c`/`d5xx.ko`, nor another source-overlay copy of `max96717.c`.

## Validation

- Applied the JetPack 6.2 patch set and built/linked `d4xx.ko`, `max96717.ko`, and `max96724.ko` for ARM64. The actual `d4xx.o` command contains target-scoped `-DHAS_MAX96724`, contains no `CONFIG_VIDEO_D5XX_SERDES`, and the resulting module retains all 18 MAX96724 dependencies.
- Compiled the same `d4xx.c` with the JetPack 5.0.2 kernel tree and GCC 9.3 without `HAS_MAX96724`; the resulting module has no MAX96724 references.
- Confirmed the built MAX96717 module exports `max96717_init_settings()` but no `max96717_setup_streaming()` symbol, and MAX96724 has no unresolved reference to the removed API.
- Verified the final `d4xx.c` diff against `dev` is `+81/-3`; its diff-only checkpatch result has no warnings or errors.
- Verified the final `max96717.c` diff against `dev` is `+142/-20`; its diff-only checkpatch result has no warnings or errors.
- Built the affected ARM64 modules with `W=1`; the modified interfaces introduce no new warnings. The reported warnings are pre-existing in `d4xx.c` and the NVIDIA camera-common header.
- Strict checkpatch for the target-scoped build capability refactor reports zero errors, warnings, or checks.
- Verified all modified embedded JP6/JP7 patch files parse successfully with `git apply --numstat`.
- Preprocessed and compiled both new DTS overlays into DTBO files. They retain the source branch's overlay-structure `dtc` warnings.
- Passed shell syntax checks for the modified build/install scripts and `git diff --check`.

The available generated NVIDIA source tree predates two MAX96712 exports already present on current `dev`, so external-module modpost used validation-only symbol entries for those two existing exports. No validation stubs are included in this PR.

Hardware validation is still required, so this PR remains a draft.

### DIFF
--- a/hardware/realsense/tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dts
+++ b/hardware/realsense/tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dts
@@ -33,7 +33,7 @@
  *   GPIO mux:     AON_GPIO(CC, 3)
  *   MAX96724 deser:  0x27
  *   MAX96717 ser:    0x40 (default), 0x60 (aliased)
- *   HKR camera SoC:  0x42 (def-addr, D585 3C)
+ *   HKR camera SoC:  0x10 (def-addr)
  *   Sensor aliases:   0x1a (Depth), 0x1b (RGB), 0x1c (IR), 0x1d (IMU)
  */
 
@@ -335,30 +335,19 @@
 							compatible = "adi,max96724";
 							csi-mode = "2x4";
 							max-src = <1>;
+							link-mask = <0x1>;
 							gmsl-link-speed = <6>;
 						};
 
 						/*
-						 * MAX96717 primary serializer at power-on address.
-						 * Used by max96717_setup_control() to move the
-						 * serializer from 0x40 to aliased 0x60.
+						 * MAX96717 serializer address is assigned by MAX96724
+						 * as 0x40 + link index (link A is 0x40).
+						 * Single 4-lane CSI-2 D-PHY input.
 						 */
-						ser_prim: max96717_prim@40 {
+						ser_a: max96717_a@40 {
 							status = "okay";
 							compatible = "adi,max96717";
 							reg = <0x40>;
-							adi,tunnel-mode;
-							is-prim-ser;
-						};
-
-						/*
-						 * MAX96717 Serializer alias after address reassignment.
-						 * Single 4-lane CSI-2 D-PHY input.
-						 */
-						ser_a: max96717_a@60 {
-							status = "okay";
-							compatible = "adi,max96717";
-							reg = <0x60>;
 							adi,tunnel-mode;
 							maxim,gmsl-dser-device = <&dser>;
 						};
@@ -369,7 +358,7 @@
 						 */
 						d5xx_depth: d5xx@1a {
 							status = "okay";
-							def-addr = <0x42>;
+							def-addr = <0x10>;
 							reg = <0x1a>;
 							compatible = "realsense,d5xx";
 							use_sensor_mode_id = "true";
@@ -448,7 +437,7 @@
 						 */
 						d5xx_rgb: d5xx@1b {
 							status = "okay";
-							def-addr = <0x42>;
+							def-addr = <0x10>;
 							reg = <0x1b>;
 							override_reg = <0x1a>;
 							compatible = "realsense,d5xx";
@@ -508,7 +497,7 @@
 						 */
 						d5xx_ir: d5xx@1c {
 							status = "okay";
-							def-addr = <0x42>;
+							def-addr = <0x10>;
 							reg = <0x1c>;
 							override_reg = <0x1a>;
 							compatible = "realsense,d5xx";
@@ -589,7 +578,7 @@
 						 */
 						d5xx_imu: d5xx@1d {
 							status = "okay";
-							def-addr = <0x42>;
+							def-addr = <0x10>;
 							reg = <0x1d>;
 							override_reg = <0x1a>;
 							compatible = "realsense,d5xx";
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -54,16 +54,11 @@ struct dser_interface {
 	 * MAX96717 serializers, which funnel all of a camera's VCs through one
 	 * pipe. Optional - NULL if the deserializer has no multi-VC support. */
 	int (*get_multi_vc_pipe_id)(struct device *dev, int vc_id);
-	/* Optional topology-specific serializer pipe mapping. */
-	int (*get_ser_pipe_id)(struct device *dev, int dser_pipe_id, int vc_id);
 	int (*bind_ser_to_dser_pipe)(struct device *dev, int dser_pipe_id, int ser_pipe_id, u32 vc_id);
 
 	int (*set_pipe)(struct device *dev, int pipe_id, u8 data_type1, u8 data_type2, u32 vc_id);
 	int (*release_pipe)(struct device *dev, int pipe_id);
 	void (*reset_oneshot)(struct device *dev);
-	void (*retrigger_datapath)(struct device *dev);
-	int (*get_active_pipe_count)(struct device *dev);
-	int (*get_link_locked)(struct device *dev);
 	/* RSDEV-12608: flush one link's line buffer after a camera HW reset;
 	 * NULL if the deser leaves no stale buffer (max9296). */
 	void (*reset_oneshot_link)(struct device *dev, u32 vc_id);
@@ -72,8 +67,6 @@ struct dser_interface {
 	int (*setup_link)(struct device *dev, struct device *s_dev);
 	int (*setup_control)(struct device *dev, struct device *s_dev);
 	int (*reset_control)(struct device *dev, struct device *s_dev);
-	int (*setup_fsync)(struct device *dev, u32 fps);
-	int (*disable_fsync)(struct device *dev);
 	
 	/* Device registration */
 	int (*sdev_register)(struct device *dev, struct gmsl_link_ctx *g_ctx);
@@ -83,7 +76,6 @@ struct dser_interface {
 	int (*power_on)(struct device *dev);
 	void (*power_off)(struct device *dev);
 	int (*init_settings)(struct device *dev);
-	int (*setup_streaming)(struct device *dev, struct device *s_dev);
 
 	/* Identification */
 	const char *name;
@@ -96,8 +88,6 @@ struct ser_interface {
 	/* Notify serializer a stream stopped; re-arms MIPI RX on last stop.
 	 * Optional - NULL if the serializer does not need it. */
 	int (*stream_stop)(struct device *dev, u32 vc_id);
-	void (*retrigger_tx)(struct device *dev);
-	int (*setup_streaming)(struct device *dev);
 
 	/* Setup and control */
 	int (*setup_control)(struct device *dev);
@@ -183,7 +173,6 @@ struct ser_interface {
 /* SerDes startup I2C readiness polling (defer probe if not responsive) */
 #define DS5_SERDES_STARTUP_TIMEOUT_MS 2000
 #define DS5_SERDES_STARTUP_RETRY_DELAY_MS 100
-#define DS5_DSER_REARM_COOLDOWN_MS 1800
 
 #define DS5_IMU_STREAM_DT		0x4040
 #define DS5_IMU_STREAM_MD		0x4042
@@ -516,7 +505,6 @@ struct ds5_sensor {
 	u16 pipe_data_type1;
 	u16 pipe_data_type2;
 	u32 pipe_vc_id;
-	unsigned int pipe_reapply_gen;
 	u16 cached_dt_value;
 	u16 cached_md_value;
 	u16 cached_override_value;
@@ -636,7 +624,6 @@ struct ds5_dev {
 	/* Pointer to the primary DS5 struct */
 	struct ds5 *ds5_primary;
 	bool serdes_setup_complete;
-	int sync_mode;
 
 	bool depth_streaming;
 	bool ir_streaming;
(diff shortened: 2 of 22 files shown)


nvidia-oot/max96717.c
```diff
@@ -104,13 +119,51 @@ int max96717_setup_control(struct device *dev)
 
 	g_ctx = priv->g_client.g_ctx;
 
+	if (!priv->pixel_mode) {
+		unsigned int val;
+		int retry;
+
+		if (prim_priv && priv != prim_priv &&
```
ejgoldik: Please don't assign device address here. The one that should assign the device address of the serializer is the max96724 deserializer (see how we do it in the max96712 deserializer). The logic there should be identical to max96712, meaning:
the deserializer loops over all links, turns them on one by one and assign an address to each serializer.

Therefore the "prim" serializer is not needed.

nvidia-oot/max96717.c
```diff
@@ -67,11 +77,16 @@ struct max96717 {
...
 	struct mutex lock;
+	bool pixel_mode;
 	/* bit[ser_vc_id] set while that pipe is streaming (set in set_pipe,
 	 * cleared in stream_stop). When it drops to 0 the MIPI RX is re-armed. */
 	u8 active_vc_mask;
+	u32 def_addr;
+	u32 pst2_ref;
 };
 
+static struct max96717 *prim_priv;
```
ejgoldik: Not needed, please see explanation below about serializer address assignment.

nvidia-oot/max96717.c
```diff
@@ -147,6 +200,11 @@ int max96717_enable_gpio_tunneling(struct device *dev)
 	max96717_write_acc(dev, priv->regmap, MAX96717_2C1_ADDR, MAX96717_2C1_ESYNC, &err);
 	max96717_write_acc(dev, priv->regmap, MAX96717_2C2_ADDR, MAX96717_2C2_ESYNC, &err);
 	max96717_write_acc(dev, priv->regmap, MAX96717_2C3_ADDR, MAX96717_2C3_ESYNC, &err);
+	if (!priv->pixel_mode) {
```
ejgoldik: Are these GPIOs really needed only for tunnel mode? Perhaps the pixel mode should be aligned. What do they do specifically?

nvidia-oot/max96717.c
```diff
@@ -188,6 +251,15 @@ int max96717_reset_control(struct device *dev)
 	}
 
 	priv->g_client.st_done = false;
+	if (!priv->pixel_mode && prim_priv && priv != prim_priv) {
+		if (prim_priv->pst2_ref)
```
ejgoldik: As written above, should be removed, address assignment should be done by the deserializer

nvidia-oot/max96724.c
```diff
@@ -0,0 +1,1491 @@
...
+	return max96724_wait_link_lock(dev, link);
+}
+
+int max96724_setup_link(struct device *dev, struct device *s_dev)
+{
+	struct max96724 *priv = dev_get_drvdata(dev);
+	unsigned int i;
+	int err;
+
+	mutex_lock(&priv->lock);
```
ejgoldik, line 689: This function is where the serializer address assignment should happen (see max96712_setup_link). Splitter mode is no longer relevant

hardware/realsense/tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dts
ejgoldik, line 1: The overlay name should be tegra234-camera-**d4xx**-overlay-fg24-4ch-**d5xx**.dts‎
I know it is very counter-intuitive, but the first "d4xx" is the driver name, not the camera name.
In the future it will be changed (along with the driver name) to rs-camera.
Same for the other overlay

kernel/realsense/d4xx.c
```diff
@@ -702,6 +705,25 @@ static const struct dser_interface max96712_interface = {
 	.name = "max96712",
 };
 
+#ifdef HAS_MAX96724
```
ejgoldik: Please don't add new ifdefs. The max96724 should be added to JP5x compilation as well.

ejgoldik (changes requested):

nvidia-oot/max96724.c
```diff
@@ -0,0 +1,1592 @@
...
+		if (regmap_read(priv->regmap, max96724_link_lock_addr(link),
+				&lock) ||
+		    !(lock & MAX96724_LINK_LOCKED))
+			return false;
+	}
+
+	return true;
+}
+
+int max96724_prepare_stream(struct device *dev, struct device *ser_dev)
```
ejgoldik: Don't think this function is needed. D4xx already has a place where it calls reset_oneshot for the specific link upon first stream start (see ds5_flush_idle_link). And the serializer has a place where it does rx_reset upon the first config and when the last stream closes (see max96717_mipi_rx_reset_pulse).

No need to aggregate these two things here

kernel/realsense/d4xx.c
```diff
@@ -4576,6 +4582,13 @@ static int ds5_serdes_setup(struct ds5 *state)
 		}
 	}
 
+	if (state->dser_ops == &max96724_interface) {
+		ret = max96724_finish_setup(state->dser_dev,
```
ejgoldik: why does it exist? Any boot time configuration should be in state->dser_ops->init_settings

ejgoldik (changes requested):

nvidia-oot/max96724.c
pinquanw: Agreed. Removed `max96724_prepare_stream()` and its call from `ds5_setup_pipeline()`, along with the now-unused `max96717_retrigger_tx()` helper. The stream path now relies on the existing D4xx/MAX96717 lifecycle instead of aggregating the deserializer and serializer reset operations here. Updated in `7994c776a`.

kernel/realsense/d4xx.c
pinquanw: Agreed. Removed `max96724_finish_setup()` and the MAX96724-specific call from `ds5_serdes_setup()`. The existing datapath setup and initial one-shot sequence now live in `max96724_init_settings()`, reached through `state->dser_ops->init_settings`, so the boot-time configuration follows the common deserializer lifecycle. Updated in `7994c776a`.

kernel/realsense/d4xx.c
```diff
@@ -5668,6 +5628,8 @@ static int ds5_mux_s_stream(struct v4l2_subdev *sd, int on)
 		msleep_range(100);
 #endif
 	}
+	if (on && ret >= 0 && state->dser_ops == &max96724_interface)
+		max96724_retrigger_datapath(state->dser_dev);
```
ejgoldik: What is this? And why is it outside the SERDES ifdef?

kernel/realsense/d4xx.c
```diff
@@ -5504,30 +5480,17 @@ static int ds5_mux_s_stream(struct v4l2_subdev *sd, int on)
...
-			/* FW reports this stream live without passing the consume
-			 * point; drop the arm so it cannot flush a live link. */
-			if (on)
-				state->ds5_dev->link_flush_pending = false;
-#endif
 			*streaming_flag = on;
 			mutex_unlock(&state->ds5_dev->lock);
 			sensor->streaming = on;
+			if (on && state->dser_ops == &max96724_interface)
+				max96724_retrigger_datapath(state->dser_dev);
```
ejgoldik: What is this? And why is it outside the SERDES ifdef?

kernel/realsense/d4xx.c
```diff
@@ -2664,6 +2675,13 @@ static int ds5_set_ser_esync_tunneling(struct ds5 *state, bool enable)
 	else
 		ret = state->ser_ops->disable_gpio_tunneling(state->ser_dev);
 
+	if (!ret && state->dser_ops == &max96724_interface) {
+		if (state->mux.last_set)
+			fps = state->mux.last_set->config.framerate;
```
ejgoldik, line 2686: Instead of asking explicitly about the interface and then calling max96724 function explicitly, better set the functions in the interface struct with NULL for the interfaces that do not have this function, and the here just check if (state->dser_ops->setup_fsync) call it. That is true for every place where you currently call max96724 functions directly

kernel/realsense/d4xx.c
```diff
@@ -4463,6 +4458,11 @@ static int ds5_gmsl_serdes_setup(struct ds5 *state)
 	mutex_lock(&serdes_lock__);
 
 	if (state->dser_primary) {
+		if (state->dser_ops == &max96724_interface) {
+			/* MAX96724 is already responsive; keep its shared I2C rail up. */
+			msleep(100);
+		} else
```
ejgoldik: This is for the first to bringup the deserializer, I do not suggest leaving it up, otherwise stray configs from earlier driver runs will persist.

kernel/realsense/d4xx.c
```diff
@@ -4483,7 +4484,10 @@ static int ds5_gmsl_serdes_setup(struct ds5 *state)
 			dev_err(dev, "gmsl deserializer link config failed\n");
 			goto error;
 		}
-		msleep(100);
+		if (state->dser_ops == &max96724_interface)
+			msleep(1000);
```
ejgoldik: Still needed? I just recently added another 600ms after power on (See 10 lines above)

ejgoldik (changes requested):

kernel/realsense/d4xx.c
pinquanw: This re-triggers the MAX96724 tunnel datapath after the camera has started producing CSI data; it is not another serializer RX reset. I moved it behind an optional retrigger_datapath callback in dser_interface and placed the call under CONFIG_VIDEO_D4XX_SERDES. The updated path was validated on FG24 with repeated stream start/stop and concurrent Depth + RGB streaming.

kernel/realsense/d4xx.c
pinquanw: Same fix here: the MAX96724-specific operation is now an optional dser_interface callback and this call site is inside CONFIG_VIDEO_D4XX_SERDES. This keeps the no-op/already-streaming path consistent without exposing a SerDes symbol in non-SerDes builds.

kernel/realsense/d4xx.c
pinquanw: Agreed. I added optional setup_fsync, disable_fsync, and retrigger_datapath callbacks to dser_interface, wired them only for MAX96724, and replaced the direct MAX96724 operational calls with NULL-checked interface callbacks.

kernel/realsense/d4xx.c
pinquanw: Agreed. I removed the MAX96724 keep-powered exception and restored the common deserializer power-off/power-on sequence, so stale state is not intentionally retained across driver bring-up.

kernel/realsense/d4xx.c
pinquanw: No longer needed. I removed the MAX96724-specific 1000 ms delay after setup_link and kept the common 100 ms delay; the existing post-power-on settling and serializer retry path are sufficient. This was validated on FG24 through probe and repeated/concurrent streaming.

ejgoldik (approved):

PR #533, opened 2026-07-28
CLAUDE.md: correct the GMSL flush invariant, add mandatory model/effort rule

Nikolai-L:
Docs-only companion to #532. Merge alongside or after it.

CLAUDE.md asserted the GMSL one-shot flush "is issued only from the HW-reset recovery path — never from `set_pipe`", and banned any link-cold-bring-up gating. RSDSO-21786 proved recovery-only is a P1 regression, so as written the file would lead the next reader to delete the new flush site as a bug.

## Changes

**Concurrency notes — RSDEV-12608 paragraph**
- Records the **two** sanctioned flush sites: post-HW-reset recovery, and link cold bring-up in `ds5_mux_s_stream()` between `ds5_configure()` and the FW stream start.
- States the real invariant: fire only when all four of the camera's `ds5_dev` streaming flags are clear; never inside `__max96712_set_pipe`/`__max96712_set_multi_vc_pipe`; never on a plain pipe realloc — flushing a live link truncates sibling in-flight frames (PIXEL_INCOMPLETE).
- Rewrites the `arm_link_reset`/`pending_link_reset`/`link_cold_bringup` ban as what it protects: **no serdes-side arm op, no flush in the max96712 `set_pipe` bodies**. Names the d4xx-side gate (`ds5_dev.link_flush_pending` + `ds5_flush_link_on_cold_bringup()`) as the sanctioned mechanism.
- Adds the RSDSO-21786 HW evidence so the site survives future review.

**Comment style** — its closing sentence still asserted the single-site invariant, which contradicted the corrected text two sections above. Also caps in-code comments at **2–3 lines**: state the why, and when one comment can't cover it, split into two short comments at the sites they explain rather than growing a paragraph above the function.

**SerDes pipe allocation** — corrects a verified-stale claim. The file said JP7 patches must be ported separately; on disk `nvidia-oot/7.0/0006` and 7.1's are **symlinks** to `6.0/0003`, so regenerating 0003 auto-ships JP7. Left uncorrected, the next SerDes change inherits the wrong carrier model.

**Post-patch instruction hygiene** — drops the reference to `.github/copilot-instructions.md`, which does not exist in this tree (`.github/` holds only `workflows/`), making half the rule unfollowable.

**New: model/effort recommendation (MANDATORY)** — every advised next step carries a model + reasoning-effort recommendation with a short why. Compound actions decomposed, justification tied to the failure mode rather than line count, cheap tiers named explicitly rather than omitted, delegation flagged separately.

**New: subagent and workflow economics** — measured on this branch's own review workflow: 20 agents ran at the session model because `model` was left unset, 10 of them per-finding refuters at high effort disproving low-severity nits. Rules: pin `model` **down**, never up — mechanical stages (refuting a stated claim, running a staged script, prose edits) get a cheaper tier; judgement stages (final judge, synthesis, root-cause) stay unset so they track whatever model the operator picked for the session, including a deliberate Fable run. Also: batch per-item fan-out instead of one verifier per finding, point agents at file paths instead of repeating a facts preamble, scale the fleet to the diff's blast radius, and never delegate waiting on a long job (cites `soak-watch`, which watched a 5.5 h soak for zero subagents).


CLAUDE.md
```diff
@@ -165,11 +168,23 @@ MAX96712 keys the sticky pipe by link (`link = vc_id / MAX9295_MAX_STREAMS`): `l
...
 - The multi-VC register table is validated on **link 0** only; link>0 extended-VC-msb (`MIPI_TX_EXT0..3 = link_id << 2`) is generalized to match the dynamic path but unproven — validate on hardware.
-- `max96712.c` has **no repo copy** — it is carried entirely by `nvidia-oot/6.0/0003-…patch` (JP6.1/6.2 symlink to 6.0; JP7 = `7.x/0006-…patch`). Edit the `sources_<v>/nvidia-oot/.../max96712.c` working copy, then regenerate 0003 via `git diff HEAD^ -- drivers/media/i2c/max96712.c` (0003 touches only this file). JP7 patches are **not** updated by JP6 work — port separately after validation.
+- `max96712.c` has **no repo copy** — it is carried entirely by `nvidia-oot/6.0/0003-…patch` (JP6.1/6.2 symlink to 6.0; JP7's `7.0/0006-…patch` and `7.1/0006-…patch` are themselves symlinks back to `6.0/0003`, verified on disk). Edit the `sources_<v>/nvidia-oot/.../max96712.c` working copy, then regenerate 0003 via `git diff HEAD^ -- drivers/media/i2c/max96712.c` (0003 touches only this file). Regenerating 0003 auto-ships JP7 too — there is no separate JP7 copy to port.
+
+## Model/effort recommendation (MANDATORY)
```
ymodlin, line 184: let's talk if we want this

CLAUDE.md
```diff
@@ -181,4 +196,6 @@ After every confirmed code patch, review both `.github/copilot-instructions.md`
...
+**In-code comments: 2–3 lines maximum.** State the *why* — never narrate what the code plainly does. If one comment cannot cover it, do **not** grow the block: split it into two short comments, each placed at the spot it explains (e.g. one at the constant/`#define`, one at the use site) rather than one paragraph parked above the function. Long rationale, HW evidence and cross-site invariants belong in the commit message, the PR description, or this file — not in a wall of `*` lines.
```
ymodlin, line 210: This is a good rule

ymodlin (approved):

PR #536, opened 2026-07-29
[D5xx] Expose RGB controls and telemetry in unified driver

pinquanw:
## Summary

Port the D5xx RGB-control and telemetry support from #527 to the unified `d4xx.c` driver introduced by #528.

- expose RGB brightness, contrast, and gamma as standard V4L2 controls
- expose SoC PVT, OHM, and projector temperatures plus the firmware error code as read-only telemetry controls
- expose Visual Preset at the expected CID using the existing control write path
- apply the D58x controls in both MAX96712 Pixel Mode and MAX96724 Tunnel Mode

## Unified-driver SKU handling

The unified driver also supports legacy D4xx devices. D58x SoC PVT temperature and the existing D4xx Readout Shaping control both use depth-window offset `0x30`, with different meanings.

`ds5_ctrl_init()` uses the device type validated and cached before V4L2 initialization:

- D58x: register the new D58x controls and do not register D4xx Readout Shaping
- D4xx: preserve Readout Shaping and do not register the D58x-only controls
- unknown device type: use the safe non-D58x registration path

Unsupported controls therefore never appear in `v4l2-ctl --list-controls` for D4xx devices.

## Relationship to existing PRs

Depends-On: #528 (merged)
Ports: #527

This PR is now based on `dev` after the stacked base PR was merged.

## Validation

- `git diff --check`
- JetPack 6.2 clean build with `./build_all.sh --clean 6.2`
- built, linked, and staged ARM64 `d4xx.ko`, `max96717.ko`, and `max96724.ko`
- resulting `d4xx.ko` SHA256: `516bbe42fb52980630e42a2cf30d410a1685ab78eea8817213bb7238b5be4711`
- deployed on FG24/MAX96724 Tunnel Mode with a D585; all D58x control nodes probed without kernel errors
- librealsense read/write/readback passed for brightness, contrast, gamma, and Visual Preset
- librealsense reads passed for SoC PVT, OHM, and projector temperatures; firmware Error Code read `0` through V4L2
- alternating emitter was validated through the MIPI V4L HWMC transport by sending the existing `SETSUBPRESET (0x7B)` pattern: three consecutive 1280x960@5 IR frames were laser-on/off/on (`76.17`, `44.53`, `76.16` mean intensity), with the two on frames differing by only `0.012`

`laser_power_on_off` is the scalar `RS2_OPTION_EMITTER_ENABLED` path (constant off/on). The per-frame `RS2_OPTION_EMITTER_ON_OFF` behavior uses the alternating-emitter HWMC subpreset and still needs to be registered by the D585 librealsense implementation.

### DIFF
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -206,6 +206,7 @@ struct ser_interface {
 #define DS5_EXPOSURE_ROI_LEFT		0x0014
 #define DS5_EXPOSURE_ROI_BOTTOM		0x0018
 #define DS5_EXPOSURE_ROI_RIGHT		0x001C
+#define DS5_VISUAL_PRESET		0x0020
 #define DS5_MANUAL_LASER_POWER		0x0024
 #define DS5_PWM_FREQUENCY		0x0028
 #define DS5_CAMERA_SYNC_MODE		0x002C
@@ -222,6 +223,18 @@ struct ser_interface {
 #define DS5_RGB_WHITE_BALANCE_TEMP	0x0018
 #define DS5_RGB_AUTO_WHITE_BALANCE	0x001C
 #define DS5_RGB_POWER_LINE_FREQ		0x0020
+#define DS5_RGB_BRIGHTNESS		0x0024
+#define DS5_RGB_CONTRAST		0x0028
+#define DS5_RGB_GAMMA			0x002C
+
+/* D58x-only telemetry offsets relative to DS5_DEPTH_CONTROL_BASE.
+ * DS5_SOC_PVT_TEMPERATURE overlaps the D4xx readout-shaping offset, so
+ * ds5_ctrl_init() registers exactly one interpretation per detected SKU.
+ */
+#define DS5_SOC_PVT_TEMPERATURE		0x0030
+#define DS5_OHM_TEMPERATURE		0x0034
+#define DS5_PROJECTOR_TEMPERATURE	0x0038
+#define DS5_ERROR_CODE			0x003C
 
 #define DS5_DEPTH_CONFIG_STATUS		0x4800
 #define DS5_RGB_CONFIG_STATUS		0x4802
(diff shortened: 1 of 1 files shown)


kernel/realsense/d4xx.c
```diff
@@ -905,6 +927,29 @@ static int ds5_read(struct ds5 *state, u16 reg, u16 *val)
 	return ret;
 }
 
+static int ds5_write_verify(struct ds5 *state, u16 reg, u16 val)
```
ymodlin: Why did you add this? I don't think that it's needed or related to the controls

pinquanw: Agreed. This helper was carried over from #527 only to verify the Visual Preset readback, but it is not needed here. I removed it and changed Visual Preset to use the existing ds5_write() path, consistent with the other controls. Updated in cc198353f.

ymodlin:
I think we will see the new controls also if ds5 camera is connected. this should be guarded.

kernel/realsense/d4xx.c
```diff
@@ -4714,6 +4889,30 @@ static int ds5_ctrl_init(struct ds5 *state, int sid)
 	if (sid == RGB_SID) {
 		struct v4l2_ctrl *ctrl;
 
+		ctrls->brightness = v4l2_ctrl_new_std(hdl, ops,
```
ejgoldik: All of these HKR specific ctrls should be guarded somehow so when the user does "v4l2-ctl -d /dev/video0 --list-controls" for a ds5 device, he won't see controls that aren't supported for their camera (this goes for all new ctrls)

ejgoldik (changes requested):

kernel/realsense/d4xx.c
pinquanw: Agreed. Fixed in 4aebb18ae: `ds5_ctrl_init()` now uses the device type already validated and cached before V4L2 initialization, and only registers brightness/contrast/gamma, Visual Preset, and the D58x telemetry controls for `DS5_DEVICE_TYPE_D58X`. Non-D58x/unknown devices never create these controls; D4xx Readout Shaping is registered only on the non-D58x path. I also deployed this version on FG24/D585 and verified the D58x controls through V4L2 and librealsense.

pinquanw:
@ymodlin Agreed; this is fixed in 4aebb18ae. The new controls are now guarded at registration time using the device type validated and cached before V4L2 initialization. They are created only for D58x, so a D4xx/DS5 device will not show them in `v4l2-ctl --list-controls`; the non-D58x path keeps the existing Readout Shaping control.

ejgoldik (approved):

PR #538, opened 2026-07-30
[RSDEV-13999][D58x][GMSL] Align IR mode and camera controls

polejoe:
Tracked-On: https://rsjira.realsenseai.com/browse/RSDEV-13399

## Problem

The unified `d4xx.c` driver had three D58x gaps that must be integrated and reviewed together:

- D58x interleaved IR Y8I at 1280x720@60 could inherit the mode0 1 GHz Tegra clock properties, enter NVCSI deskew, and panic in `nvcsi_deskew_setup`.

### IR mode selection

- Detect D58x through the cached runtime product type.
- Select mode0 for D58x Y8 and mode1 for D58x interleaved IR formats after `set_fmt`.
- Keep D4xx mode selection unchanged.

## Commit stack

1. `e46acbc` — D58x IR Tegra mode selection.

## Validation

IR mode selection was hardware validated by the original PR leg:

- Built JetPack 6.2.2 kernel/modules from `dev` without legacy `--d5xx --fg` flags.
- Deployed with the D5x MAX96724 tunnel overlay and verified the installed 5.15.185-tegra artifacts.
- Captured D58x Y8I 1280x720@60 with metadata; strict video/metadata counter checks passed and the NVCSI deskew panic did not recur.
- Ran the baseline E2E suite for Depth, IR, RGB, IMU, combined CIT profiles, SerDes, module reload, node inventory, and controls.

The combined three-commit PR additionally passed:

- JP6.2.1 ARM64 `d4xx.o` single-object compile.
- Strict checkpatch: 0 errors, 0 warnings, 0 checks.
- `git diff --check`.

D585 Sync Mode hardware validation remains pending; this PR does not claim runtime validation for the new `0/2/3` and stream/reset lifecycle paths.

## Follow-up scope

Remaining D5x features can continue as separate commits on this PR, with the accumulated E2E regression run before each additional feature is accepted.

PR #540, opened 2026-07-30
[RSDEV-13999][D58x][GMSL] Port D5x addtional streams to the unified driver

polejoe:
## Scope

This PR ports the remaining D58x GMSL features from the historical
`d500_gmsl_dev` branch to the unified `dev` driver. The four features are
kept in separate commits and were validated in order, with the accumulated
regression rerun before the next feature was added:

1. `cf5434b` — UVC CSI metadata sidecars
2. `8298630` — flat NV12 over CSI RAW8
3. `912fbd9` — native Y12I IR calibration
4. `558159b` — RGB calibration Bayer formats

The branch targets `dev` directly and does not depend on PR #538. The D58x
IR deskew issue is already handled on `dev` by the D58x tunnel-overlay fix
merged through PR #541.

The semantic audit against `d500_gmsl_dev` also confirmed that the other
D5x GMSL work is already present on `dev`, including the MAX96717/MAX96724
tunnel and FG24 support, 90 FPS modes, stream recovery/synchronization, DFU
handling, D58x controls/telemetry, and the D58x deskew policy.

## Latest `dev` rebase

`dev` advanced to `d7a77c6` when PR #542 added D58x RGB scalar controls.
Both branches had introduced `ds5_is_d58x()`, so the metadata commit
conflicted at that helper.

The resolution keeps all PR #542 control offsets, routing, Hue, and
power-line-frequency behavior; keeps one helper with the const-qualified
signature; and retains PR #542's `ds5_rgb_ctrl_offset()`. `git range-diff`
shows the later three feature commits as patch-equivalent. The metadata
commit differs only in reusing the helper already present on `dev`.

### Problem and root cause

D58x firmware can emit a shared 12-byte UVC metadata header with a runtime
payload length, while the unified host path retained Tegra's fixed D4xx
metadata format and size policy. Tegra VI had no per-sensor contract for
metadata format negotiation, queue capacity, or runtime `bytesused`
calculation.

The first carriers also used patch contexts from older NVIDIA baselines.
JP6.1 is no longer source-identical to JP6.0, and JP7.1 has its own current
source context.

### Fix

- Add per-sensor embedded-metadata operations to the JP6/JP7 Tegra carriers.
- Register them only for a detected D58x.
- Validate the UVC header, runtime payload length, and first metadata block.
- Retain legacy HKRM parsing for older D58x firmware.
- Preserve Tegra's original metadata behavior for D4x and sensors without
  registered metadata operations.
- Use version-specific carriers where NVIDIA source baselines differ.

### Problem and root cause

D58x firmware can send a standard NV12 image as a flat CSI RAW8 payload, but
the unified D58x RGB path exposed only YUYV. Tegra had no D58x NV12 mapping
or capture geometry for `width x (height * 3 / 2)`, and its image properties
remained tied to the fixed YUYV bit depth after a format change.

### Fix

- Advertise NV12 only from the detected D58x RGB format table.
- Map it to CSI RAW8 with `MEDIA_BUS_FMT_UYYVYY8_0_5X24`.
- Update Tegra image properties and MIPI clock from the selected YUYV/NV12
  format.
- Add baseline-specific VI carriers that capture the flat RAW8 surface while
  retaining the standard userspace NV12 buffer ABI.

### Problem and root cause

D58x IR calibration firmware emits native Y12I: little-endian 16-bit left
and right slots with 12 valid bits per sample. The unified driver still
exposed this through a legacy 24-bit RGB carrier and lacked D58x Y12I
transport tuples, VI byte-surface handling, and optional RAW16 byte-order
restoration. Tegra also selected the first Y16I tuple instead of the
transport advertised by the active subdevice.

The original JP7.1/JP7.2 symlinks reused a shared carrier whose patch context
does not match those baselines:

- JP7.1 and JP7.2 already define `TEGRA_IMAGE_DT_RAW16`.
- JP7.2 conditionally wraps the legacy vb2 wait callbacks for Linux 7.0, so
  the `channel.c` queue layout differs.

These were patch-application failures, not IR data-path failures.

### Fix

- Use opaque UD30/T_R8 as the default byte-transparent Y12I carrier exposed
  through the standard Y16I ABI.
- Retain RAW16/T_R16 as an explicit diagnostic fallback and restore its slot
  byte order in a D58x-only frame-finish callback.
- Prefer the active subdevice's advertised transport when multiple VI tuples
  share Y16I.
- Register frame post-processing only for detected D58x devices.
- Replace the JP7.1 and JP7.2 symlinks with independent carriers aligned to
  `jetson_38.4` and `jetson_39.2.0`.

### Problem and root cause

D58x RGB calibration firmware can emit unpacked GRBG10 and native
cache-line-packed GRBG10P, but the unified driver exposed only YUYV/NV12.
The D58x RGB table had no calibration profiles or transport tuples, Tegra VI
lacked the required RAW16/UD30 byte surfaces, and the shared mux enumerator
remapped format indices after the first three entries through the legacy
IR/depth table.

### Fix

- Add D58x-only 1600x1300@25 BA10, pgAA, and GR0P formats.
- Receive BA10 and pgAA over RAW16/T_R16, restore the container byte order,
  and pack pgAA rows after VI completion.
- Receive GR0P over UD30/T_R8 while preserving its 64-byte cache-line ABI.
- Select the complete per-stream format table before applying the legacy
  shared IR/depth mux fallback.
- Add compatible OOT carriers for every JP6/JP7 baseline.

## D4x compatibility

- All new stream formats are selected only after
  `DS5_DEVICE_TYPE_D58X` detection.
- D4x format tables, media-bus codes, CSI data types, and stream paths are
  unchanged.
- D58x metadata and frame post-processing operations are registered only for
  D58x. D4x keeps the existing Tegra metadata fallback and has no frame
  post-processing callback.
- Duplicate-FOURCC lookup prefers an actively advertised transport but keeps
  the previous first-match result as its fallback.
- This PR changes no DTS. D58x validation uses
  `tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dtbo`.
- CI covers all supported JetPack build baselines. The hardware E2E run used
  D58x; a physical D4x hardware run is not claimed.

### Per-feature checkpoints

- Metadata: clean JP6.2.2 full kernel/OOT build, deployment, reboot, baseline
  Depth/IR/RGB/IMU, node/control/SERDES, retained-SERDES reload, and
  multi-stream E2E passed.
- NV12: complete ten-version CI passed. NV12 1280x960@60 captured 180/180
  frames before and after three retained-SERDES reload cycles, with exact
  sequence, a 1,843,200-byte standard NV12 buffer, zero new VI capture
  errors, and a successful 180-frame YUYV restore. The accumulated
  Depth/IR/RGB/IMU and multi-stream E2E passed.
- IR calibration: complete ten-version CI passed, including the corrected
  JP7.1 and JP7.2 carriers. Default UD30 Y16I captured 180/180 frames at
  1600x1300@15 twice, including after three retained-SERDES reload cycles,
  with exact sequence and zero correctable/uncorrectable error delta. The
  accumulated Depth/IR/RGB/NV12/IMU and multi-stream E2E passed.
- RGB calibration: every carrier applied to its fully patched NVIDIA
  baseline. BA10, pgAA, and GR0P each captured 180/180 frames at
  1600x1300@25 with exact sequence and zero kernel capture-error delta.
  Payload checks decoded non-zero 10-bit Bayer samples; all GR0P cache-line
  padding remained zero.

### Final rebased head on Jetson 67

- Rebased directly onto current `dev` (`d7a77c6`); current head is
  `558159b`.
- The exact rebased head passed the complete GitHub Actions matrix:
  **10/10 PASS** for JP5.0.2, JP5.1.2, JP5.1.6, JP6.0, JP6.1, JP6.2,
  JP6.2.2, JP7.0, JP7.1, and JP7.2. JP7.1 and JP7.2 both completed patch
  application and the full carrier build successfully.
- Reused the existing `sources_6.2.2` tree; no kernel sources were downloaded.
- Clean full JP6.2.2 kernel/OOT build passed on the final functional tree.
- Exact rebased HEAD modules-only rebuild, deployment, and reboot passed
  artifact SHA256, vermagic, module loading, boot, D5x FG24 tunnel overlay,
  and MAX96724 A/B link-lock checks.
- Final 180-frame feature matrix: **PASS** for NV12, default UD30 Y16I, BA10,
  pgAA, GR0P, YUYV restore, Depth Z16, IR Y8I restore, and IMU. Every case
  returned exact sequence `0..179`; kernel capture-error delta was zero.
- Applicable full E2E: **20/20 PASS** in 97.95 seconds:
  - 8/8 Depth+RGB concurrency profiles
  - 8/8 Depth+RGB+IR concurrency profiles
  - stream start/stop
  - metadata arrival, parsing, and hardware timestamp progression
- Post-test Jetson kernel log contained zero capture-error matches. The D5x
  serial test window contained no ASSERT, PANIC, FATAL, HardFault, or
  watchdog event.

## Known non-default diagnostic behavior

The default and release-gated IR calibration transport is UD30 and passes
without capture errors. The explicit RAW16 diagnostic fallback produces
continuous monotonic Y16I buffers on packet DT `0x2e`, but Tegra marks every
buffer with the known correctable `err_data 16384` flag. The final target and
all final regressions were restored to the default UD30 path.

The generic repository V4L2 package also contains D4x-only control tests,
firmware-policy assumptions, an ARM64 `v4l2_format` layout issue in its test
harness, per-frame 5% timing assertions, and legacy metadata field offsets.
Those generic-test limitations are not used to waive a driver-path failure;
the D58x applicable full E2E and the dedicated long-run feature matrix above
are the hardware gates for this PR.


lemon6630 (approved):

pinquanw (approved):

ymodlin:
Why did you merge this without our approval??

will be removed ASAP

PR #543, opened 2026-07-31
[RSDEV-12730][D58X][GMSL] Harden exposure control

lemon6630:
## What

Harden D4xx/D58x Exposure readback while preserving the existing userspace ABI:

- keep `V4L2_CID_EXPOSURE_ABSOLUTE` as a standard scalar INTEGER control
- serialize Exposure GET with device-level SET and stream operations
- read a complete MSW-first / LSW-second FW snapshot
- restart the complete pair after a transport error or an invalid combined value
- use a fixed 15-attempt transport budget; remove the 1,000 ms wall-clock retry
- publish only complete values inside the advertised Exposure range

## Why

The old Host path ignored both register-read return values and assembled whatever
was left in the two half-word variables. That allowed corrupt values such as
2147483648 to reach userspace.

Reading Exposure MSW starts the FW snapshot and LSW completes that same snapshot,
so a failed or invalid pair must restart from MSW. Normal cached reads with the
FW change below complete in about 4–6 ms. D585 testing still found transient GMSL
transport contention during SET/stream overlap: one successful test needed 12
non-publishable attempts (`-EREMOTEIO` plus an out-of-range `0xFF` pair) before
the next complete pair succeeded. A bounded Host transport retry is therefore
still needed even with the new FW, but it is not used for cache freshness or
old-FW compatibility.

A true one-element `V4L2_CTRL_TYPE_U32` control is pointer-backed in V4L2 and
rejects the repository's existing streamApp and librealsense `size=0` scalar
requests before the driver callback. The supported Depth 1..200000 and RGB
1..10000 ranges fit in signed 32 bits, so the standard INTEGER control preserves
the contract for D4xx and D58x without an ABI split.

## Firmware dependency

Companion FW PR:
[RealSenseAI-Internal/soc-rs-soc-device-mgr#430](https://github.com/RealSenseAI-Internal/soc-rs-soc-device-mgr/pull/430)

That PR moves synchronous Application GET out of the I2C response path and
serves mapped reads from a per-camera cache/shadow, with asynchronous refresh
and SET write-through. It removes the previous 600–940 ms normal startup GET
tail. The Host retry in this PR handles only the remaining transport/snapshot
integrity failures; it cannot and does not provide cache freshness.

The attached Jetson has a D585, so no separate D4xx hardware smoke test was run.
D4xx remains on the standard scalar V4L2 control ABI it used before this PR.

## Validation

- strict full-PR checkpatch: 0 errors, 0 warnings, 0 checks
- full clean JP6.2.1 NVIDIA OOT build: PASS
- final targeted NVIDIA OOT rebuild: PASS
- final module:
  - `srcversion=7019F13C69197CA259799F5`
  - `vermagic=5.15.148-tegra`
  - SHA-256 `15f276b2e7d5feefef6350e150172247818f97641057caa91d7ee52795915a8b`
- Depth startup: 5 streams, 400/400 valid Exposure reads, zero read/stream
  failures, maximum GET 5.824 ms
- RGB:
  - idle 200/200 valid reads, maximum GET 5.952 ms
  - 5 stream starts, 400/400 valid reads, zero read/stream failures, maximum
    GET 5.844 ms
- Depth Exposure boundary SET/GET: 1, 10000, 200000 — 3/3 exact
- live Depth stream: 100 Exposure+Gain SET/GET iterations, zero SET/GET/stream
  failures; longest transport-contended GET 480.204 ms and maximum convergence
  485.011 ms
- Depth+IR mixed stress at 1280x960@30:
  - 600 Depth + 600 IR frames
  - zero sequence gaps and zero intervals over one second
  - 240 mixed AE/Exposure/Gain operations with readback, zero failures
- no terminal 15-attempt Exposure error, kernel warning, timeout, or call trace
  with the final module
- controls restored to Exposure 10000 on Depth and RGB

The reported frame-stop symptom was not reproduced with the current FW/Host
pair.

### DIFF
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -296,6 +296,7 @@ enum ds5_mux_pad {
 /* I2C retry configuration */
 #define DS5_I2C_RETRY_COUNT	5
 #define DS5_I2C_RETRY_DELAY_US	5000
+#define DS5_EXPOSURE_READ_RETRY_COUNT	3
 
 /* DFU definition section */
 #define DFU_MAGIC_NUMBER "/0x01/0x02/0x03/0x04"
@@ -3502,6 +3503,7 @@ static int ds5_g_volatile_ctrl(struct v4l2_ctrl *ctrl)
 			
 	u32 data;
 	int ret = 0;
+	int retry;
 	struct ds5_sensor *sensor = (struct ds5_sensor *)ctrl->priv;
 	u16 base;
 	u16 reg;
@@ -3559,11 +3561,28 @@ static int ds5_g_volatile_ctrl(struct v4l2_ctrl *ctrl)
 		if (state->is_imu)
 			return -EINVAL;
 		/* see ds5_hw_set_exposure */
-		ds5_read(state, base | DS5_MANUAL_EXPOSURE_MSB, &reg);
-		data = ((u32)reg << 16) & 0xffff0000;
-		ds5_read(state, base | DS5_MANUAL_EXPOSURE_LSB, &reg);
-		data |= reg;
-		*ctrl->p_new.p_u32 = data;
+		for (retry = 0; retry < DS5_EXPOSURE_READ_RETRY_COUNT; retry++) {
+			ret = ds5_read(state,
+				       base | DS5_MANUAL_EXPOSURE_MSB, &reg);
+			if (!ret) {
+				data = ((u32)reg << 16) & 0xffff0000;
+				ret = ds5_read(state,
+					       base | DS5_MANUAL_EXPOSURE_LSB,
+					       &reg);
+			}
+			if (!ret) {
+				data |= reg;
+				if (data >= ctrl->minimum &&
+				    data <= ctrl->maximum) {
+					*ctrl->p_new.p_u32 = data;
+					break;
+				}
+				ret = -ERANGE;
+			}
+			if (retry < DS5_EXPOSURE_READ_RETRY_COUNT - 1)
+				usleep_range(DS5_I2C_RETRY_DELAY_US,
+					     DS5_I2C_RETRY_DELAY_US + 500);
+		}
 		break;
 
 	case V4L2_CID_BRIGHTNESS:
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -3033,7 +3033,11 @@ static int ds5_s_ctrl(struct v4l2_ctrl *ctrl)
 		break;
 
 	case V4L2_CID_EXPOSURE_ABSOLUTE:
-		ret = ds5_hw_set_exposure(state, base, ctrl->val);
+		if (!ctrl->p_new.p_u32)
+			ret = -EINVAL;
+		else
+			ret = ds5_hw_set_exposure(state, base,
+						  *ctrl->p_new.p_u32);
 		break;
 	case V4L2_CID_BRIGHTNESS:
 		if (state->is_rgb)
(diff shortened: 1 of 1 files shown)


polejoe (approved):

kernel/realsense/d4xx.c
```diff
@@ -3558,12 +3613,13 @@ static int ds5_g_volatile_ctrl(struct v4l2_ctrl *ctrl)
 	case V4L2_CID_EXPOSURE_ABSOLUTE:
 		if (state->is_imu)
 			return -EINVAL;
-		/* see ds5_hw_set_exposure */
-		ds5_read(state, base | DS5_MANUAL_EXPOSURE_MSB, &reg);
-		data = ((u32)reg << 16) & 0xffff0000;
-		ds5_read(state, base | DS5_MANUAL_EXPOSURE_LSB, &reg);
-		data |= reg;
-		*ctrl->p_new.p_u32 = data;
+		mutex_lock(&state->lock);
```
ymodlin, line 3616: **Potential BUG — worst case ~75 ms of I2C retries while holding `state->lock`.**

`ds5_read_exposure()` retries 15 times with a 5 ms sleep between attempts, all inside `state->lock`. That same mutex is held across the whole `ds5_s_ctrl()` switch (line 3078) and the stream/DFU paths, so a librealsense poll of exposure can stall a concurrent control write or stream start — and an out-of-range readback (e.g. FW reporting 0 while AE is active or before streaming) burns the full budget on every read and then fails `G_EXT_CTRLS` with `-ERANGE`. Suggest dropping the retry count (2–3 is enough for a torn MSW/LSW pair) and taking the lock per pair rather than around the whole loop.

kernel/realsense/d4xx.c
```diff
@@ -963,6 +964,60 @@ static int ds5_read_poll(struct ds5 *state, u16 reg, u16 *val)
...
+				__func__, attempt + 1, ret);
+		}
+
+		if (attempt < DS5_EXPOSURE_READ_RETRY_COUNT - 1)
+			usleep_range(DS5_I2C_RETRY_DELAY_US,
+				     DS5_I2C_RETRY_DELAY_US + 500);
+	}
+
+	dev_err(&state->client->dev,
+		"%s(): failed after %u attempts, err %d\n",
```
ymodlin, line 1015: `dev_err` on a volatile getter is log spam: librealsense polls exposure continuously, so a device that transiently reports out-of-range prints on every read. `ds5_read_poll()` was chosen precisely to avoid this (CLAUDE.md); use `dev_dbg` or `dev_err_ratelimited`.

kernel/realsense/d4xx.c
```diff
@@ -4970,12 +5026,10 @@ static int ds5_ctrl_init(struct ds5 *state, int sid)
 					1, MAX_RGB_EXP, 1, DEF_RGB_EXP);
 	}
 
-	if ((ctrls->exposure) && (sid >= DEPTH_SID && sid < IMU_SID)) {
+	if (ctrls->exposure) {
```
ymodlin, line 5029: Dropping `type = V4L2_CTRL_TYPE_U32` is a userspace-visible ABI change (compound/has-payload → scalar). It is self-consistent here — `ds5_s_ctrl()` reads `ctrl->val`, and the in-repo consumers (`streamApp` `Stream.cpp` and `test_controls.py`) use the scalar `v4l2_ext_control.value` with `size = 0`, which only works for the scalar form. Please confirm the librealsense MIPI backend also reads it scalar before this lands, and note that removing the `sid < IMU_SID` guard is safe only because the IMU sid no longer creates this control (`else if (sid == RGB_SID)`).

ymodlin (commented):
Automated review of the exposure-hardening change (3 comments). Direction looks right — `ds5_read_poll()` is the correct primitive here per CLAUDE.md, and making the control a plain scalar matches how `ds5_s_ctrl()` already consumes it via `ctrl->val`.

PR #544, opened 2026-07-31
RW16--[D5xx][GMSL] Expose Bayer10 calibration surfaces over UD30

polejoe:
## Summary

Expose the two D5xx RGB calibration surfaces without a frame-wide software conversion:

| V4L2 surface | HKR input | Source `data_type` | Packet override | VI tuple | BPL / sizeimage |
|---|---|---:|---:|---|---:|
| RW16 | unpacked GRBG10, 16-bit LE slots | `0x2e` | `0x30` | UD30 + T_R8 | 3200 / 4,160,000 |
| GR0P | private GRBG10P CL64 | `0x30` | `0x30` | UD30 + T_R8 | 2048 / 2,662,400 |

- Add the RGB packet-DT override register at `0x403c`.
- Use `data_type` only as the HKR source selector and `override_dt` as the final packet DT.
- Keep the logical 1600x1300@25 geometry while preserving every carrier byte.
- Remove the pgAA format entry and D5xx frame packing callback.
- Keep RW16 as the existing SDK-compatible RAW16 FourCC and GR0P as the private packed surface.

No SerDes control behavior or generic D4xx format path is changed.

## Validation

- JetPack 6.2.2 full kernel/NVIDIA OOT build: PASS
- Deployment verification on Jetson 67: module SHA256 and boot checks PASS
- Targeted E2E on D5x tunnel mode:
  - GR0P 1600x1300@25: 180 frames, 0 sequence drops
  - RW16 1600x1300@25: 180 frames, 0 sequence drops
- Full enabled GMSL regression: 15/15 stages/cases PASS
- HKR serial log confirms the explicit source/override/packet tuples.

Related: soc-rs-soc-io-drivers#492, soc-rs-soc-device-mgr#431, soc-rs-hkr-test#343


pinquanw (approved):

kernel/realsense/d5xx.c
```diff
@@ -2427,12 +2368,13 @@ static int d5x_configure(struct d5x *state)
 	}
 
 	if (override_addr != 0) {
-		dt_value = sensor->config.format->data_type;
-		if (sensor->cached_override_value != dt_value) {
-			ret = d5x_write(state, override_addr, dt_value);
+		override_value = state->is_rgb ?
```
ymodlin, line 2371: **Potential BUG — writes 0 to the override register for plain RGB formats.**

The RGB branch uses `format->override_dt` directly, but only the two RAW/UD30 entries in `d5x_rgb_formats_d58x[]` set `override_dt`; the YUYV and NV12 entries leave it 0. So selecting YUYV/NV12 now writes `0x00` to `override_addr`, where the old code wrote `dt_value`. This is the exact hazard the comment 30 lines above warns about ("avoids overwriting a correct DT with 0 (which caused INVALID_DT on subsequent attempts)"). `packet_dt` already resolves to `override_dt ?: data_type`, so using `packet_dt` unconditionally — or guarding the RGB branch with `override_dt != 0` — fixes it.

ymodlin (commented):
Automated review of the UD30 / Bayer10 calibration change. One functional finding on the override register; the `packet_dt` derivation, the T_R8 `frame_x`/bytesperline math and the `match.datatype = USER_1` pairing all look consistent.

PR #542, opened 2026-07-30
[RSDEV-12730][D58X][GMSL] Expose RGB scalar controls

lemon6630:
Tracked-On: https://rsjira.realsenseai.com/browse/RSDEV-12730

## Summary

- expose the intended D58X RGB scalar controls on the unified `d4xx` driver
- select the D58X RGB extension offsets from the runtime device type while preserving the existing D4xx control ABI
- expose Hue only for D58X and constrain its power-line-frequency menu to the supported Disabled/50 Hz/60 Hz values

## Root cause

The unified driver used the D4xx RGB control offsets and menu contract for D58X. On D58X, the scalar controls live in a separate RGB extension window; the same offsets in the depth block carry temperature and error telemetry. Hue is D58X-specific, and the D58X power-line-frequency control does not support the Auto menu item.

## Validation

- `git diff --check`: PASS
- strict kernel checkpatch on the code diff: 0 errors, 0 warnings, 0 checks

## D58X RGB scalar validation

The seven controls were tested on the real RGB alias `/dev/v4l-subdev16` with SC1.2 FreeRTOS build 13023.

| Control | Advertised contract | Values tested |
| --- | --- | --- |
| Saturation | 0..100, default 64 | 0, 100, 75 |
| Sharpness | 0..100, default 50 | 0, 100, 35 |
| White Balance Temperature | 2800..6500, step 10, default 4600 | 2800, 6500, 5000 |
| Hue | -180..180, default 0 | -180, 180, -90 |
| Auto White Balance | Boolean, default 1 | 0, 1, 0 |
| Power Line Frequency | Disabled/50 Hz/60 Hz, default 60 Hz | 0, 1, 2 |
| AE Priority | Boolean, default 0 | 0, 1 |

- valid SET/volatile-GET checks: 20/20 PASS
- all seven representative values survived Host module rebind: 7/7 PASS
- invalid public values followed V4L2 policy: integer/Boolean values were clamped or step-aligned, while invalid Power Line menu item 3 returned `ERANGE` and retained mode 2
- all seven controls were restored to their advertised defaults after testing

PR #549, opened 2026-08-03
[RSDEV-13999][D58x][GMSL] Port D5x addtional streams to the unified driver

pinquanw:
Scope
This PR ports the remaining D58x GMSL features from the historical
d500_gmsl_dev branch to the unified dev driver. The four features are
kept in separate commits and were validated in order, with the accumulated
regression rerun before the next feature was added:

cf5434b — UVC CSI metadata sidecars
8298630 — flat NV12 over CSI RAW8
912fbd9 — native Y12I IR calibration
558159b — RGB calibration Bayer formats
The branch targets dev directly and does not depend on PR https://github.com/realsenseai/realsense_mipi_platform_driver/pull/538. The D58x
IR deskew issue is already handled on dev by the D58x tunnel-overlay fix
merged through PR https://github.com/realsenseai/realsense_mipi_platform_driver/pull/541.

The semantic audit against d500_gmsl_dev also confirmed that the other
D5x GMSL work is already present on dev, including the MAX96717/MAX96724
tunnel and FG24 support, 90 FPS modes, stream recovery/synchronization, DFU
handling, D58x controls/telemetry, and the D58x deskew policy.

Latest dev rebase
dev advanced to d7a77c6 when PR https://github.com/realsenseai/realsense_mipi_platform_driver/pull/542 added D58x RGB scalar controls.
Both branches had introduced ds5_is_d58x(), so the metadata commit
conflicted at that helper.

The resolution keeps all PR https://github.com/realsenseai/realsense_mipi_platform_driver/pull/542 control offsets, routing, Hue, and
power-line-frequency behavior; keeps one helper with the const-qualified
signature; and retains PR https://github.com/realsenseai/realsense_mipi_platform_driver/pull/542's ds5_rgb_ctrl_offset(). git range-diff
shows the later three feature commits as patch-equivalent. The metadata
commit differs only in reusing the helper already present on dev.

Metadata sidecars
Problem and root cause
D58x firmware can emit a shared 12-byte UVC metadata header with a runtime
payload length, while the unified host path retained Tegra's fixed D4xx
metadata format and size policy. Tegra VI had no per-sensor contract for
metadata format negotiation, queue capacity, or runtime bytesused
calculation.

The first carriers also used patch contexts from older NVIDIA baselines.
JP6.1 is no longer source-identical to JP6.0, and JP7.1 has its own current
source context.

Fix
Add per-sensor embedded-metadata operations to the JP6/JP7 Tegra carriers.
Register them only for a detected D58x.
Validate the UVC header, runtime payload length, and first metadata block.
Retain legacy HKRM parsing for older D58x firmware.
Preserve Tegra's original metadata behavior for D4x and sensors without
registered metadata operations.
Use version-specific carriers where NVIDIA source baselines differ.
NV12
Problem and root cause
D58x firmware can send a standard NV12 image as a flat CSI RAW8 payload, but
the unified D58x RGB path exposed only YUYV. Tegra had no D58x NV12 mapping
or capture geometry for width x (height * 3 / 2), and its image properties
remained tied to the fixed YUYV bit depth after a format change.

Fix
Advertise NV12 only from the detected D58x RGB format table.
Map it to CSI RAW8 with MEDIA_BUS_FMT_UYYVYY8_0_5X24.
Update Tegra image properties and MIPI clock from the selected YUYV/NV12
format.
Add baseline-specific VI carriers that capture the flat RAW8 surface while
retaining the standard userspace NV12 buffer ABI.
IR calibration
Problem and root cause
D58x IR calibration firmware emits native Y12I: little-endian 16-bit left
and right slots with 12 valid bits per sample. The unified driver still
exposed this through a legacy 24-bit RGB carrier and lacked D58x Y12I
transport tuples, VI byte-surface handling, and optional RAW16 byte-order
restoration. Tegra also selected the first Y16I tuple instead of the
transport advertised by the active subdevice.

The original JP7.1/JP7.2 symlinks reused a shared carrier whose patch context
does not match those baselines:

JP7.1 and JP7.2 already define TEGRA_IMAGE_DT_RAW16.
JP7.2 conditionally wraps the legacy vb2 wait callbacks for Linux 7.0, so
the channel.c queue layout differs.
These were patch-application failures, not IR data-path failures.

Fix
Use opaque UD30/T_R8 as the default byte-transparent Y12I carrier exposed
through the standard Y16I ABI.
Retain RAW16/T_R16 as an explicit diagnostic fallback and restore its slot
byte order in a D58x-only frame-finish callback.
Prefer the active subdevice's advertised transport when multiple VI tuples
share Y16I.
Register frame post-processing only for detected D58x devices.
Replace the JP7.1 and JP7.2 symlinks with independent carriers aligned to
jetson_38.4 and jetson_39.2.0.
RGB calibration
Problem and root cause
D58x RGB calibration firmware can emit unpacked GRBG10 and native
cache-line-packed GRBG10P, but the unified driver exposed only YUYV/NV12.
The D58x RGB table had no calibration profiles or transport tuples, Tegra VI
lacked the required RAW16/UD30 byte surfaces, and the shared mux enumerator
remapped format indices after the first three entries through the legacy
IR/depth table.

Fix
Add D58x-only 1600x1300@25 BA10, pgAA, and GR0P formats.
Receive BA10 and pgAA over RAW16/T_R16, restore the container byte order,
and pack pgAA rows after VI completion.
Receive GR0P over UD30/T_R8 while preserving its 64-byte cache-line ABI.
Select the complete per-stream format table before applying the legacy
shared IR/depth mux fallback.
Add compatible OOT carriers for every JP6/JP7 baseline.
D4x compatibility
All new stream formats are selected only after
DS5_DEVICE_TYPE_D58X detection.
D4x format tables, media-bus codes, CSI data types, and stream paths are
unchanged.
D58x metadata and frame post-processing operations are registered only for
D58x. D4x keeps the existing Tegra metadata fallback and has no frame
post-processing callback.
Duplicate-FOURCC lookup prefers an actively advertised transport but keeps
the previous first-match result as its fallback.
This PR changes no DTS. D58x validation uses
tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dtbo.
CI covers all supported JetPack build baselines. The hardware E2E run used
D58x; a physical D4x hardware run is not claimed.
Validation
Per-feature checkpoints
Metadata: clean JP6.2.2 full kernel/OOT build, deployment, reboot, baseline
Depth/IR/RGB/IMU, node/control/SERDES, retained-SERDES reload, and
multi-stream E2E passed.
NV12: complete ten-version CI passed. NV12 1280x960@60 captured 180/180
frames before and after three retained-SERDES reload cycles, with exact
sequence, a 1,843,200-byte standard NV12 buffer, zero new VI capture
errors, and a successful 180-frame YUYV restore. The accumulated
Depth/IR/RGB/IMU and multi-stream E2E passed.
IR calibration: complete ten-version CI passed, including the corrected
JP7.1 and JP7.2 carriers. Default UD30 Y16I captured 180/180 frames at
1600x1300@15 twice, including after three retained-SERDES reload cycles,
with exact sequence and zero correctable/uncorrectable error delta. The
accumulated Depth/IR/RGB/NV12/IMU and multi-stream E2E passed.
RGB calibration: every carrier applied to its fully patched NVIDIA
baseline. BA10, pgAA, and GR0P each captured 180/180 frames at
1600x1300@25 with exact sequence and zero kernel capture-error delta.
Payload checks decoded non-zero 10-bit Bayer samples; all GR0P cache-line
padding remained zero.
Final rebased head on Jetson 67
Rebased directly onto current dev (d7a77c6); current head is
558159b.
The exact rebased head passed the complete GitHub Actions matrix:
10/10 PASS for JP5.0.2, JP5.1.2, JP5.1.6, JP6.0, JP6.1, JP6.2,
JP6.2.2, JP7.0, JP7.1, and JP7.2. JP7.1 and JP7.2 both completed patch
application and the full carrier build successfully.
Reused the existing sources_6.2.2 tree; no kernel sources were downloaded.
Clean full JP6.2.2 kernel/OOT build passed on the final functional tree.
Exact rebased HEAD modules-only rebuild, deployment, and reboot passed
artifact SHA256, vermagic, module loading, boot, D5x FG24 tunnel overlay,
and MAX96724 A/B link-lock checks.
Final 180-frame feature matrix: PASS for NV12, default UD30 Y16I, BA10,
pgAA, GR0P, YUYV restore, Depth Z16, IR Y8I restore, and IMU. Every case
returned exact sequence 0..179; kernel capture-error delta was zero.
Applicable full E2E: 20/20 PASS in 97.95 seconds:
8/8 Depth+RGB concurrency profiles
8/8 Depth+RGB+IR concurrency profiles
stream start/stop
metadata arrival, parsing, and hardware timestamp progression
Post-test Jetson kernel log contained zero capture-error matches. The D5x
serial test window contained no ASSERT, PANIC, FATAL, HardFault, or
watchdog event.
Known non-default diagnostic behavior
The default and release-gated IR calibration transport is UD30 and passes
without capture errors. The explicit RAW16 diagnostic fallback produces
continuous monotonic Y16I buffers on packet DT 0x2e, but Tegra marks every
buffer with the known correctable err_data 16384 flag. The final target and
all final regressions were restored to the default UD30 path.

The generic repository V4L2 package also contains D4x-only control tests,
firmware-policy assumptions, an ARM64 v4l2_format layout issue in its test
harness, per-frame 5% timing assertions, and legacy metadata field offsets.
Those generic-test limitations are not used to waive a driver-path failure;
the D58x applicable full E2E and the dedicated long-run feature matrix above
are the hardware gates for this PR.


ymodlin (commented):
Automated review — no blocking findings in the areas checked, so no inline comments.

Covered: locking around the shared slot globals (`ds5_inited[]`/`dser_inited[]`/`serdes_lock__`, `ds5_primary`, streaming flags), the GMSL one-shot/line-buffer-flush invariant (no new flush call sites, nothing added inside `set_pipe` — good), pipe alloc/release symmetry on the error paths, memory safety in the new stream/format plumbing, and the rule that camera-model-dependent behaviour is decided from the runtime device type rather than a DT property (`ds5_is_d58x()` gating looks right).

One note for other reviewers so it does not get "fixed": in `ds5_pack_raw16_to_sgrbg10p()` the destination row start advances by the RAW16 stride (`width * 2`) while each row only writes `width * 5/4` packed bytes. That is intentional and correct for `MEDIA_BUS_FMT_RS_SGRBG10P_RAW16_1X16` — the packed rows are carried in a RAW16-strided surface, and keeping src/dst strides equal is what makes the in-place pass over the scratch copy safe.

kernel/realsense/d4xx.c
```diff
@@ -1964,6 +2236,124 @@ static const struct ds5_format *ds5_sensor_find_format(
...
+		    MEDIA_BUS_FMT_RS_Y12I_UD30_4X8 &&
+	     sensor->config.format->mbus_code !=
+		    MEDIA_BUS_FMT_RS_Y12I_RAW16_4X8))
+		return;
+
+	/*
+	 * Keep Y16I logical geometry while VI byte-DMAs the complete
+	 * width * 4-byte row without interpreting the 12-bit samples.
+	 */
+	mode = &state->mux.sd.sensor_props.sensor_modes[1];
```
ymodlin, line 2334: **BUG (out-of-bounds write)** — `sensor_modes[1]` is indexed without checking `sensor_props.num_modes`.

`tegra234-camera-d4xx-overlay-d5xx.dts` (and `-fg12-16ch-d5xx`, `-seeed-d5xx`) define **only `mode0`** per camera node, so `num_modes == 1` and this writes past the end of the devm-allocated mode array — kernel heap corruption. `state->mux.sd.mode_prop_idx = 1` then points VI at a nonexistent mode.

Guard it, the way NVIDIA's own code does (`0001-embedded-metadata-support-for-D4xx.patch`: `if (idx < s_data->sensor_props.num_modes)`):

```c
if (state->mux.sd.sensor_props.num_modes < 2)
	return;
```

Either that, or add a `mode1` to every D5xx overlay — but the driver should not trust DT here.

kernel/realsense/d4xx.c
```diff
@@ -1964,6 +2236,124 @@ static const struct ds5_format *ds5_sensor_find_format(
...
+		break;
+	case MEDIA_BUS_FMT_RS_SGRBG10P_1X10:
+		pixel_format = V4L2_PIX_FMT_RS_SGRBG10P;
+		bit_depth = 10;
+		break;
+	default:
+		return;
+	}
+
+	mode = &state->mux.sd.sensor_props.sensor_modes[0];
```
ymodlin, line 2299: **BUG** — same missing bound check as the Y12I path: the `sensor_modes` pointer is NULL-checked but `num_modes` is never validated before indexing `[0]`. Add `if (!state->mux.sd.sensor_props.num_modes) return;`.

kernel/realsense/d4xx.c
```diff
@@ -618,6 +684,148 @@ struct ds5 {
...
+		return 0;
+	if (!data || !bytesused || WARN_ON_ONCE(bytesused & 1U))
+		return -EFAULT;
+
+	mutex_lock(&state->frame_postprocess_lock);
+	ret = ds5_prepare_frame_postprocess_scratch(state, bytesused);
+	if (ret)
+		goto unlock;
+
+	memcpy(state->frame_postprocess_scratch, data, bytesused);
```
ymodlin, line 798: **Enhancement (per-frame cost)** — the scratch buffer, the `kvmalloc`, the mutex and this full-frame `memcpy` are all avoidable: **both transforms are safe in place.**

- `ds5_swab16_frame()` — each 16-bit word is independent.
- `ds5_pack_raw16_to_sgrbg10p()` — iteration `k` reads `[8k, 8k+8)` *before* writing `[5k, 5k+5)`, and `5k+5 <= 8k+8` always holds, so the write pointer never overruns unread source.

At 1600x1300x2 that is a ~4.2 MB copy per frame (~104 MB/s at 25 fps) for no benefit. Dropping the scratch also removes the `-ENOMEM` path and the `GFP_KERNEL` allocation from the capture callback.

kernel/realsense/d4xx.c
```diff
@@ -2173,8 +2573,10 @@ static int ds5_configure(struct ds5 *state)
 
 #ifdef CONFIG_VIDEO_D4XX_SERDES
 	data_type1 = sensor->config.format->data_type;
+	if (y12i_raw16_carrier && state->is_y8 &&
+	    data_type1 == GMSL_CSI_DT_RAW_16)
+		md_fmt = 0x00;
```
ymodlin, line 2578: **Question / possible BUG** — this silently forces `md_fmt = 0x00`, which feeds both `data_type2` (SerDes pipe) **and** the `md_addr` FW register write below. So enabling the `y12i_raw16_carrier` module param disables metadata for the IR stream without `metadata_enabled` reflecting it, and userspace gets no indication.

If the intent is only "do not allocate a metadata pipe for RAW16", scope it to `data_type2` and leave the FW `md_value` alone. Otherwise please log it once.

kernel/realsense/d4xx.c
```diff
@@ -1595,9 +1803,9 @@ static const struct ds5_resolution d58x_y8_sizes[] = {
 	DS5_RES(424, 240, ds5_framerate_to_90)
 };
 
-static const struct ds5_resolution d58x_calibration_sizes[] = {
+/* Native Y12I calibration: one 16-bit L and R container per logical pixel. */
+static const struct ds5_resolution d58x_y12i_sizes[] = {
```
ymodlin, line 1807: **Possible regression** — `d58x_calibration_sizes` carried **two** resolutions (1600x1300 and 256x144@90); `d58x_y12i_sizes` keeps only 1600x1300. Combined with the removal of the `GMSL_CSI_DT_RGB_888` entry from `ds5_depth_formats_d58x`, the D58x **depth** stream loses its calibration format entirely and IR loses the 256x144 profile.

If librealsense enumerates either, this breaks it. Please confirm intentional and call it out in the PR description.

kernel/realsense/d4xx.c
```diff
@@ -1964,6 +2236,124 @@ static const struct ds5_format *ds5_sensor_find_format(
 /* 1-8 */
 #define MIPI_CSI2_TYPE_USER_DEF(i)	(0x30 + (i) - 1)
 
+#ifdef CONFIG_TEGRA_CAMERA_PLATFORM
+static void ds5_tegra_update_mipi_clock(struct sensor_signal_properties *signal,
```
ymodlin, line 2240: **Cross-generation build risk** — this block is gated on `CONFIG_TEGRA_CAMERA_PLATFORM` only, while the post-process block correctly also requires `LINUX_VERSION_CODE >= KERNEL_VERSION(5, 15, 0)`.

Per CLAUDE.md, `d4xx.c` is symlinked into **every** JetPack build tree and CI builds 5.0.2 -> 7.2. If JP5's `struct sensor_signal_properties` has no `serdes_pixel_clock` member, JP5 fails to compile here. Worth confirming against the 5.10 tree, or reusing the same version guard.

kernel/realsense/d4xx.c
```diff
@@ -116,6 +118,56 @@ struct ser_interface {
...
+#ifndef GMSL_CSI_DT_RAW_16
+#define GMSL_CSI_DT_RAW_16 0x2E
+#endif
+
+#ifndef GMSL_CSI_DT_USER_1
+#define GMSL_CSI_DT_USER_1 0x30
+#endif
+
+#ifndef MEDIA_BUS_FMT_RS_SGRBG10P_1X10
+#define MEDIA_BUS_FMT_RS_SGRBG10P_1X10 0x5002
```
ymodlin, line 130: **Enhancement (drift hazard)** — these codes are defined twice: here as `#ifndef` fallbacks and again in `0015`/`0016` (`0x5002`-`0x5006`, `Y16I`, `GR0P`). They agree today, but the `#ifndef` means a build with the nvidia-oot patches **not** applied still compiles and advertises formats VI cannot decode — a silent runtime failure instead of a build error.

Prefer a single source of truth (the patched header) and let the build break if it is missing.

nvidia-oot/6.1/0016-Add-D5x-RGB-Bayer-byte-surfaces.patch
```diff
@@ -0,0 +1,103 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
ymodlin, line 1: **CLAUDE.md guideline** — this file is byte-identical (md5 `f03bd114`) to `6.0/`, `6.2/`, `7.0/`, `7.1/` and `7.2/`. CLAUDE.md: *"`6.0/` holds the real files; `6.1/`/`6.2/` symlink the byte-identical ones back to `6.0/` and only keep a real file where content differs."*

`0015` in this PR already does it right (`6.1/0015 -> ../6.0/0015`). Please make the five duplicate `0016` copies symlinks to `6.0/` so a future edit cannot land on one carrier only.

CLAUDE.md
```diff
@@ -83,6 +83,10 @@ RealSense D4XX camera module
 ### Key directories
 
 - **`kernel/realsense/d4xx.c`** — The main driver. Single-file V4L2 subdevice driver handling I2C communication, MIPI CSI-2 stream config, firmware control (DFU), calibration data, metadata capture, and V4L2 controls (exposure, laser power, AE ROI, etc.). Registers four sensor subdevices per camera: Depth, RGB, IR (Y8/Y8I/Y12I), and IMU.
+- **D58x CSI metadata framing** — Register Tegra embedded-metadata ops only for a detected D58x. New firmware uses the existing 12-byte USB/UVC metadata header over CSI, so publish runtime `bytesused` from its validated length while retaining legacy HKRM parsing. Expose `D4XX` as the native FOURCC with `RSMD` compatibility; the fixed 4096-byte value is queue capacity, not actual payload size. D4x and sensors without ops keep the legacy Tegra fallback.
```
ymodlin, line 86: **CLAUDE.md guideline (comment style + placement)** — these four bullets are multi-sentence paragraphs under **`### Key directories`**, but they describe features, not directories.

CLAUDE.md's own rule: *"Keep comments concise and high-signal... a few skimmable lines or bullets, not walls of text."* Suggest a separate `### D58x stream formats` subsection with 1-2 lines each, keeping the non-obvious invariants (UD30 default vs RAW16 diagnostic, NV12-as-RAW8 geometry) and dropping the restated context.

ymodlin (commented):
**[rs-agentic] automated review** — internal contribution, so this covers BUGs, enhancements and CLAUDE.md guideline compliance. Draft PR; no approval implied.

Solid, well-separated port. One **blocker** below (out-of-bounds write on the shipped D5xx overlay), plus a few carrier/hygiene items.

**Must fix**
- `sensor_modes[1]` / `sensor_modes[0]` are indexed with no `num_modes` bound check. `tegra234-camera-d4xx-overlay-d5xx.dts` defines only `mode0` per camera node (`num_modes == 1`), so the Y12I path writes past the devm-allocated mode array and sets `mode_prop_idx = 1` to a mode that does not exist. NVIDIA's own code guards this (`if (idx < s_data->sensor_props.num_modes)` in `0001-embedded-metadata-support-for-D4xx.patch`).

**Repo-level items (not inline-able)**
- **Stale CLAUDE.md claim.** This PR deletes both `is_calib` `reset_oneshot()` call sites, but CLAUDE.md still documents them: *"The coarse `reset_oneshot(0x0F)` (all-links) on the Y12I/`is_calib` path is a separate, ungated reset, out of scope here (RSDEV-13028)."* Post-patch instruction hygiene requires correcting that sentence in the same change. (Removing the all-links reset looks right per RSDEV-13028 — just document it.)
- **Patch-carrier duplication.** `0016-Add-D5x-RGB-Bayer-byte-surfaces.patch` is byte-identical (md5 `f03bd114`) in all six of `6.0/ 6.1/ 6.2/ 7.0/ 7.1/ 7.2/` as six real files. CLAUDE.md: *"`6.0/` holds the real files; `6.1/`/`6.2/` symlink the byte-identical ones back to `6.0/` and only keep a real file where content differs."* `0015` does this correctly — please mirror it for `0016`. (`0013`/`0014` genuinely differ per version, so those staying real is fine.)
- **Branch conflicts.** GitHub reports `mergeable: false` against `dev` — needs a rebase.
- Title typo: "addtional" -> "additional".

ymodlin:
**[rs-agentic] note** — the earlier automated review on this PR ("no blocking findings in the areas checked") was posted by a concurrent run of this bot and is **superseded** by [the review below](https://github.com/realsenseai/realsense_mipi_platform_driver/pull/549#pullrequestreview-4916349643), which did find a blocker: the unguarded `sensor_modes[1]` write in `ds5_tegra_update_y12i_mode()` is out-of-bounds on the shipped D5xx overlays (`num_modes == 1`). Please treat the later review as authoritative.

PR #564, opened 2026-08-07
[RSDEV-13621][D58x] Fix GMSL DFU completion and re-enumeration

lemon6630:
## Tracking

- JIRA: [RSDEV-13621](https://rsjira.realsenseai.com/browse/RSDEV-13621)

## Summary

This PR fixes both phases required for D58x GMSL DFU to complete without a manual `d4xx` module reload:

- wait for the D58x `dfuMANIFEST_WAIT_RESET` terminal state and correctly finalize aligned images
- preserve successful manifestation state across character-device close, then recover the D585 GMSL link only after HKR finishes its post-DFU PWDNB initialization

## Root cause

- D58x transitions `dfuMANIFEST_SYNC -> dfuMANIFEST -> dfuMANIFEST_WAIT_RESET`; treating state 0x07 as terminal reports failure before firmware announces its reboot.
- Finalization previously lived only in the short-final-block path, so a 1024-byte-aligned image could skip download-complete and manifestation handling.
- After manifestation, the existing hardware-reset recovery restored the serializer alias too early. HKR performs another GMSL/PWDNB initialization roughly 1.5 seconds later and resets the alias again, leaving the camera absent until `d4xx` is reloaded.
- During this reboot window, the DFU magic value can be observed transiently and must not be treated as the final recovery state.

## Changes

- Select the expected manifestation terminal state per device family.
- Poll D58x through states 0x06/0x07 to 0x08 with a bounded 180-second deadline while preserving D4xx behavior at state 0x07.
- Share finalization between the short-block path and `.flush`, including close-time error propagation.
- Retain a successful manifestation marker across DFU character-device close.
- For D585 post-DFU reset, wait for HKR GMSL/PWDNB initialization before restoring the serializer alias.
- Ignore transient DFU magic during that manifestation reset and clear the marker only after full device/link recovery.

The Recovery device-family hint remains session-local and does not modify `cached_device_type`.

## Validation

- `git diff --check`: passed.
- `./build_all.sh 6.2`: passed for Jetson kernel `5.15.148-tegra`.
- Deployed the resulting `d4xx.ko` on Jetson `192.168.112.132`.
- Completed compressed-image DFU on D585 serial `333622320064` in both directions:
  - `7.58.45946.14332 -> 7.58.46031.14359`
  - `7.58.46031.14359 -> 7.58.45946.14332`
- Both `rs-fw-update` operations returned success and independently enumerated the exact target firmware version.
- No `rmmod`, `modprobe`, or system reboot was used between updates.
- After the second DFU, opened `640x480@30 Z16` depth streaming and received 61 frames before a clean stop/close.

The corresponding librealsense lifecycle changes are tracked in [librealsense PR #15619](https://github.com/realsenseai/librealsense/pull/15619).

## Related PRs

- [PR #595](https://github.com/realsenseai/realsense_mipi_platform_driver/pull/595) proposes retries for transient GETSTATUS reads; that is orthogonal to terminal-state, aligned-image finalization, and post-DFU GMSL recovery.

## Companion firmware/test PRs

- IO: https://github.com/RealSenseAI-Internal/soc-rs-soc-io-drivers/pull/520
- tests: https://github.com/RealSenseAI-Internal/soc-rs-hkr-test/pull/356
- build: https://github.com/RealSenseAI-Internal/soc-rs-hkr-trunk/pull/130
- FreeRTOS: https://github.com/RealSenseAI-Internal/soc-rs-soc-os-freertos/pull/50
- device manager: https://github.com/RealSenseAI-Internal/soc-rs-soc-device-mgr/pull/452
- apps: https://github.com/RealSenseAI-Internal/soc-rs-hkr-apps/pull/271
- prebuilts: https://github.com/RealSenseAI-Internal/soc-rs-soc-prebuilts/pull/85

### DIFF
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -145,6 +145,12 @@ struct ser_interface {
 #define DS5_DEVICE_TYPE_D45X		6
 #define DS5_DEVICE_TYPE_D43X		5
 #define DS5_DEVICE_TYPE_UNKNOWN		0
+/*
+ * FW version major byte identifies the family in recovery, where DEVICE_TYPE
+ * (0x0310) is not served: D58x reports 7 or 8, D4xx reports 5.
+ */
+#define DS5_D58X_FW_MAJOR_MIN		7
+#define DS5_D58X_FW_MAJOR_MAX		8
 
 /* GVD response payload size per product line */
 #define DS5_GVD_LEN_D4XX		276
@@ -297,6 +303,13 @@ enum ds5_mux_pad {
 #define PIPE_NOT_CONFIGURED	-1
 
 #define DFU_WAIT_RET_LEN 6
+
+/*
+ * Deadline for the D58x manifest wait; the manifest can legitimately take
+ * up to ~90 s, so the bound is generous and only guards against a device
+ * stuck in an intermediate manifest state.
+ */
+#define DFU_MANIFEST_TIMEOUT_MS 180000
 
 #define DS5_START_POLL_TIME	10
 #define DS5_START_MAX_TIME	2000
@@ -583,6 +596,8 @@ struct ds5_dfu_dev {
 	struct class *ds5_class;
 	int device_open_count;
 	enum dfu_state dfu_state_flag;
+	enum dfu_fw_state manifest_state;
+	bool manifest_complete;
 	unsigned char *dfu_msg;
 	u16 msg_write_once;
 	// unsigned char init_v4l_f; // need refactoring
@@ -3114,6 +3129,14 @@ static int ds5_set_calibration_data(struct ds5 *state,
 #define DS5_HW_RESET_TIMEOUT_MS		10000
 #define DS5_HW_RESET_MAX_RETRIES	(DS5_HW_RESET_TIMEOUT_MS / DS5_HW_RESET_POLL_INTERVAL_MS)
 
+/*
+ * D585 reinitializes its serializer from HKR roughly 1.5 seconds after a
+ * post-DFU reboot.  Recovering the serializer alias before that point is
+ * ineffective because the later HKR PWDNB cycle resets it back to the
+ * default address.
+ */
+#define DS5_D585_DFU_SERDES_SETTLE_MS	1500
+
 /* Minimum interval between consecutive HW resets (ms).
  * Rapid back-to-back resets degrade the GMSL link.
  */
@@ -3244,6 +3267,8 @@ static int ds5_hw_reset_with_recovery(struct ds5 *state)
 	u16 d585_product_id = READ_ONCE(state->ds5_dev->d585_product_id);
 	bool d585_proto_reset = d585_product_id == D585_2C_PROTO_PID ||
 				 d585_product_id == D585_3C_PROTO_PID;
+	bool post_dfu_reset = d585_proto_reset &&
+		READ_ONCE(state->dfu_dev.manifest_complete);
 	unsigned long ds5_last_reset_jiffies = READ_ONCE(state->ds5_dev->last_reset_jiffies);
 	unsigned long ts, timeout;
 
@@ -3270,6 +3295,42 @@ static int ds5_hw_reset_with_recovery(struct ds5 *state)
 			msleep(jiffies_to_msecs(remaining));
 		}
 	}
+
+#ifdef CONFIG_VIDEO_D4XX_SERDES
+	/*
+	 * D585 prototype-only workaround for the serializer/control-path recovery
+	 * issue post-DFU manifestation. Restore the serializer alias and control
+	 * tunnel before the first camera I2C access after the SDK-triggered HW reset.
+	 */
+	if (post_dfu_reset) {
+		struct ds5 *primary = state->ds5_dev->ds5_primary;
+
+		if (!primary || !primary->dser_ops->recover_link)
+			return -ENODEV;
+
+		mutex_lock(&serdes_lock__);
+		ret = primary->ser_ops->reset_control(primary->ser_dev);
+		if (!ret)
+			ret = primary->dser_ops->recover_link(primary->dser_dev,
+						      primary->ser_dev,
+						      primary->gmsl_link);
+		if (!ret)
+			ret = primary->ser_ops->setup_control(primary->ser_dev);
+		if (!ret)
+			ret = primary->ser_ops->init_settings(primary->ser_dev);
+		mutex_unlock(&serdes_lock__);
+		if (ret) {
+			dev_err(&state->client->dev,
+				"%s(): pre-HW-reset D585 SerDes recovery failed (%d)\n",
+				__func__, ret);
+			return ret;
+		}
+
+		dev_info(&state->client->dev,
+			 "%s(): pre-HW-reset D585 GMSL control path recovered\n",
+			 __func__);
+	}
+#endif
 
 	/* 1. Stop active streams on the device before reset.
 	 *    This ensures FW and SERDES are in a clean state.
@@ -3355,6 +3416,18 @@ static int ds5_hw_reset_with_recovery(struct ds5 *state)
 		if (!primary || !primary->dser_ops->recover_link)
 			return -ENODEV;
 
+		/*
+		 * The D585 manifestation reset is followed by HKR's own GMSL PWDNB
+		 * cycle.  Let that finish before restoring the serializer alias;
+		 * otherwise the alias is lost again just after this function returns.
+		 */
+		if (post_dfu_reset) {
+			dev_info(&state->client->dev,
+				 "%s(): waiting for D585 post-DFU GMSL initialization\n",
+				 __func__);
+			msleep(DS5_D585_DFU_SERDES_SETTLE_MS);
+		}
+
 		mutex_lock(&serdes_lock__);
 		ret = primary->ser_ops->reset_control(primary->ser_dev);
 		if (!ret)
@@ -3402,6 +3475,12 @@ static int ds5_hw_reset_with_recovery(struct ds5 *state)
 
 		ret = ds5_read_poll(state, DS5_DFU_MAGIC_REG, &ready_status);
 		if (!ret && ready_status == DS5_DFU_MAGIC_LSW) {
+			if (post_dfu_reset) {
+				dev_dbg(&state->client->dev,
+					"%s(): transient DFU state during manifestation reset\n",
+					__func__);
+				continue;
+			}
 			dev_warn(&state->client->dev,
 				"%s(): Device in DFU/recovery mode after reset\n", __func__);
 			state->dfu_dev.dfu_state_flag = DS5_DFU_RECOVERY;
@@ -3492,6 +3571,8 @@ static int ds5_hw_reset_with_recovery(struct ds5 *state)
 	}
 
 	WRITE_ONCE(state->ds5_dev->last_reset_jiffies, jiffies);
+	if (post_dfu_reset)
+		WRITE_ONCE(state->dfu_dev.manifest_complete, false);
 
 	return 0;
 }
@@ -7388,6 +7469,8 @@ static int ds5_dfu_wait_for_get_dfu_status(struct ds5 *state,
 	u16 status, dfu_state_len = 0x0000;
 	unsigned char dfu_asw_buf[DFU_WAIT_RET_LEN];
 	unsigned int dfu_wr_wait_msec = 0;
+	unsigned long manifest_deadline = jiffies +
+			msecs_to_jiffies(DFU_MANIFEST_TIMEOUT_MS);
 
 	do {
 		ds5_write_with_check(state, 0x5008, 0x0003); // Get Write state
@@ -7419,7 +7502,18 @@ static int ds5_dfu_wait_for_get_dfu_status(struct ds5 *state,
 		dfu_wr_wait_msec = (((unsigned int)dfu_asw_buf[3]) << 16)
 						| (((unsigned int)dfu_asw_buf[2]) << 8)
 						| dfu_asw_buf[1];
-	} while (dfu_asw_buf[4] == dfuDNBUSY && exp_state == dfuDNLOAD_IDLE);
+	/*
+	 * D58x manifestation runs dfuMANIFEST_SYNC -> dfuMANIFEST ->
+	 * dfuMANIFEST_WAIT_RESET; keep polling through the intermediate
+	 * states until the terminal wait-reset state is reached, bounded by
+	 * manifest_deadline so a stuck device errors out instead of hanging.
+	 */
+	} while ((dfu_asw_buf[4] == dfuDNBUSY &&
+		  exp_state == dfuDNLOAD_IDLE) ||
+		 ((dfu_asw_buf[4] == dfuMANIFEST_SYNC ||
+		   dfu_asw_buf[4] == dfuMANIFEST) &&
+		  exp_state == dfuMANIFEST_WAIT_RESET &&
+		  time_before(jiffies, manifest_deadline)));
 
 	if (dfu_asw_buf[4] != exp_state) {
 		dev_notice(&state->client->dev,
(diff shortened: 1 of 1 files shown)


kernel/realsense/d4xx.c
```diff
@@ -6653,7 +6657,14 @@ static int ds5_dfu_wait_for_get_dfu_status(struct ds5 *state,
 		dfu_wr_wait_msec = (((unsigned int)dfu_asw_buf[3]) << 16)
 						| (((unsigned int)dfu_asw_buf[2]) << 8)
 						| dfu_asw_buf[1];
-	} while (dfu_asw_buf[4] == dfuDNBUSY && exp_state == dfuDNLOAD_IDLE);
+	/* D58x manifestation runs dfuMANIFEST_SYNC -> dfuMANIFEST ->
+	 * dfuMANIFEST_WAIT_RESET; keep polling through the intermediate
+	 * states until the terminal wait-reset state is reached. */
+	} while ((dfu_asw_buf[4] == dfuDNBUSY &&
```
ymodlin, line 6724: **Potential BUG:** The new `do..while` branch for `dfuMANIFEST_SYNC`/`dfuMANIFEST` states has no explicit timeout. If the device gets stuck in one of the intermediate manifest states and never transitions to `dfuMANIFEST_WAIT_RESET`, this loop will never exit. The existing `dfuDNBUSY` branch relies on `dfu_wr_wait_msec` delays, but there's no retry counter or deadline for the new manifest states. Consider adding a maximum iteration guard (e.g., reuse or extend the existing timeout mechanism in this function).

> 🤖 *Auto-generated by rs-agentic bot*

ymodlin (commented):
Automated review for PR #564: [D58x] DFU: wait for dfuMANIFEST_WAIT_RESET terminal state

kernel/realsense/d4xx.c
lemon6630: Good catch — addressed in 05ccddb. The manifest poll is now bounded by a jiffies deadline (DFU_MANIFEST_TIMEOUT_MS = 180 s, generous since a legitimate manifest can take ~90 s). On timeout the state stays != dfuMANIFEST_WAIT_RESET, so the existing 'Wrong dfu_state' -EINVAL return fires instead of spinning. Only the new manifest clause is bounded; the proven D4xx dfuDNBUSY path is untouched.

kernel/realsense/d4xx.c
```diff
@@ -6622,6 +6631,8 @@ static int ds5_dfu_wait_for_get_dfu_status(struct ds5 *state,
 	u16 status, dfu_state_len = 0x0000;
 	unsigned char dfu_asw_buf[DFU_WAIT_RET_LEN];
 	unsigned int dfu_wr_wait_msec = 0;
+	unsigned long manifest_deadline = jiffies +
```
ashrafk93, line 6687: **[rs-agentic]** ℹ️ `manifest_deadline` is computed once at function entry, shared across both the `dfuDNBUSY` phase and the new MANIFEST polling phase. If the DN-BUSY phase itself consumes most of the 180 s (e.g. a slow device or many retries), the deadline may already be expired when the MANIFEST polling condition first activates, causing an immediate false timeout. Consider initializing the deadline only when entering the manifest wait branch, or use a separate larger deadline for the manifest phase.

ashrafk93 (commented):
**[rs-agentic auto-review]** PR #564 reviewed.

ymodlin (dismissed):
1. No ticket number
2. This PR effect both d4xx and d585 should be consulted before sending a PR I want to understand the design and the requirments.

lemon6630:
@ymodlin Thanks for raising this. I agree that, because this is the unified D4xx/D5xx driver, I should have aligned on the design and requirements with both owners before requesting review.

The PR is now linked to [RSDEV-13621](https://rsjira.realsenseai.com/browse/RSDEV-13621). The requirement is: a D585 GMSL DFU that starts from Recovery must complete through the HKR reboot and restore the four logical sensors and V4L2 nodes without rebooting Jetson or manually reloading the Host driver, while preserving the existing D4xx DFU behavior.

The intended compatibility boundaries are:

- D4xx still treats `dfuMANIFEST` (0x07) as terminal; only a confirmed D58x waits for `dfuMANIFEST_WAIT_RESET` (0x08).
- The FW-major fallback is used only when cold Recovery cannot provide a valid device type, and major >= 7 identifies D58x; it never overrides a valid D4xx device type.
- Automatic re-probe is scheduled only when the DFU session began in Recovery, completed successfully, and `ds5_is_d58x()` is true. Operational DFU and all D4xx sessions do not schedule it.
- The four-client tracking and module-exit cleanup are shared infrastructure because all aliases use the same unified driver. Although their runtime action is D58x-gated, this shared lifecycle code still needs D4xx-owner review and D4xx regression coverage before merge.

The D585 path has been validated end to end on JP6.2.1/MAX96724, including HKR reboot, automatic primary-first re-probe, seven V4L2 nodes, SDK enumeration, and no Oops/WARN. I will not merge this until we have aligned on the design and the required D4xx regression scope. I am happy to walk through the state machine and re-probe lifecycle together.

kernel/realsense/d4xx.c
ymodlin: change approved

kernel/realsense/d4xx.c
ymodlin: Re-checked this one — it looks like a false positive, no change needed.

`manifest_deadline` is initialised on every call to `ds5_dfu_wait_for_get_dfu_status()`, and the two `while` clauses are mutually exclusive via `exp_state`: the DN-BUSY clause requires `exp_state == dfuDNLOAD_IDLE`, the manifest clause requires `exp_state == dfuMANIFEST_WAIT_RESET`. The DN-BUSY phase runs in a *separate* invocation (called with `dfuDNLOAD_IDLE`), so it cannot consume the manifest call's deadline — the 180 s clock starts fresh when the manifest wait begins.

kernel/realsense/d4xx.c
```diff
@@ -7015,24 +7110,169 @@ static int ds5_v4l_init(struct i2c_client *c, struct ds5 *state)
...
 #endif
+	bool auto_reprobe = false;
 	int ret = 0, retry = 10;
 	mutex_lock(&state->lock);
 	state->dfu_dev.device_open_count--;
+#ifdef CONFIG_VIDEO_D4XX_SERDES
+	auto_reprobe = state->dfu_dev.dfu_state_flag == DS5_DFU_DONE &&
+		state->dfu_dev.reprobe_after_dfu && ds5_is_d58x(state);
+	if (auto_reprobe)
+		ds5_dfu_schedule_reprobe(state);
```
ymodlin: **Potential BUG:** the return value of `ds5_dfu_schedule_reprobe()` is discarded, but `auto_reprobe` stays `true` regardless. At line ~7294 that causes an early `return 0` which skips the FW-version verification loop.

So when scheduling fails (`-ENOSPC` when all `ds5_dfu_reprobe[]` slots are busy, or `-ENODEV` when `client_count != 4`), the device is neither re-probed nor verified — it is silently left unusable after a successful DFU.

Suggest `auto_reprobe = ds5_dfu_schedule_reprobe(state) == 0;` so the fallback verification path still runs.

kernel/realsense/d4xx.c
```diff
@@ -6713,6 +6733,16 @@ static int ds5_dfu_detach(struct ds5 *state)
...
 			__func__, buf.DFU_isLocked ? "locked" : "unlocked");
+	/* Recovery never ran ds5_wait_device_type(), so cached_device_type is
+	 * UNKNOWN; recover the family from the FW major byte so the DFU
+	 * manifest wait picks the D58x terminal state (dfuMANIFEST_WAIT_RESET).
+	 */
+	if (!ret && state->ds5_dev &&
+	    !ds5_is_valid_device_type(
+		    READ_ONCE(state->ds5_dev->cached_device_type)) &&
+	    ((buf.FW_lastVersion >> 24) & 0xff) >= DS5_D58X_FW_MAJOR_MIN)
+		WRITE_ONCE(state->ds5_dev->cached_device_type,
```
ymodlin: **BUG** — writing a *guessed* family into `cached_device_type` poisons the FW-confirmed device type.

That field is consumed by `ds5_is_d58x()` → `ds5_rgb_ctrl_offset()` (also ~L4978/L5200), and by the `ds5_wait_device_type()` fast path (~L2765) which returns success as soon as the cached value is valid — without ever reading `DS5_DEVICE_TYPE`. So a wrong guess silently routes RGB control writes to D58x register offsets, and a readiness gate can pass unconfirmed. CLAUDE.md: *"Do not let pre-reset cache values short-circuit post-reset readiness checks."*

The `>= 7` FW-major heuristic is also fragile if D4xx FW majors ever reach 7.

Suggest keeping the recovery hint in a separate field (e.g. `dfu_family_hint`) consumed **only** by the DFU manifest wait, and reserving `cached_device_type` for FW-read values.

kernel/realsense/d4xx.c
ymodlin: Withdrawing this one — `ds5_dfu_schedule_reprobe()` is not part of this PR (no reprobe code in the current diff); it belongs to #574 and is tracked there. Please ignore.

kernel/realsense/d4xx.c
```diff
@@ -7102,7 +7140,7 @@ static ssize_t ds5_dfu_device_write(struct file *flip,
 			if (!ret)
 				ret = ds5_write(state, 0x4a04, 0x00); /*Download complete */
 			if (!ret)
-				ret = ds5_dfu_wait_for_get_dfu_status(state, dfuMANIFEST);
+				ret = ds5_dfu_wait_for_get_dfu_status(state, manifest_state);
```
Nir-Az: Not introduced here, but this call sits inside `if (dfu_part_blocks)`, so the whole finalization — `0x4a04` download-complete, the manifest wait, and `DS5_DFU_DONE` — is skipped whenever a `write()` length happens to be an exact multiple of `DFU_BLOCK_SIZE` (1024). For an image streamed in one pass that makes completion depend on `image_size % 1024 != 0`; a 1024-aligned signed image would never finalize, and the caller still gets a successful write back with no error anywhere. Since this PR is already reworking the terminal-state logic on exactly this call, it seems like the right place to move the download-complete plus manifest wait out of the partial-block branch.

Nir-Az:
**AR — drivers team.** After #564/#595/#606 land, this is what's still needed for D5xx GMSL DFU to behave like D4xx GMSL (no workarounds, no manual driver reload):

1. Who issues `0x20` after DFU? D4xx FW self-resets; D58x parks at `dfuMANIFEST_WAIT_RESET`. If the driver issued it on `DS5_DFU_DONE`, the SDK would need no D5xx-specific path at all. This one decision drives everything else.
2. #606 covers only proto PIDs `0x0C07`/`0x0C08` — final `0x0C04`/`0x0C05`/`0x0C06` still have no post-reset link recovery.
3. `0x4a04` + manifest wait + `DS5_DFU_DONE` sit inside `if (dfu_part_blocks)`, so a `DFU_BLOCK_SIZE`-aligned image never finalizes and `write()` still returns success (see inline comment above).
4. Once FW answers `0x0310` in recovery, drop #564's FW-major guess. Also: one documented DFU chardev naming scheme plus the family exposed in sysfs, so the SDK can stop parsing the device-tree `compatible`.
5. #595 is a hard dependency, not an optional companion — error polling stays enabled on D500, so that bus contention is our normal path rather than an edge case.

Nir-Az:
**AR — FW team (HKR / D58x).** Companion to the drivers AR above; these are the FW-side asymmetries that force both the driver and the SDK to special-case D5xx GMSL DFU:

1. Self-reset after manifest the way D4xx does at `0x07`, or make `0x20` the documented contract on every D58x SKU. Please pick one — it decides whether #564's per-family terminal state is permanent.
2. Support `0x20` on final PIDs, not just the GMSL proto. Its absence is why the SDK's post-DFU reset is compiled out today.
3. Answer `DS5_DEVICE_TYPE` (0x0310) in recovery, or put the family in the DFU chardev's `DFU info:` line. Right now the driver guesses from the FW major byte and the SDK guesses from the device tree — two heuristics for one fact.
4. Stop returning `0xFFFF`/`0xFF` on `0x5004`/`0x4e00` during the manifest window, or document the retry contract (#595).
5. Report real `iString` DFU progress — without it the SDK falls back to a fixed 120 s ramp instead of actual progress.

kernel/realsense/d4xx.c
lemon6630: Fixed in 81ce083. Finalization now uses a shared ds5_dfu_finalize_download() helper. A short final block still finalizes in write(), while a 1024-byte-aligned image remains IN_PROGRESS until the file flush callback sends 0x4a04, waits for the session terminal manifest state, and sets DS5_DFU_DONE. The flush return value also propagates manifestation failure through ofstream close instead of reporting false success. I built the full JP6.2.1 module set with #606 and booted it on Jetson .172; active d4xx is 1.0.6.7 and the DFU/V4L2 nodes are present.

lemon6630:
Addressed the review points as follows:

1. Post-DFU `0x20` remains outside #564. For the current prototype integration, librealsense sends it only after the DFU file closes successfully; #606 owns the driver/SerDes recovery path.
2. #606 still supports only prototype PIDs `0x0C07`/`0x0C08`. The SDK runtime adapter is correspondingly gated in the local integration; final-PID support remains a separate requirement.
3. Fixed in `81ce083`: a 1024-byte-aligned image now finalizes from `.flush`, while a short final block still finalizes from `write()`. Both paths share the same helper, and close-time manifestation errors are returned to userspace.
4. The Recovery FW-major hint remains session-local and does not modify `cached_device_type`. Replacing the hint with `0x0310`/sysfs is pending the authoritative FW/ABI contract.
5. #595 remains an orthogonal reliability dependency; it was not silently folded into this terminal-state/aligned-image change.

Validation used a full JP6.2.1 module-set build containing #564 plus exact #606 head `4697b4e`. It is booted on Jetson `.172`; active `d4xx` is 1.0.6.7 and the DFU plus seven V4L2 nodes are present. I did not start a firmware write because the bench camera already fails GVD/HWMC reads before DFU, so #606 cannot cache the prototype PID in the current boot.

kernel/realsense/d4xx.c
Nir-Az: Thanks — the shared helper plus the flush hook covers the aligned case. One thing the new path introduces though: `flush` finalizes whenever the state is still `IN_PROGRESS`, with no completeness check, so an aborted transfer now sends `0x4a04` on a partial image — Ctrl-C on rs-fw-update, a viewer cancel, a crashed process, or a short read of the image file. Before this commit `release` just dropped to `DS5_DFU_IDLE` and nothing was manifested, so aborting mid-image was safe. Can we gate the flush finalization on an actual end-of-image signal rather than `IN_PROGRESS` alone? On a crashed process the `close()` return is discarded as well, so it would fail silently.

kernel/realsense/d4xx.c
lemon6630: Addressed in 81ce083 and retained at current head 4e72952. Recovery no longer writes the guessed family into cached_device_type. The open path initializes a DFU-session-local manifest_state, and ds5_dfu_detach narrows FW major 7..8 only to that field. cached_device_type remains reserved for the firmware value read from DS5_DEVICE_TYPE, so readiness and RGB register routing cannot be short-circuited by the recovery hint.

kernel/realsense/d4xx.c
lemon6630: You are right that close is not an independent proof of end-of-image. For a block-aligned multi-write stream, the driver receives neither an expected total length nor a separate commit operation, so flush cannot distinguish intentional EOF from an abort. D5xx also accepts variable image sizes, so byte counting cannot safely close that gap inside this PR alone. I traced the current HKR side as well: download-complete passes the buffered bytes and actual size to dfu_update_v2, which validates the compressed container, layout/size/CRC, authentication, and flash readback; a truncated image returns a DFU error and never reaches wait-reset or reboot. That prevents silent activation of a partial image, but it does not provide explicit caller intent or a recoverable close error after process death. I am therefore keeping this thread open. The complete fix needs a coordinated length-bearing start or explicit commit operation in the chardev ABI plus the SDK caller change; otherwise the choice is to drop aligned-image support or accept close as EOF with firmware validation.

ejgoldik (approved):

PR #566, opened 2026-08-07
[RSDEV-11627] Limit 980 kHz SerDes I2C acceleration to identified D585 links

lemon6630:
## Summary

Addresses [RSDEV-11627](https://rsjira.realsenseai.com/browse/RSDEV-11627): enable 980 kHz-class SerDes I2C master timing **only for a runtime-identified D585 camera and its own GMSL link**. D400 cameras on the same deserializer retain the pre-feature timing.

The eligibility check requires both `DS5_DEVICE_TYPE_D58X` and a GVD PID of `0x0c07` (D585 2C prototype) or `0x0c08` (D585 3C prototype). SerDes model and DT compatibility alone do not enable acceleration. Unrecognized PIDs and recovery probes without a confirmed identity remain at baseline; future production PIDs require an explicit qualification/allowlist update.

## Changes

- Remove unconditional speed changes from shared MAX96717/MAX96712/MAX96724 initialization. Those initialization function bodies match the pre-feature base, ignoring whitespace.
- After operational device/PID discovery, the primary camera instance requests acceleration for its MAX96717 and **only its `gmsl_link`** on MAX96712 or MAX96724. Other links are not programmed. Both pixel and tunnel paths use this runtime policy.
- Preserve the slave setup/hold registers, timeout/reserved fields and the existing 100–110 us write delay. Change only master `MST_BT` bits through locked read/modify/write operations.
- Save the original master timing in private per-serializer/per-link state. Restore it on primary teardown or probe failure; roll back both endpoints if enabling fails, warning and continuing probe/reset at baseline if both restores succeed. If rollback fails, retain the first enable error and report restoration errors; uncertain timing is not treated as a safe fallback. Reapply after camera reset only once device identity is valid again. Repeated enables retain the original saved value. Teardown restores timing after releasing the camera slot lock while retaining the topology lock to prevent slot reuse.
- Provide the new per-link API in both independent JP5 MAX96712 patch carriers and the shared JP6/JP7 carrier, plus their headers. MAX96717/MAX96724 remain single-source across generations. No existing public struct layout or exported function signature is changed; the module set must be rebuilt together for the new helper exports.
- Validate device/private-state/regmap pointers in every new SerDes helper. Use a shared D585 prototype PID predicate without changing the existing reset eligibility.
- Record the runtime-PID, per-link and shared-bus constraints in the existing CLAUDE.md.

### Shared host I2C bus

**Keep the pre-feature 400 kHz host-adapter request during DFU on JP5/JP6, with the existing restoration on release.** The adapter clock is shared: testing only the camera PID before requesting 980 kHz would still change the bus used by neighboring D400 devices. JP7 keeps its existing host-adapter behavior. This revision accelerates only the identified D585's SerDes master timing; it does not make a bus-wide 980 kHz host request.

| Camera/link | Serializer + selected deserializer link | Other deserializer links | Host DFU clock |
|---|---|---|---|
| Confirmed D585 PID, supported SerDes | 980 kHz master timing | Unchanged | Pre-feature behavior |
| D400 / D457 | Unchanged | Unchanged | Pre-feature behavior |
| Unknown PID, unidentified recovery, unsupported SerDes pair | Unchanged | Unchanged | Pre-feature behavior |

## Validation of this revision

- Production-function ASan/UBSan tests passed for all 16-bit PID values across 16 device-type values: only the two explicit D585 identities request enabling; unknown/D400 identities do not.
- SerDes helper tests passed for all 256 initial register values and every link: preserved unrelated bits and links, separate link enable/restore, reapply after simulated reset, saved original timing, read/write errors including a partially applied failed write, restore retry, original errno, lock release and write delays. Ran against all three MAX96712 patch carriers.
- Regenerated all three MAX96712 patches with accurate diff metadata and verified application/reversal against their source baselines. Compared the shared initialization functions and DFU open/release to the pre-feature base.
- ARM64 `-Werror` object compilation passed for `d4xx.c`, MAX96712, MAX96717 and MAX96724 with JP6.2.1 and JP7.1 build flags. The local build trees lacked some current branch headers; isolated test headers were reconstructed from this branch's existing patch stack. No installed build environment was modified.
- Both JP5 MAX96712 source variants and the common MAX96717/MAX96724 sources compiled with `-Werror` against the available JP5.0.2 configured kernel and matching branch headers. This is not a full independent JP5.1.2 build or a complete module/link/deployment test.
- Review follow-up tests passed for null/unbound devices and absent regmaps, safe fallback versus failed rollback, and primary-only teardown with the camera slot mutex released and topology mutex retained during I2C.
- New standalone test imports without running its compiler or parsing pytest arguments. Source whitespace checks passed.

Run the host-only checks with:

```sh
python3 test/standalone_tests/test_d585_i2c_scope.py \
    --max96712-source /path/to/applied/drivers/media/i2c/max96712.c
```

## Hardware qualification still pending

This PID-scoped revision has **not been deployed**. Mixed D585+D400 enumeration, controls, DFU, reset/unbind/reprobe, streaming and per-link timing readback remain hardware qualification work, including failure paths where I2C restoration cannot complete.

Historical D585 tunnel-mode DFU results (about 720 s to 475 s for roughly 19 MB) used the earlier global 980 kHz host setting and broader SerDes writes; **they are not performance results for this revision**. The previous MAX96712 pixel-mode register A/B showed why slave setup/hold timing must remain unchanged, but likewise does not qualify this new runtime policy. Keep the PR draft until the revised scope is qualified on hardware.


pinquanw (dismissed):

ymodlin (commented):
Automated review of the Fast-mode Plus change. Two notes; flagging now since the register values are the risky part.

nvidia-oot/max96724.c
```diff
@@ -1165,8 +1165,18 @@ int max96724_init_settings(struct device *dev)
 	int i;
 	struct max96724 *priv = dev_get_drvdata(dev);
 
+	/* Raise the port-0 link-A I2C master to Fast-mode Plus (980 kbps):
+	 * 0x640 SLV_SH=0b00/SLV_TO=0b001, 0x641 MST_BT=0b111/MST_TO=0b001. */
+	struct reg_pair i2c_fast_plus[] = {
+		{0x0640, 0x09},
```
ymodlin: **Potential BUG — only link A's control-channel master is raised.** `0x0640/0x0641` covers port-0 link A, but `DFU_I2C_BUS_CLK_RATE` is raised to 980 kHz for the whole Tegra bus, so cameras on the other links keep Fast-mode timing while the host drives Fm+. On the 4-camera fg12-16ch topologies that risks control-channel corruption during DFU on links B/C/D — either program all links or keep the host at 400 kHz until they are.

Also: `0x0640/0x0641/0x09/0x71` are magic numbers; this file otherwise uses named `MAX96724_*_ADDR`/value defines.

kernel/realsense/d4xx.c
```diff
@@ -307,7 +307,8 @@ enum ds5_mux_pad {
 #ifdef CONFIG_TEGRA_CAMERA_PLATFORM
 #define DFU_I2C_STANDARD_MODE		100000
 #define DFU_I2C_FAST_MODE			400000
-#define DFU_I2C_BUS_CLK_RATE		DFU_I2C_FAST_MODE
+#define DFU_I2C_FAST_MODE_PLUS		980000	/* Fm+; SerDes MST_BT=0b111 */
```
ymodlin, line 335: Worth noting in the commit message that this knob is a no-op on JP7: per CLAUDE.md the noble i2c-tegra rewrite ignores `adap->bus_clk_rate`, so Fm+ will only take effect on 5.x/6.x. `DFU_I2C_STANDARD_MODE` is now unused by anything — consider whether it still earns its keep.

nvidia-oot/max96724.c
pinquanw: Good catch. Updated MAX96724 port 0 to configure Fm+ timing on all four links: A (0x640/0x641), B (0x650/0x651), C (0x660/0x661), and D (0x670/0x671). The raw register addresses and 0x09/0x71 values are now represented by named MAX96724 macros. Force-pushed in d04f4bbeb.

kernel/realsense/d4xx.c
pinquanw: Agreed. Removed the unused DFU_I2C_STANDARD_MODE and DFU_I2C_FAST_MODE definitions. The amended commit message now states that the Tegra adapter-clock change is effective only on 5.x/6.x; JP7 uses the noble i2c-tegra implementation and leaves the host bus unchanged, while the SerDes Fm+ configuration still applies. Force-pushed in d04f4bbeb.

ymodlin:
Closing since there is no progress for almost a month.
reopen if needed.

nvidia-oot/max96724.c
ymodlin: change approved

kernel/realsense/d4xx.c

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
```diff
@@ -1058,6 +1066,22 @@ index 8e237ee..c1e57e2 100644
 +
 +	mutex_lock(&priv->lock);
 +
++	/*
++	 * Match the MAX96717 Fast-mode Plus timing on port 0 of every link:
++	 * SLV_SH=0b00 and MST_BT=0b111 (980 kbps). Preserve timeout and
++	 * reserved bits; only the I2C speed fields need to change.
++	 */
++	for (link = 0; link < MAX96712_MAX_LINKS; link++) {
++		err |= regmap_update_bits(priv->regmap,
```
ymodlin: **Potential BUG — bypasses the mandatory SerDes inter-command delay.** Every other register write in this driver goes through `max96712_set_registers()` → `max96712_write_reg()`, which ends with `usleep_range(100, 110)` / *"delay before next i2c command as required for SERDES link"*. `regmap_update_bits()` skips it, so this issues 8 back-to-back RMW transactions with zero spacing — and they are the *first* writes in `init_settings()`, on the control channel whose own timing they are changing. Suggest a small `max96712_update_bits()` helper (read via `max96712_read_reg()`, write via `max96712_write_reg()`) so the delay is preserved.

Identical in `kernel/nvidia/5.0.2/0017` (~L867) and `kernel/nvidia/5.1.2/0007` (~L950).

Minor, same lines: `err |= regmap_update_bits(...)` ORs negative errnos, so two different failures produce a value that is neither — the PR body's claim that "register errors propagate through the initialization return value" only holds as a non-zero/zero signal, not as a usable errno.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
```diff
@@ -1058,6 +1066,22 @@ index 8e237ee..c1e57e2 100644
 +
 +	mutex_lock(&priv->lock);
 +
++	/*
++	 * Match the MAX96717 Fast-mode Plus timing on port 0 of every link:
```
ymodlin: Nit: the comment claims it matches MAX96717 timing, but it doesn't fully — MAX96717/MAX96724 in this same PR force `SLV_TO`/`MST_TO = 0b001` (`0x09`/`0x71`), while this masked update leaves both timeouts at their power-on defaults. Either align the timeouts or reword to "speed fields only".

nvidia-oot/max96717.c
```diff
@@ -393,6 +393,15 @@ int max96717_init_settings(struct device *dev)
 		{MAX96717_GPIO8_B_ADDR, MAX96717_GPIO_B_NO_PULL},
 	};
 
+	/*
+	 * Raise the camera-side I2C master to Fast-mode Plus (980 kbps):
+	 * 0x40 SLV_SH=0b00/SLV_TO=0b001, 0x41 MST_BT=0b111/MST_TO=0b001.
+	 */
+	struct reg_pair i2c_fast_plus[] = {
+		{0x0040, 0x09},
```
ymodlin: Nit (carry-over): MAX96724 and now MAX96712 both use named macros; this is the last site still hardcoding `0x0040`/`0x0041` and `0x09`/`0x71`. `max96717.h` already follows the `MAX96717_*_ADDR` convention.

ymodlin (commented):
New commit `3cc7d97` (MAX96712 pixel-mode Fm+) reviewed.

Verified good: all three carriers (`5.0.2/0017`, `5.1.2/0007`, `6.0/0003`) carry a byte-identical block; hunk new-side counts, later-hunk offsets, diffstats and `index` post-image hashes are all consistent (per CLAUDE.md §"JP5 carries the whole max96712 driver in two independent patch files"). `MAX96712_MAX_LINKS` is defined before use; masked RMW correctly preserves the timeout/reserved bits.

One BUG below plus two nits.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
ymodlin: Still open after `c8b4806`. The loop is 4 RMWs now instead of 8, but `regmap_update_bits()` still bypasses the mandatory `usleep_range(100, 110)` that `max96712_write_reg()` applies, and these are still the *first* I2C transactions in `init_settings()` — on the control channel whose own timing they change. `err |=` on negative errnos is also unchanged.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
ymodlin: change approved — `SLV_SH` is gone and the now-unused `*_SLV_*` macros were removed with it.

One follow-up on the same block: CLAUDE.md "Comment style" forbids in-code comments that describe a removal or a deliberate omission. `Keep the existing host-facing SLV_SH timing: forcing SLV_SH=0b00 breaks reads on the tested 400 kHz host.` is exactly that — drop those two lines; the rationale already lives in the PR body.

nvidia-oot/max96717.c
ymodlin: Still open — `nvidia-oot/max96717.c` was not touched in this push, so `0x0040`/`0x0041`/`0x09`/`0x71` remain hardcoded while max96712 and max96724 now both use named macros.

ymodlin:
**Push review — `c8b4806` (max96712 pixel-mode links)**

Delta vs the previously reviewed `3cc7d97` is a strict reduction: the `SLV_SH` masked update and its three macros are dropped, only `MST_BT=0b111` on `0x0641/0x0651/0x0661/0x0671` remains, comment reworded. Verified: the new block is byte-identical across all three carriers (`5.0.2/0017`, `5.1.2/0007`, `6.0/0003`), the JP5.1.6 / 6.1 / 6.2 / 7.0 / 7.1 symlinks still resolve, and every hunk's line counts, later-hunk offsets and diffstat are self-consistent in all three patches.

No new defect introduced. Two prior comments remain unaddressed (SerDes inter-command delay bypassed by `regmap_update_bits`; max96717 magic numbers) — replied inline. Hardware regression listed in the PR body is still the gate.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
pinquanw: Agreed; this still applied to the four-link loop. Fixed in 221dedd1c.

`max96712_update_bits()` now reads through `max96712_read_reg()` and writes through `max96712_write_reg()`, retaining the existing 100–110 us post-write delay. The caller holds `priv->lock` across the RMW. The speed loop stops at the first failure, unlocks, and returns the original errno instead of OR-ing errors.

The implementation is identical in all three patch carriers. Patch-application and cross-compile checks passed; fault injection at each link's read/write verified the delay, first errno, early exit and lock release. `SLV_SH`, timeout and reserved bits remain unchanged.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
pinquanw: Addressed in 221dedd1c. The in-code comment is now a single line describing the purpose of the port-0 speed setting. The measured `SLV_SH` regression and the requirement to retain the existing slave timing are documented in the PR description and commit message.

The MAX96712 update changes only `MST_BT`; its timeout fields remain unchanged, so the description no longer implies that all timing fields match MAX96717/MAX96724.

nvidia-oot/max96717.c
pinquanw: Agreed; fixed in 221dedd1c. Added `MAX96717_I2C0_ADDR`, `MAX96717_I2C1_ADDR`, `MAX96717_I2C_FAST_PLUS_SLAVE_VAL` and `MAX96717_I2C_FAST_PLUS_MASTER_VAL` alongside the existing private register definitions in `max96717.c`. The initialization table now uses those names; register addresses and values are unchanged. JP6.2.1 and JP7.1 cross-compilation passed with `-Werror`.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
ymodlin: change approved

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
ymodlin: change approved

nvidia-oot/max96717.c
ymodlin: change approved

nvidia-oot/max96724.c
```diff
@@ -1240,8 +1251,35 @@ int max96724_init_settings(struct device *dev)
...
 	struct max96724 *priv = dev_get_drvdata(dev);
 
+	/*
+	 * Raise the port-0 I2C master on every link to Fast-mode Plus
+	 * (980 kbps): SLV_SH=0b00/SLV_TO=0b001 and
+	 * MST_BT=0b111/MST_TO=0b001.
+	 */
+	struct reg_pair i2c_fast_plus[] = {
+		{MAX96724_I2C_P0_SLAVE_ADDR(0),
+		 MAX96724_I2C_FAST_PLUS_SLAVE_VAL},
```
ymodlin: **Potential BUG — this is the exact write the MAX96712 rig proved harmful.** MAX96724 uses the same port-0 I2C map as MAX96712 (`0x0640`/`0x0641`, stride `0x10`, defaults `0x26`/`0x76`). On the max96712 side this PR was reduced to a masked `MST_BT`-only update after `0x0640=0x06` failed 30/30 FW-version reads; here the slave regs `0x0640/0x0650/0x0660/0x0670` are still written whole with `0x09`, clobbering `SLV_SH`/`SLV_TO` on all four links. The tunnel bench predates that finding, and the `255.255.243.85` FW-version read noted in the PR body is the same symptom class.

Suggest the same treatment as max96712: masked `MST_BT` update, leave the slave regs alone. Also, the comment's `SLV_SH=0b00/SLV_TO=0b001` does not decode from `0x09` under that map.

Minor: `MAX96724_MAX_LINKS` already exists — the 8-entry table could be the same 4-iteration loop used in max96712.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
ymodlin: change approved

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
ymodlin: change approved

nvidia-oot/max96717.c
ymodlin: change approved

nvidia-oot/max96724.c
```diff
@@ -1240,8 +1251,35 @@ int max96724_init_settings(struct device *dev)
 	int i;
 	struct max96724 *priv = dev_get_drvdata(dev);
 
+	/*
+	 * Raise the port-0 I2C master on every link to Fast-mode Plus
+	 * (980 kbps): SLV_SH=0b00/SLV_TO=0b001 and
+	 * MST_BT=0b111/MST_TO=0b001.
+	 */
+	struct reg_pair i2c_fast_plus[] = {
+		{MAX96724_I2C_P0_SLAVE_ADDR(0),
```
ymodlin: **Potential BUG — the `SLV_SH=0b00` that was reverted on MAX96712 is still forced here.** This PR dropped `0x0640=0x09` on MAX96712 because forcing `SLV_SH=0b00` on the host-facing slave failed 30/30 register reads at 400 kHz. MAX96724's `0x0640/0x0650/0x0660/0x0670` are the same host-facing slave registers and still get `0x09`. The PR's own tunnel-mode data shows the matching symptom — the bogus `255.255.243.85` DEPTH firmware-version read — and has not been rerun for this update. Suggest limiting MAX96724 to a masked `MST_BT` update as well, or re-running tunnel mode before merge.

ymodlin:
Don't do merge commit, rebase and make sure there is one commit

nvidia-oot/max96724.c
pinquanw: The master-only update is a reasonable fix; implemented in e99a9aad0. MAX96724 now loops over `MAX96724_MAX_LINKS`, updates only `MST_BT` in `0x0641/0x0651/0x0661/0x0671`, and preserves all slave timing, timeout and reserved fields. The RMW runs under `priv->lock`, writes through the existing delayed wrapper, and returns the first error after unlocking.

JP6.2.1/JP7.1 cross-compiles and extracted-function fault-injection checks passed. Tunnel-mode hardware regression remains pending; the earlier invalid version read is not yet proven to have this cause.

One decoding clarification: `0x09` does select `SLV_SH=00` (bits 5:4) and `SLV_TO=001` (bits 2:0); it also sets bit 3. The documented master reset value is `0x56`; `0x76` is the 980 kbps selection with the default timeout. These details do not change the rationale for replacing the whole-register writes.

nvidia-oot/max96724.c
pinquanw: Addressed in e99a9aad0: MAX96724 now uses the same master-only masked-update approach as MAX96712, covering all four links and retaining existing slave timing and timeout values. Write spacing and first-error propagation are preserved. Compile and fault-injection checks passed; the PR description explicitly requires a new tunnel-mode hardware regression and distinguishes it from the historical bench results.

pinquanw:
@ymodlin Agreed. Rebased onto the latest `dev` (`a57b229c8`), removed the merge commit, and folded the MAX96724 review fix into `e99a9aad0`. The PR now contains exactly one non-merge commit. Force-pushed with an explicit lease against the previously reviewed head.

nvidia-oot/max96724.c
ymodlin: change approved

nvidia-oot/max96724.c
ymodlin: change approved

nvidia-oot/max96717.c
```diff
@@ -393,6 +397,15 @@ int max96717_init_settings(struct device *dev)
 		{MAX96717_GPIO8_B_ADDR, MAX96717_GPIO_B_NO_PULL},
 	};
 
+	/*
+	 * Raise the camera-side I2C master to Fast-mode Plus (980 kbps):
+	 * 0x40 SLV_SH=0b00/SLV_TO=0b001, 0x41 MST_BT=0b111/MST_TO=0b001.
+	 */
+	struct reg_pair i2c_fast_plus[] = {
+		{MAX96717_I2C0_ADDR, MAX96717_I2C_FAST_PLUS_SLAVE_VAL},
```
ymodlin: **Potential BUG — max96717 is now the only site still doing a whole-register write.** This push moved max96724 to a masked `MST_BT`-only update (like max96712) exactly because forcing `SLV_SH=0b00` failed 30/30 FW-version reads on HW. `0x0040 = 0x09` here still forces `SLV_SH=0b00`, changes `SLV_TO` off its power-on value, and sets bit 3 — which is not one of the two fields the comment names (you noted "it also sets bit 3" in the max96724 thread). Only `0x0041` `MST_BT` is needed for the 980 kbps master.

Suggest dropping the `0x0040` entry and making `0x0041` a masked `MST_BT` update, so all three devices follow the same rule. If the `0x0040` write is deliberate, the PR body should carry the HW data that isolates it.

Minor, same block: the comment says "camera-side I2C master" but `0x0040` is the slave register, and per CLAUDE.md "Comment style" it should state the *why*, not decode the fields.

pinquanw: Addressed in 04b22bb06. MAX96717 now updates only 0x0041 MST_BT with mask/value 0x70; 0x0040 and master timeout/other bits are preserved. This removes unvalidated changes beyond the speed increase—the MAX96712 A/B failure does not prove a MAX96717 fault.

The caller-held mutex, existing 100–110 us write delay and exact first-error return are retained; GPIO prerequisite or speed-update failures exit immediately. The comment now states the purpose. Pixel/tunnel fault-injection tests passed; full driver build and HW regression of this latest change are pending. Updated the PR description and kept the PR at one commit.

ymodlin: change approved

ymodlin:
Reviewed `04b22bb` (only `nvidia-oot/max96717.c` changed). No bugs found in the new code — all three devices now do the same masked `MST_BT`-only update, the RMW goes through the delayed write wrapper under `priv->lock`, and the early-exit returns the first errno unchanged.

One doc nit before merge: the PR body still says *"MAX96717: configure `0x0040=0x09` and `0x0041=0x71` during initialization"*. `0x0040` is no longer written and `0x0041` is now a masked update — worth correcting, since the slave-timing question is what the pending HW regression is gating on.

kernel/realsense/d4xx.c
```diff
@@ -9294,6 +9345,16 @@ static int ds5_probe(struct i2c_client *c
 			(state->fw_version >> 8) & 0xff, state->fw_version & 0xff,
 			(state->fw_build >> 8) & 0xff, state->fw_build & 0xff);
 
+#ifdef CONFIG_VIDEO_D4XX_SERDES
+	if (state->ser_primary) {
+		mutex_lock(&serdes_lock__);
+		ret = ds5_configure_d585_i2c(state, true);
+		mutex_unlock(&serdes_lock__);
+		if (ret)
+			goto e_chardev;
```
ymodlin, line 9362: **Potential BUG — an optional speed-up now fails the whole probe.** `ds5_configure_d585_i2c()` already rolls both endpoints back to baseline on failure, so a transient I2C error on the `MST_BT` write leaves a fully working camera at pre-feature timing — but `goto e_chardev` tears down the DFU chardev and aborts probe, so the camera never enumerates (no `/dev/video*`).

Suggest `dev_warn()` + continue: acceleration is an optimization, not a probe prerequisite.

kernel/realsense/d4xx.c
```diff
@@ -4345,6 +4384,14 @@ static int ds5_hw_reset_with_recovery(struct ds5 *state)
 		"%s(): GMSL link recovered (device type 0x%04x)\n",
 		__func__, dev_type);
 
+#ifdef CONFIG_VIDEO_D4XX_SERDES
+	mutex_lock(&serdes_lock__);
+	ret = ds5_configure_d585_i2c(state, true);
+	mutex_unlock(&serdes_lock__);
+	if (ret)
+		return ret;
```
ymodlin, line 4401: **Potential BUG — same class as the probe site.** A failed acceleration write aborts `ds5_hw_reset_with_recovery()` before the FW-version verification and the remaining recovery steps, turning a benign optimization failure into a failed camera recovery. Timing is already rolled back at this point — warn and continue.

kernel/realsense/d4xx.c
```diff
@@ -5899,6 +5946,10 @@ static bool ds5_release_slot(struct ds5 *state)
 
 	mutex_lock(&state->ds5_dev->lock);
 	if (state->ds5_dev->ds5_primary == state) {
+		ret = ds5_configure_d585_i2c(state, false);
```
ymodlin: **Potential BUG — I2C under `ds5_dev->lock`.** This runs 2 reads + 2 writes with `usleep_range(100, 110)` on the serializer and deserializer while holding `serdes_lock__` **and** the per-camera slot lock; the probe/recovery call sites hold only `serdes_lock__`. This is the unbind/teardown path, where the camera may already be gone — the transfers then retry/time out while sibling subdevs block on `ds5_dev->lock`.

CLAUDE.md §Concurrency: snapshot under lock, do I2C after unlocking. Moving the call below `mutex_unlock(&state->ds5_dev->lock)` and gating it on `released` keeps the same semantics.

Minor, same hunk: the replacement comment narrates what the code does rather than the why, and drops the locking contract — CLAUDE.md §Concurrency records that this helper takes `serdes_lock__` internally.

kernel/realsense/d4xx.c
```diff
@@ -975,6 +980,40 @@ static inline bool ds5_is_d58x(struct ds5 *state)
...
+static int ds5_configure_d585_i2c(struct ds5 *state, bool enable)
+{
+	u16 pid = READ_ONCE(state->ds5_dev->d585_product_id);
+	int ret, dser_ret, restore_ret;
+
+	if (!state->ser_ops->set_i2c_fast_mode ||
+	    !state->dser_ops->set_i2c_fast_mode)
+		return 0;
+	if (enable && (!ds5_is_d58x(state) ||
+	    (pid != D585_2C_PROTO_PID && pid != D585_3C_PROTO_PID)))
```
ymodlin: Nit: third copy of the D585 prototype-PID test (also ~L4140 and ~L4859). A small `ds5_is_d585_proto(state)` inline keeps the allowlist in one place — CLAUDE.md requires an explicit update when a production PID is qualified, which is easier to get right with a single site.

nvidia-oot/max96717.c
```diff
@@ -403,6 +438,7 @@ int max96717_init_settings(struct device *dev)
 	err |= max96717_set_registers(dev, gpio_release,
 				     ARRAY_SIZE(gpio_release));
 
+
```
ymodlin: Nit: stray blank line — the only change to `max96717_init_settings()`, which the PR states matches the pre-feature base. Drop it.

ymodlin (commented):
Re-reviewed in full — commit `1339e4a` replaces the approach (runtime PID gating + per-link save/restore), so the earlier approvals are superseded.

Re-verified and still good:
- No speed writes left in any `init_settings()`; all three max96712 carriers carry a byte-identical `max96712_set_i2c_fast_mode()`, and hunk offsets/counts validate in all three.
- Only `MST_BT` is touched (mask/value `0x70`); slave setup/hold, timeouts and reserved bits are preserved.
- Writes go through the delayed `max9671x/max9672x_write_reg()` wrappers, the RMW is under `priv->lock`, and the first errno is returned (no `err |=`). No nested locking between the helper and the write wrapper.
- Named macros everywhere; `link` is bounds-checked before indexing.

5 findings below — 3 potential BUGs, 2 nits.

kernel/realsense/d4xx.c
pinquanw: Agreed; fixed in 6c3e9ff59. The shared helper now warns and returns success after both endpoints restore baseline timing, so optional acceleration failure no longer aborts probe. If either rollback fails, it still reports the rollback errors and propagates the original enable error: a working baseline cannot be assumed in that case. Fault-injection tests cover both outcomes.

kernel/realsense/d4xx.c
pinquanw: Fixed in 6c3e9ff59 through the shared helper: reset recovery continues to firmware verification after a failed speed-up with successful baseline restoration. Failed rollback remains fatal rather than hiding uncertain link timing. Fault-injection and JP6.2.1/JP7.1 object compilation passed; hardware recovery qualification remains pending.

kernel/realsense/d4xx.c
pinquanw: Agreed; fixed in 6c3e9ff59. Timing restoration now runs after releasing ds5_dev->lock, gated on released, while retaining serdes_lock__ to prevent slot reuse. The comment records that internal locking contract. An extracted-production-function test verifies primary-only restoration, the lock state during I2C, shared-deserializer ownership, and cleanup when restoration fails.

kernel/realsense/d4xx.c
pinquanw: Addressed in 6c3e9ff59. ds5_is_d585_proto() now owns the PID allowlist and is reused by all three sites. Acceleration additionally requires runtime D58x device type; existing PID-only reset eligibility is unchanged. The exhaustive type/PID policy test passed.

nvidia-oot/max96717.c
pinquanw: Removed in 6c3e9ff59. MAX96717 and MAX96724 init_settings() now match the pre-feature base exactly, including whitespace.

pinquanw:
Addressed the five latest review findings and the Aikido null-pointer finding in `6c3e9ff59`; the PR is now exactly one commit, force-pushed with a lease, and remains draft. Replied to all five available review threads.

The original Aikido inline comment (`4045104316`) disappeared after the push (reply API returns 404; it is no longer in the comment list), so recording its response here: null device/private-state/regmap checks now precede mutex access in every new SerDes speed helper, including all MAX96712 patch carriers. Invalid-input ASan/UBSan tests pass. No manual ignore is needed for that finding.

The PR description now reflects master-only masked writes and PID/per-link scope. Optional acceleration failure falls back only after successful restoration of both endpoints; failed rollback remains fatal. Teardown performs I2C outside the camera slot lock while retaining topology protection.

Fault-injection tests and JP6.2.1/JP7.1 object compilation passed; the JP5 source variants compiled against the available JP5.0.2 kernel. Mixed-camera hardware regression remains pending; no deployment was performed.

kernel/realsense/d4xx.c
ymodlin: change approved

kernel/realsense/d4xx.c
ymodlin: change approved

kernel/realsense/d4xx.c

kernel/realsense/d4xx.c

nvidia-oot/max96717.c
ymodlin: change approved

ymodlin (dismissed):

kernel/realsense/d4xx.c
```diff
@@ -4345,6 +4393,14 @@ static int ds5_hw_reset_with_recovery(struct ds5 *state)
 		"%s(): GMSL link recovered (device type 0x%04x)\n",
 		__func__, dev_type);
 
+#ifdef CONFIG_VIDEO_D4XX_SERDES
+	mutex_lock(&serdes_lock__);
+	ret = ds5_configure_d585_i2c(state, true);
```
ymodlin: **Potential BUG — reset path re-enables acceleration without the `ser_primary` gate.**

Probe (L9357) and teardown (`ds5_release_slot`) both gate on `state->ser_primary`; this site does not. `ds5_s_ctrl` opcode `0x20` runs on any of the camera's four subdev instances, and CLAUDE.md records that a sibling subdev can outlive the primary that owns the `ds5_dev` slot (sysfs unbind).

So after the primary released the slot, a sibling-triggered reset re-arms 980 kHz on the serializer + `gmsl_link` with nothing left to restore it — `ds5_release_slot()` won't run again for that slot. A D400 probing that link afterwards inherits the accelerated master timing, since its own `ds5_configure_d585_i2c()` returns early at the D58x check.

Suggest gating this call on `state->ser_primary` too.

ymodlin (commented):
**rs-agentic review — push `c65edddf` (synchronize)**

This push is a **rebase onto `dev`**, not a code change: the only content deltas vs the previously reviewed `6c3e9ff5` are dev's FG12 12CH PoC work (`fangzhu,fg12-4ch/12ch-poc-enable`, `MAX96712_FANGZHU_FG12_*_POC_ADDR`) picked up in the three max96712 carriers, plus the new 12ch overlay/README files that belong to `dev`. No feature code changed.

Rebase sanity checks pass: all three regenerated carriers are internally consistent — every `@@` hunk's line counts match its body, and the `1 file changed, N insertions(+), M deletions(-)` trailers match the actual counts (5.0.2 +1277/-3, 5.1.2 +1285/-40, 6.0 +1274/-35). All 13 prior threads already carry a `change approved` reply.

One finding below that predates this push but is still open. Not approving — the PR's own hardware qualification is still pending.

ymodlin:
**Re-review of `2d7b6e7` — rebase-only push, no new findings.**

This `synchronize` was a pure rebase of `c65edddf` onto the new `dev` tip (`7924138`, "Auto-increment version to 1.0.7.2"). I diffed every file in the PR between the two heads: all twelve are byte-identical except `MODULE_VERSION("1.0.7.1")` → `("1.0.7.2")` in `kernel/realsense/d4xx.c`. No functional delta, so no new review was warranted and none was submitted.

**New findings posted this run:** 0 (BUG 0 / RACE 0 / SECURITY 0 / PERF 0 / DT 0).

**Prior threads triaged (14 root, all mine):**
- ALREADY_SETTLED — 13. Each already carries my verified "change approved"; the anchors are unchanged by this rebase.
- STILL_OPEN — 1: [`d4xx.c:4398`](https://github.com/realsenseai/realsense_mipi_platform_driver/pull/566#discussion_r4070150394) (missing `ser_primary` gate on the reset-path re-enable). No reply and no code change at the anchor, so I am not restating it — re-verified at `2d7b6e7` only to confirm it is not stale: `ds5_init_ds5_dev()` sets `ds5_dev->ds5_primary = state` on the same instance that gets `ser_primary = true`, probe (L9357) and `ds5_release_slot()` (L5963, via `released`) both gate on that instance, and the reset site at L4396 still does not.
- HUMAN_REPLIED / FIXED / PARTIALLY_FIXED / NOT_FIXED / REGRESSED — 0.

**Suppressed as duplicates:** 1 — a candidate note that `ds5_release_slot()` calls `ds5_configure_d585_i2c()` outside `#ifdef CONFIG_VIDEO_D4XX_SERDES`. Discarded, not posted: the enclosing function already references `serdes_lock__` (defined at L812 inside that guard) at preprocessor depth 0 on `dev`, so this is pre-existing and not introduced here.

**Review event submitted:** none. Zero findings to batch, and an empty review on an unchanged diff would be noise.

**Prior REQUEST_CHANGES dismissed this run:** no — none is active. My earlier change request (`5275176596`) was already dismissed on 2026-09-22 at `6c3e9ff5`; my latest review (`5276220560`) is COMMENTED, so there is nothing to clear.

**Wants hardware validation** (carried over, unchanged by the rebase): the tunnel-mode regression for the MAX96724 master-only `MST_BT` update, and the post-reset recovery path on a real D585 — the earlier `255.255.243.85` FW-version read was never proven to have been caused by the reverted whole-register slave writes.

No PR approval was submitted.

_Auto-generated by the rs-agentic reviewer._

kernel/realsense/d4xx.c
lemon6630: Fixed in 6d41134b9. The post-reset 980 kHz reapply now uses the same `state->ser_primary` gate as probe and teardown. A sibling subdev can still complete reset recovery and FW verification, but it can no longer re-arm SerDes acceleration after the primary has released the slot. The D585 scope/fault-injection suite passes, and JP6.2 `d4xx.o` cross-compiles with `-Werror`.

ymodlin: change approved - verified at `6d41134`: the reset-path call at `kernel/realsense/d4xx.c:4397` is now wrapped in `if (state->ser_primary)`.

`ds5_setup_and_link()` sets `state->ser_primary = false` on every peer instance (L6028), so a sibling-triggered `ds5_hw_reset_with_recovery()` can no longer re-arm 980 kHz. The post-release variant is closed too: once `ds5_release_slot()` has cleared `ds5_dev->ds5_primary`, the `d585_proto_reset` block at L4294-4298 hits `if (!primary) return -ENODEV` and bails long before L4397, so the un-restorable re-arm cannot happen on either path.

**One side effect the gate introduced** (not a new thread - same call site):

`ds5_hw_reset_with_recovery()` resets and re-initializes the *primary's* serializer through `primary = state->ds5_dev->ds5_primary` (L4294-4325: `reset_control` / `recover_link` / `setup_control` / `init_settings`), but re-applies acceleration through `state`. Those are different handles for a sibling-initiated reset.

Failure: `DS5_CAMERA_CID_HWMC` opcode `0x20`, `DS5_CAMERA_CID_HWMC_RW` opcode `0x20` and `DS5_CAMERA_CID_HW_RESET` are registered on every subdev with `sid` in `[DEPTH_SID, IMU_SID)` (L6883-6913), so an SDK reset issued on the RGB or IR node of a D585 runs with `state->ser_primary == false`. The primary's MAX96717 is put back through `init_settings()` - which this PR deliberately restored to the pre-feature base, i.e. baseline `MST_BT` - while L4397 is skipped. The deserializer's `gmsl_link` keeps its 980 kHz `MST_BT` and `max96717`'s `priv->i2c_fast_mode` still reads `true`, so the two endpoints disagree and nothing re-syncs them until teardown. Whether the serializer register actually survives the camera reset is the hardware-dependent part; the driver's own comment at L4182-4185 ("restore the serializer alias and control tunnel ... after the SDK-triggered HW reset") says it does not.

Verdict: PLAUSIBLE - the skipped re-apply is certain from the code; the resulting endpoint mismatch needs a scope/bench check on a sibling-node reset.
Fix: re-apply against the same handle the block above already uses, e.g. gate on `state->ds5_dev->ds5_primary` and pass that instance to `ds5_configure_d585_i2c()`.

_Auto-generated by the rs-agentic reviewer. Thread-level only; not a PR approval._

ymodlin (commented):
Reviewed `6d41134` (full re-review of the diff vs `origin/dev`: `kernel/realsense/d4xx.c`, the three MAX96712 patch carriers, `max96717.c`/`max96724.c` + headers, CLAUDE.md, new standalone test).

No new findings this run. All 14 prior threads I opened are settled: 13 previously verified fixed, and the last one (reset path missing the `ser_primary` gate) is verified fixed at this SHA - the gate is present at `d4xx.c:4397`, and the post-release variant is additionally blocked by the `if (!primary) return -ENODEV` bail at L4294-4298.

One non-blocking side effect of that gate is noted as a reply in the existing thread rather than a new comment: a sibling-initiated D585 reset re-initializes the *primary's* serializer via `state->ds5_dev->ds5_primary` but re-applies acceleration via `state`, so acceleration is silently not re-applied while the deserializer link keeps its 980 kHz `MST_BT`. Marked PLAUSIBLE - needs a bench check.

Still needs hardware validation, not statically verifiable: (1) tunnel-mode regression on MAX96724 after the master-only masked update; (2) a sibling-node HWMC `0x20` reset on a D585 with a scope on the control channel; (3) that `max96717_init_settings()` after a camera power cycle really leaves `0x0041 MST_BT` at its default.

No PR approval was submitted.

_Auto-generated by the rs-agentic reviewer._

ymodlin:
## rs-agentic reviewer - run report for `6d41134`

### New findings posted
**0 new root comments.** Nothing in this push produced a groundable production defect that was not already under discussion.

Coverage this run (full re-read, not diff-context-only):
- `kernel/realsense/d4xx.c` - `ds5_configure_d585_i2c()`, `ds5_is_d585_proto()`, the probe enable/`e_chardev` restore, the reset-path re-enable, `ds5_release_slot()`, `ds5_setup_and_link()` slot/primary assignment, `ds5_remove()` ordering.
- `nvidia-oot/max96717.c`, `nvidia-oot/max96724.c` + `max96712.h`/`max96717.h`/`max96724.h`.
- All three MAX96712 patch carriers (`nvidia-oot/6.0/0003`, `kernel/nvidia/5.0.2/0017`, `kernel/nvidia/5.1.2/0007`).
- `CLAUDE.md`, `test/standalone_tests/test_d585_i2c_scope.py`.

Checks that came back clean and are worth recording:
- **Patch hygiene.** All three carriers have self-consistent hunk arithmetic. 6.1/6.2/6.2.1/6.2.2 and 7.0/7.1/7.2 all symlink to `nvidia-oot/6.0/0003`, and `kernel/nvidia/5.1.6/0007` symlinks to `5.1.2/0007`, so the three edited files cover every JetPack. The added `max96712_set_i2c_fast_mode()` body is byte-identical across all three (md5 `a959229...`), and `MAX96712_MAX_LINKS` is defined ahead of `struct max96712` in each.
- **JP5 build surface.** `apply_patches.sh` L97-111 and `apply_patches_ext.sh` L55-71 link the same `nvidia-oot/max96717.[ch]` / `max96724.[ch]` into the JP5 tree, and both `kernel/nvidia/max96712.h` and `nvidia-oot/max96712.h` gained the declaration, so the new ops resolve on 5.x as well as 6.x/7.x.
- **Kernel ABI.** Only new `EXPORT_SYMBOL`s; the new members went into the private `struct max96712`/`max96717`/`max96724` in `.c` files, not into any public header. No genksyms CRC exposure, nothing that rebuilds the NVIDIA display stack.
- **Locking.** `ds5_configure_d585_i2c()` is called under `serdes_lock__` only, never under `ds5_dev->lock`; `ds5_release_slot()` drops `ds5_dev->lock` before the I2C restore and keeps `serdes_lock__`; neither `ds5_remove()` (L9420) nor the `ds5_serdes_setup()` error path (L6588) holds `serdes_lock__` on entry, matching the CLAUDE.md contract. The SerDes helpers take only their own `priv->lock`, and `max96712_read_reg`/`max96712_write_reg`/`max96717_write_reg` are lock-free, so the RMW does not self-deadlock.
- **Error ladders.** `ds5_probe()`'s `e_chardev` restores timing before chardev teardown, `err` is declared unconditionally at the top of the function, and the enable-path rollback is idempotent against the second restore attempt.

Not a finding, recorded for the record: `ds5_release_slot()` sits outside any `CONFIG_VIDEO_D4XX_SERDES` guard while calling a helper defined inside one - but the function already references `serdes_lock__` and `struct dser_control` on `dev`, so a `!CONFIG_VIDEO_D4XX_SERDES` build was already broken before this PR. Pre-existing, out of scope here.

### Prior threads triaged (14 total)
| Verdict | Count | Threads |
|---|---|---|
| ALREADY_SETTLED | 13 | 3766186721, 3766186728, 4034147932, 4034147937, 4034147940, 4035947169, 4035989262, 4043292040, 4045103421, 4045103432, 4045103437, 4045103445, 4045103449 |
| FIXED (verified this run) | 1 | 4070150394 |
| HUMAN_REPLIED / PARTIALLY_FIXED / NOT_FIXED / REGRESSED / STILL_OPEN | 0 | - |

**4070150394** (reset path re-enables acceleration without the `ser_primary` gate) - verified at `6d41134`: the call at `d4xx.c:4397` is now gated on `state->ser_primary`, and `ds5_setup_and_link()` sets `ser_primary = false` on every peer instance (L6028), so a sibling-triggered reset can no longer re-arm 980 kHz. The post-release variant is independently blocked: once `ds5_release_slot()` has cleared `ds5_dev->ds5_primary`, the `d585_proto_reset` block at L4294-4298 returns `-ENODEV` before reaching L4397. This matters because `ds5_release_slot()` does *not* clear `state->ser_primary`, and `ds5_remove()` only unregisters the subdev (`ds5_mux_remove()`) long after the restore - without that `!primary` bail the window would still be open.

### Findings suppressed as duplicates
None. One observation was folded into an existing thread instead of becoming a root comment (per the no-duplicate-thread rule): the `ser_primary` gate makes the reset path re-apply acceleration against `state` while the serializer reset/re-init immediately above uses `state->ds5_dev->ds5_primary`. A D585 reset issued on the RGB or IR node (`DS5_CAMERA_CID_HWMC` / `HWMC_RW` opcode `0x20` and `DS5_CAMERA_CID_HW_RESET` are registered on every subdev with `sid` in `[DEPTH_SID, IMU_SID)`, L6883-6913) therefore re-initializes the primary's MAX96717 to baseline `MST_BT` and skips the re-apply, leaving the deserializer link at 980 kHz and `priv->i2c_fast_mode` reading `true`. Marked **PLAUSIBLE** - the skipped re-apply is certain from the code, the endpoint mismatch depends on whether the serializer register survives the camera reset. Posted as a reply on thread 4070150394.

### Review event
**COMMENT** (review `5286358396`). No new CONFIRMED BUG/RACE/SECURITY finding was raised this run, so the threshold for REQUEST_CHANGES was not met.

### Dismissal
Not applicable. My earlier REQUEST_CHANGES (`5275176596`) was already dismissed on 2026-09-22; my most recent review before this run was `COMMENTED`. There is no active change request of mine to clear.

### Wants hardware validation
1. Tunnel-mode regression on MAX96724 after the master-only masked update - the author flagged this as pending, and the historical bogus `255.255.243.85` DEPTH firmware-version read has not been re-tested against the current implementation.
2. A sibling-node HWMC `0x20` reset on a D585, with a scope on the control channel, to settle whether the skipped re-apply leaves a real timing mismatch.
3. Confirmation that `max96717_init_settings()` after a camera power cycle leaves `0x0041 MST_BT` at its hardware default - this is what turns item 2 from cosmetic into a real asymmetry.
4. That a D400 sharing the deserializer with a D585 is genuinely unaffected on its own `gmsl_link` while the D585 link runs at 980 kHz.

**No PR approval was submitted.**

_Auto-generated by the rs-agentic reviewer._

ymodlin:
This change will affect d4xx cameras that are on the same bus.
We need to talk about this offline, for now this PR cannot be approved.

ymodlin (changes requested):
Will change -1 after the design meeting.

PR #568, opened 2026-08-09
max96712 update 4 lane rate to 1300Mbps

ejgoldik:
- Updated max96712 4 lane rate to 1300Mbps (From 1100)
- Updated HKR lane rate to 1200Mbps (From 1000)


polejoe (changes requested):
If it need to modify all these dts?  looks most are for D4x.
not sure if D5x use all these dts?

ejgoldik:
> If it need to modify all these dts? looks most are for D4x. not sure if D5x use all these dts?

It's for both, as we change the rate for all 4 lanes setups, which include ds5

ghu3 (approved):

polejoe (approved):

ymodlin (approved):

PR #567, opened 2026-08-07
[D585][GMSL] Set tunnel mode CSI lane rates to 1300M Serializer / 1500M Deserializer

polejoe:
## Depends on

https://github.com/realsenseai/realsense_mipi_platform_driver/pull/568

## Summary

- Configure the D5x FG24/MAX96724 tunnel path for 1300 Mbps/lane from D585 to MAX96717 and 1500 Mbps/lane from MAX96724 to Host, using four lanes on both CSI segments.
- Make 1500 Mbps/lane the MAX96724 fallback and keep the device-tree lane rate, NVIDIA bandwidth value, and per-format SerDes pixel clocks consistent.
- Apply the camera-side override only to D58x and validate it against firmware-reported MIPI capabilities. D4xx always keeps its existing 1000 Mbps/lane rate and its existing lane count.
- Keep continuous clock enabled and deskew calibration disabled at 1500 Mbps/lane.

## Validation

- Jetson67, JP 6.2.2 / L4T 36.5 / kernel 5.15.185-tegra, D5x tunnel mode with MAX96717/MAX96724.
- Full clean build and deployment passed. The live device tree reported 4 lanes, 1300 Mbps/lane input, 1500 Mbps/lane output, continuous clock, and deskew disabled.
- Direct V4L2, wire metadata enabled: Depth Z16 + IR Y8I + RGB YUYV, all 1280x960@60, 5 seconds per cycle, 300 consecutive start/stop cycles.

| Path | Host good frames | Host FPS mean (min-max) | V4L2 error / sequence drop | Metadata V4L2 / payload error | HKR HIF request / done | HKR done-cadence FPS |
|---|---:|---:|---:|---:|---:|---:|
| Depth | 76,435 | 60.00 (59.90-60.60) | 0 / 0 | 0 / 0 | 86,045 / 86,045 | 60.46 |
| IR | 76,627 | 60.00 (59.70-60.10) | 0 / 0 | 0 / 0 | 78,219 / 78,219 | 60.00 |
| RGB | 80,269 | 60.00 (59.80-60.20) | 0 / 0 | 0 / 0 | 78,347 / 78,347 | 60.00 |

| Result | Count |
|---|---:|
| App pass / Host-clean cycles | 300/300 / 300/300 |
| VI corr_err / EMBED_RUNAWAY / PIXEL_RUNAWAY | 0 / 0 / 0 |
| VI timeout / recovery | 0 / 0 |
| Panic / Oops / Host reboot | 0 / 0 / 0 |

[...]

PR #570, opened 2026-08-10
[D585][GMSL] Configure pixel mode CSI lane rates to 1300M Serializer / 1400M Deserializer

polejoe:
## Summary

- Configure D585 pixel/MAX96712 mode for 1300 Mbps/lane from D585 to SER and 1400 Mbps/lane from DES to Host.
- Add opt-in MAX96712 CSI output-rate and continuous-clock properties.
- Keep the MAX96717 CSI input in continuous-clock mode and apply the 16-bpp override required after metadata 8-to-16 double-loading.

## D4xx isolation

- Only the D5x FG12 overlay opts into the new lane-rate and clock properties.
- D4xx MAX96712 overlays do not opt into the PR 570 properties and retain the PR 568 base defaults: 1300 Mbps/lane for four lanes and 2500 Mbps/lane for two lanes.
- D4xx uses MAX9295 rather than MAX96717. The MAX96717 16-bpp override remains inside the pixel-only branch; tunnel-mode reset values are unchanged.
- This PR has no `d4xx.c`, D4xx overlay, MAX9295, or MAX9296 diff relative to PR #567.

## Dependencies

Depends-On: https://github.com/realsenseai/realsense_mipi_platform_driver/pull/567

PR #567 is stacked on [PR #568](https://github.com/realsenseai/realsense_mipi_platform_driver/pull/568). This PR is based directly on the current PR #567 head.

## Validation

- MAX96712 JP 6.2.2 incremental patch dry-run: passed.
- JP 6.2.2 DTB/DTBO build, including the D5x FG12 overlay: passed.
- JP 6.2.2 MAX96717 module build with `-Werror`: passed.
- 300-cycle direct-V4L2 3VC test at 1300/1400: 101/300 clean cycles, 3038 `EMBED_RUNAWAY` events, and zero VI timeouts.

The lane-rate change removes the severe VI-timeout path but does not fully resolve the remaining metadata/frame-boundary issue.

### DIFF
--- a/hardware/realsense/tegra234-camera-d4xx-overlay-fg12-4ch-d5xx.dts
+++ b/hardware/realsense/tegra234-camera-d4xx-overlay-fg12-4ch-d5xx.dts
@@ -281,6 +281,8 @@
 							compatible = "maxim,max96712";
 							csi-mode = "2x4";
 							lane_cnt = <4>;
+							csi-lane-rate-mbps = <1400>;
+							force-csi-out-clock;
 							link-mask = <0x1>;
 							channel = "a";
 							fangzhu,fg12-reverse-lane-polarity;
@@ -302,6 +304,7 @@
 							compatible = "intel,d4xx";
 							use_sensor_mode_id = "true";
 							cam-type = "IMU";
+							mipi-lane-rate-mbps = <1300>;
 							nvidia,gmsl-ser-device = <&ser_a>;
 							nvidia,gmsl-dser-device = <&dser>;
 							ports {
@@ -325,9 +328,12 @@
 								active_w = "640";
 								active_h = "480";
 								tegra_sinterface = "serial_c";
+								discontinuous_clk = "no";
+								deskew_initial_enable = "false";
+								deskew_periodic_enable = "false";
 								mclk_khz = "24000";
 								pix_clk_hz = "74250000";
-								serdes_pix_clk_hz = "325000000";
+								serdes_pix_clk_hz = "350000000";
 								line_length = "1280"; /* 2200 */
 								embedded_metadata_height = "0";
 							};
@@ -350,6 +356,7 @@
 							compatible = "intel,d4xx";
 							use_sensor_mode_id = "true";
 							cam-type = "Y8";
+							mipi-lane-rate-mbps = <1300>;
 							nvidia,gmsl-ser-device = <&ser_a>;
 							nvidia,gmsl-dser-device = <&dser>;
 							ports {
@@ -374,9 +381,12 @@
 								active_w = "1280";
 								active_h = "720";
 								tegra_sinterface = "serial_c";
+								discontinuous_clk = "no";
+								deskew_initial_enable = "false";
+								deskew_periodic_enable = "false";
 								mclk_khz = "24000";
 								pix_clk_hz = "74250000";
-								serdes_pix_clk_hz = "325000000";
+								serdes_pix_clk_hz = "350000000";
 								line_length = "1280"; /* 2200 */
 								embedded_metadata_height = "1";
 							};
@@ -399,6 +409,7 @@
 							compatible = "intel,d4xx";
 							use_sensor_mode_id = "true";
 							cam-type = "RGB";
+							mipi-lane-rate-mbps = <1300>;
 							nvidia,gmsl-ser-device = <&ser_a>;
 							nvidia,gmsl-dser-device = <&dser>;
 							ports {
@@ -422,9 +433,12 @@
 								active_w = "1920";
 								active_h = "1080";
 								tegra_sinterface = "serial_c";
+								discontinuous_clk = "no";
+								deskew_initial_enable = "false";
+								deskew_periodic_enable = "false";
 								mclk_khz = "24000";
 								pix_clk_hz = "74250000";
-								serdes_pix_clk_hz = "325000000";
+								serdes_pix_clk_hz = "350000000";
 								line_length = "1280"; /* 2200 */
 								embedded_metadata_height = "1";
 							};
@@ -446,6 +460,7 @@
 							compatible = "intel,d4xx";
 							use_sensor_mode_id = "true";
 							cam-type = "Depth";
+							mipi-lane-rate-mbps = <1300>;
 							nvidia,gmsl-ser-device = <&ser_a>;
 							nvidia,gmsl-dser-device = <&dser>;
 							ports {
@@ -469,9 +484,12 @@
 								active_w = "1280";
 								active_h = "720";
 								tegra_sinterface = "serial_c";
+								discontinuous_clk = "no";
+								deskew_initial_enable = "false";
+								deskew_periodic_enable = "false";
 								mclk_khz = "24000";
 								pix_clk_hz = "74250000";
-								serdes_pix_clk_hz = "325000000";
+								serdes_pix_clk_hz = "350000000";
 								line_length = "1280"; /* 2200 */
 								embedded_metadata_height = "1";
 							};
--- a/nvidia-oot/6.2/0013-max96712-configurable-csi-lane-rate.patch
+++ b/nvidia-oot/6.2/0013-max96712-configurable-csi-lane-rate.patch
@@ -1,0 +1,101 @@
+diff --git a/drivers/media/i2c/max96712.c b/drivers/media/i2c/max96712.c
+index 88de8179..74848273 100644
+--- a/drivers/media/i2c/max96712.c
++++ b/drivers/media/i2c/max96712.c
+@@ -45,8 +45,12 @@
+ 
+ #define MAX96712_BACKTOP25_ADDR 0x418
+ #define MAX96712_BACKTOP25_PHY1_SW_OVRR_OFFS 0x5
+-#define MAX96712_BACKTOP25_2500Mbps 0x19
+-#define MAX96712_BACKTOP25_1300Mbps 0xd
++#define MAX96712_CSI_LANE_COUNT_DEFAULT 2
++#define MAX96712_CSI_LANE_RATE_2_LANE_DEFAULT_MBPS 2500
++#define MAX96712_CSI_LANE_RATE_4_LANE_DEFAULT_MBPS 1300
++#define MAX96712_CSI_LANE_RATE_MIN_MBPS 100
++#define MAX96712_CSI_LANE_RATE_MAX_MBPS 2500
++#define MAX96712_CSI_LANE_RATE_STEP_MBPS 100
+ 
+ #define MAX96712_MIPI_TX_EXT0_ADDR 0x800
+ #define MAX96712_MIPI_TX_EXT1_ADDR 0x801
+@@ -154,7 +158,9 @@ struct max96712 {
+ 	int link_multi_vc_pipe[MAX96712_MAX_LINKS];
+ 	bool multi_vc_configured[MAX96712_MAX_PIPES];
+ 	bool force_clk0;
++	bool force_csi_out_clock;
+ 	int lane_cnt;
++	u32 csi_lane_rate_mbps;
+ 	bool fangzhu_fg12_reverse_lane_polarity;
+ 	bool fangzhu_fg12_poc_enable;
+ };
+@@ -424,6 +430,10 @@ static int max96712_probe(struct i2c_client *client,
+ 		return -ENODEV;
+ 	}
+ 	mutex_init(&priv->lock);
++	/* Preserve legacy defaults when the device has no OF node. */
++	priv->lane_cnt = MAX96712_CSI_LANE_COUNT_DEFAULT;
++	priv->csi_lane_rate_mbps =
++		MAX96712_CSI_LANE_RATE_2_LANE_DEFAULT_MBPS;
+ 
+ 	np = client->dev.of_node;
+ 	if (np) {
+@@ -451,6 +461,8 @@ static int max96712_probe(struct i2c_client *client,
+ 		} else {
+ 			priv->force_clk0 = false;
+ 		}
++		priv->force_csi_out_clock =
++			of_property_read_bool(np, "force-csi-out-clock");
+ 
+ 		err = of_property_read_u32(np, "lane_cnt", &value);
+ 		if (err < 0) {
+@@ -465,6 +477,22 @@ static int max96712_probe(struct i2c_client *client,
+ 			priv->lane_cnt = value;
+ 		}
+ 
++		priv->csi_lane_rate_mbps = priv->lane_cnt == 4 ?
++			MAX96712_CSI_LANE_RATE_4_LANE_DEFAULT_MBPS :
++			MAX96712_CSI_LANE_RATE_2_LANE_DEFAULT_MBPS;
++		err = of_property_read_u32(np, "csi-lane-rate-mbps", &value);
++		if (!err) {
++			if (value < MAX96712_CSI_LANE_RATE_MIN_MBPS ||
++			    value > MAX96712_CSI_LANE_RATE_MAX_MBPS ||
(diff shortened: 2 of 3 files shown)


nvidia-oot/max96717.c
```diff
@@ -322,16 +326,13 @@ static int max96717_set_registers(struct device *dev, struct reg_pair *map,
...
 static int max96717_mipi_rx_reset_pulse(struct device *dev)
 {
-	struct max96717 *priv = dev_get_drvdata(dev);
-	u8 reset = priv->pixel_mode ? MAX96717_MIPI_RX0_RESET : 0x08;
-	u8 normal = priv->pixel_mode ? MAX96717_MIPI_RX0_NORMAL : 0x00;
 	int err;
 
 	err  = max96717_write_reg(dev, MAX96717_MIPI_RX0_ADDR,
-				  reset);
+				  MAX96717_MIPI_RX0_RESET);
```
ymodlin, line 332: ⚠️ **[Auto-generated: rs-agentic bot]** — Potential behavioral regression.

The old code used `priv->pixel_mode` to select register values:
- pixel_mode → `0x48`/`0x40` (bit6=1, non-continuous clock)
- tunnel_mode → `0x08`/`0x00` (bit6=0, continuous clock)

This PR unconditionally uses `0x08`/`0x00` (continuous clock) for **all callers**, including any non-D5xx pixel-mode devices. Confirm `max96717_mipi_rx_reset_pulse()` is only called in D5xx/continuous-clock contexts, or guard with a device check.

kernel/realsense/d4xx.c
```diff
@@ -6212,10 +6214,33 @@ static int ds5_hw_init(struct i2c_client *c, struct ds5 *state)
...
+	mipi_caps_valid = !ret;
 	if (!ret)
 		dev_dbg(sd->dev, "%s(): %d: %u lanes, phy %x, data rate %u-%u\n",
 			 __func__, __LINE__, n_lanes, phy, drate_min, drate_max);
 
+	/* D58x camera-to-serializer rate is a property of the selected
+	 * board/link overlay. D4xx and older DTS files stay at 1000 Mbps. */
+	lane_rate = D4XX_MIPI_LANE_RATE;
+	if (ds5_is_d58x(state) && sd->dev->of_node &&
+	    !of_property_read_u32(sd->dev->of_node,
```
ymodlin, line 6226: 💡 **[Auto-generated: rs-agentic bot]** — Edge case: when `mipi_caps_valid` is `false` (I2C read failures on boot), the DT-specified `lane_rate` is accepted and written to the device **without** FW range validation. If I2C is flaky at probe time this could configure an unsupported rate. Consider logging a warning when DT rate is used with `!mipi_caps_valid`.

nvidia-oot/6.2/0013-max96712-configurable-csi-lane-rate.patch
```diff
@@ -0,0 +1,94 @@
...
+ 		priv->fangzhu_fg12_poc_enable =
+@@ -913,13 +934,16 @@ int max96712_init_settings(struct device *dev)
+ 		{MAX96712_VIDEO_PIPE_EN_ADDR, 0x00}, //0xF4, Disable all pipes for now
+ 	};
+ 
+-	/* 1100Mbps per lane for 4 lanes, 2500Mbps per lane for 2 lanes */
+-	uint8_t lane_rate = (priv->lane_cnt == 4) ? MAX96712_BACKTOP25_1100Mbps : MAX96712_BACKTOP25_2500Mbps;
++	/* BACKTOP25 encodes the per-lane rate in 100 Mbps units. */
++	uint8_t lane_rate = priv->csi_lane_rate_mbps /
++		MAX96712_CSI_LANE_RATE_STEP_MBPS;
```
ymodlin, line 74: 🔧 **[Auto-generated: rs-agentic bot]** — Style: kernel code uses `u8`, not `uint8_t`. Change to:
```c
u8 lane_rate = priv->csi_lane_rate_mbps / MAX96712_CSI_LANE_RATE_STEP_MBPS;
```
(See `checkpatch.pl` rule: prefer kernel integer types.)

nvidia-oot/6.2/0013-max96712-configurable-csi-lane-rate.patch
```diff
@@ -0,0 +1,94 @@
...
++	u32 csi_lane_rate_mbps;
+ 	bool fangzhu_fg12_reverse_lane_polarity;
+ 	bool fangzhu_fg12_poc_enable;
+ };
+@@ -424,6 +426,8 @@ static int max96712_probe(struct i2c_client *client,
+ 		return -ENODEV;
+ 	}
+ 	mutex_init(&priv->lock);
++	priv->lane_cnt = 2;
++	priv->csi_lane_rate_mbps = 2500;
```
ymodlin, line 32: 💡 **[Auto-generated: rs-agentic bot]** — The defaults set here (`lane_cnt=2`, `csi_lane_rate_mbps=2500`) are immediately overridden inside the `if (np)` block below (`priv->csi_lane_rate_mbps = priv->lane_cnt == 4 ? 1100 : 2500`). The pre-`np` initialization is only a safety net if `of_node` is NULL. Consider a single initialization path or a comment explaining the two-phase default.

ymodlin (commented):
**rs-agentic bot automated review** — PR #570 `[D585][GMSL] Configure pixel CSI lane rates`

Author is an org member; full review applied.

**Overall:** The lane-rate refactor is well-structured. A few issues noted inline below — one potential behavioral regression, one style violation, and one edge-case worth hardening.

nvidia-oot/max96717.c
```diff
@@ -359,6 +360,8 @@ int max96717_init_settings(struct device *dev)
 		{MAX96717_VIDEO_TX0_ADDR, 0x60}, /* 0x110 - VIDEO_TX0 AUTO_BPP=0 ENC_MODE=10 */
 		{MAX96717_VIDEO_TX1_ADDR, 0x10}, /* 0x111 - VIDEO_TX1 BPP=16 forced (matches DT) */
 		{MAX96717_FRONTTOP_10_ADDR, 0x4}, /* 0x312 - Fronttop_10 double 8bit */
+		/* 0x31E - BPP override enable, 16bpp after double */
+		{MAX96717_FRONTTOP_22_ADDR, 0x30},
```
ashrafk93, line 364: **[rs-agentic]** ℹ️ `FRONTTOP_22` (0x31E, BPP-override-enable, 16bpp-after-double) is added unconditionally to `ser_cfg_pre[]`. Please confirm that `max96717_init_settings()` is only ever called on the pixel-mode path. If it can be reached in tunnel mode, this register write would misconfigure the serializer BPP encoding and likely corrupt frame data.

ashrafk93 (commented):
**[rs-agentic auto-review]** PR #570 reviewed.

ymodlin (changes requested):
Why do you change this from the DT?

ymodlin:
I dont understand what you are trying to change here, this PR is not clear and will not be merged has is.

nvidia-oot/max96717.c
ejgoldik, line 364: What does this have to do with the lane rate? Why set override enable when BPP=16 is already set in reg 0x111?

nvidia-oot/max96717.c
```diff
@@ -322,16 +326,13 @@ static int max96717_set_registers(struct device *dev, struct reg_pair *map,
...
-	u8 normal = priv->pixel_mode ? MAX96717_MIPI_RX0_NORMAL : 0x00;
 	int err;
 
 	err  = max96717_write_reg(dev, MAX96717_MIPI_RX0_ADDR,
-				  reset);
+				  MAX96717_MIPI_RX0_RESET);
 	usleep_range(2000, 2100);
 	err |= max96717_write_reg(dev, MAX96717_MIPI_RX0_ADDR,
-				  normal);
+				  MAX96717_MIPI_RX0_NORMAL);
```
ejgoldik, line 335: Why this change? Is it related to lane rates?

nvidia-oot/6.2/0013-max96712-configurable-csi-lane-rate.patch
```diff
@@ -0,0 +1,94 @@
...
+ 
+ 	np = client->dev.of_node;
+ 	if (np) {
+@@ -451,6 +456,8 @@ static int max96712_probe(struct i2c_client *client,
+ 		} else {
+ 			priv->force_clk0 = false;
+ 		}
++		priv->force_csi_out_clock =
```
ejgoldik, line 40: What is this change?

nvidia-oot/6.2/0013-max96712-configurable-csi-lane-rate.patch
ejgoldik, line 32: Default values should be using the defines, not magic numbers

ejgoldik (changes requested):

nvidia-oot/max96717.c
polejoe: This is not the lane-rate encoding itself; it is a companion pixel-pipe setting in the validated configuration. `0x111` sets the BPP used by the video encoding path, whereas `0x31E` enables the Fronttop per-pipe BPP software override. After `0x312` double-loads embedded 8-bit metadata into the 16-bit pixel pipe, `0x31E=0x30` keeps pipe Z explicitly at 16 bpp. The MAX96717 datasheet documents these as separate controls.

This write is now in `pixel_cfg_mid[]` and is executed only when `priv->pixel_mode` is true; tunnel mode is unchanged.

Reference: https://www.analog.com/media/en/technical-documentation/data-sheets/max96717.pdf

nvidia-oot/max96717.c
polejoe: This is not the numeric lane-rate programming; it aligns the D-PHY clock mode end to end. The MAX96717 datasheet supports both continuous and burst/non-continuous input clocking, but continuous is the reset default (`MIPI_RX0[6]=0`), and ADI's pixel-mode programming examples use `MIPI_RX0=0x00`. D585 already supplies a continuous clock, while the MAX96712 output and Orin DT are configured for the same mode, so the reset pulse must preserve bit 6 cleared.

This is not a deskew requirement at 1300 Mbps/lane; ADI only requires initial deskew at 1.5 Gbps/lane and above. It is clock-mode alignment. Tunnel mode already used these reset values, and D4xx uses the MAX9295 path rather than this MAX96717 driver.

References:
- https://www.analog.com/media/en/technical-documentation/data-sheets/max96717.pdf
- https://www.analog.com/media/en/technical-documentation/user-guides/max96717fr-user-guide.pdf

nvidia-oot/6.2/0013-max96712-configurable-csi-lane-rate.patch
polejoe: This adds an opt-in DT control for MAX96712 `MIPI_PHY0[7]` (`force_csi_out_en`), which keeps the deserializer CSI output clocks running. It makes the DES-to-Orin side explicitly match `discontinuous_clk = "no"` instead of relying on the normal-mode clock state.

This is clock-mode alignment, not part of the 1400-Mbps DPLL encoding itself. Only the D5x FG12 pixel-mode overlay enables the property; D4xx overlays omit it and retain the original register value.

nvidia-oot/6.2/0013-max96712-configurable-csi-lane-rate.patch
polejoe: Agreed. This was updated to use named constants for the default lane count and both legacy lane-rate defaults. Behavior is unchanged: two lanes default to 2500 Mbps/lane and four lanes to 1100 Mbps/lane; only the D5x FG12 pixel-mode overlay overrides the output to 1400 Mbps/lane.

nvidia-oot/6.2/0013-max96712-configurable-csi-lane-rate.patch
ymodlin: change approved

nvidia-oot/6.2/0013-max96712-configurable-csi-lane-rate.patch
ymodlin: change approved — defaults now use named constants, and the 4-lane default is back to 1300 Mbps (`MAX96712_CSI_LANE_RATE_4_LANE_DEFAULT_MBPS`), matching the old `BACKTOP25_1300Mbps` behaviour.

kernel/realsense/d4xx.c
ymodlin: No longer applicable here — `d4xx.c` is out of this PR's diff (3 files now). The `mipi_caps_valid` path belongs to the base PR; tracking it there.

nvidia-oot/max96717.c
ymodlin: Accepted on the D4xx point — max96717 is D5xx-only (D4xx uses MAX9295), and tunnel mode kept `0x08`/`0x00`, so only pixel mode changes.

One caveat though: this is unconditional, so it also changes the 3 *other* pixel-mode max96717 overlays that this PR does not touch — `fg12-16ch-d5xx`, `fg12-16ch-cams-0-1-d5xx-d4xx`, `seeed-d5xx`. Same for the new `FRONTTOP_22` write. The PR description's "only the D5x FG12 overlay opts in" holds for the DT properties but not for these two register changes. Worth a quick sanity stream on one 16ch board.

hardware/realsense/tegra234-camera-d4xx-overlay-fg12-4ch-d5xx.dts
```diff
@@ -281,6 +281,8 @@
 							compatible = "maxim,max96712";
 							csi-mode = "2x4";
 							lane_cnt = <4>;
+							csi-lane-rate-mbps = <1400>;
```
ymodlin, line 284: **Potential BUG — JP6.0 / JP6.1 get the DT rate bump without the driver support.**

`0013-max96712-configurable-csi-lane-rate.patch` exists only in `nvidia-oot/6.2/` (6.2.1/6.2.2 are dir symlinks to it). `6.0/` and `6.1/` have no `0013`, and their `0003` is the shared `6.0/0003` with the old `lane_cnt == 4 ? 1300 : 2500` logic.

But this overlay is enabled for **all** JP6.x by the shared `hardware/nvidia/t23x/nv-public/6.x/0001-overlay-enable-d4xx-camera-for-Orin-jp6.patch`. So on JP6.0/6.1:

- `csi-lane-rate-mbps = <1400>` and `force-csi-out-clock` are unknown properties → silently ignored, BACKTOP25 stays at 1300 Mbps;
- `serdes_pix_clk_hz` is now `350000000`, so NVCSI/CIL is clocked for 1400 Mbps.

That is the exact mismatch CLAUDE.md warns about ("a `BACKTOP25` rate bump must be mirrored into every max96712 4-lane overlay ... or NVCSI runs at the wrong CIL clock") — expect `DATA_LANE_ERR*_RXFIFO_FULL` / `PD_CRC_ERR` / VI request timeouts on 6.0/6.1. Nothing fails at build time, so it will only show up on a rig.

Either port `0013` to `6.0/` (6.1 symlinks) or regenerate `0003`, per the CLAUDE.md guidance that max96712 lives in `0003` across generations.

nvidia-oot/max96717.c
ymodlin: change approved — the `FRONTTOP_22` write moved into `pixel_cfg_mid[]`, which is applied only under `if (priv->pixel_mode)` (`max96717_init_settings()`, line ~409). Tunnel mode takes `tunnel_cfg[]` and is unaffected.

Note it does still apply to the other 3 pixel-mode overlays — raised in the `MIPI_RX0` thread.

PR #571, opened 2026-08-11
POC TrustZone blocked --[Jetson][NVCSI] test: probe per-VC DT override access

polejoe:
## Scope

Draft register-access proof of concept for the Orin NVCSI per-stream/per-VC Data Type override register. This branch is intentionally isolated from the NV12-over-UD30 product implementation and must not be merged as a production data path.

## What was tried

- Added a default-disabled debugfs probe for the documented NVCSI DT override register.
- Restricted access to idle CSI and serialized it with stream start/stop.
- Implemented read, optional write/readback, and immediate restore with range and register-value validation.
- Tested stream 2 / VC1 on Orin JP6.2.2.

## Result

- The register read returned `0xffffffff`.
- The CBB reported `FIREWALL_ERR` for Non-Secure CCPLEX access.
- The probe returned `-EACCES` before issuing any write.

## Remaining gaps

- The register is TrustZone protected, so Linux `ioremap` or `devmem` cannot provide the required access on the tested platform.
- NVCSI uses shadowed state; a supported load/commit and stream lifecycle sequence is not exposed by the public Linux camera API.
- The stock RCE/capture IVC ABI does not expose the per-VC NVCSI override control.
- A complete Option 3 implementation still needs both NVCSI Color Parser override and VI channel override, followed by byte-integrity and full regression validation.

NVIDIA must provide a supported secure/RCE control path before this approach can move beyond a hardware-access POC.

## Safety

- Disabled by default.
- Rejects access while CSI is active.
- Does not write when the initial read is inaccessible or invalid.
- This draft has no integration with the production NV12 and IR/RGB Calib route

### DIFF
--- a/nvidia-oot/6.2/0017-Add-NVCSI-DT-override-MMIO-POC.patch
+++ b/nvidia-oot/6.2/0017-Add-NVCSI-DT-override-MMIO-POC.patch
@@ -1,0 +1,254 @@
+From: zhou <shengyong.zhou@realsenseai.com>
+Date: Mon, 10 Aug 2026 19:47:57 +0800
+Subject: [PATCH] media: tegra: add NVCSI DT override MMIO POC
+
+Add a default-disabled debugfs probe for the Orin NVCSI per-stream,
+per-VC Data Type override register. The probe is restricted to idle CSI,
+rejects inaccessible or invalid reads, and restores the original value
+immediately after write/readback.
+
+This is a register-access POC only. It does not configure the streaming
+data path or resolve the TrustZone, shadow commit, and RCE lifecycle gaps.
+
+On Orin JP6.2.2, the stream 2 / VC1 register read returned 0xffffffff and
+the CBB reported FIREWALL_ERR for the Non-Secure CCPLEX access. The probe
+returned -EACCES before issuing a write, confirming that direct Host MMIO
+cannot implement the override on the tested platform.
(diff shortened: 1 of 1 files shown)


nvidia-oot/6.2/0017-Add-NVCSI-DT-override-MMIO-POC.patch
```diff
@@ -0,0 +1,254 @@
+From: zhou <shengyong.zhou@realsenseai.com>
```
ymodlin, line 1: `apply_patches.sh` applies `nvidia-oot/<ver>/*` with a glob (`git apply "${PWD}/$2/$1"/*`), so landing this file means every JP6.2 build ships a debugfs endpoint that pokes raw NVCSI MMIO — and only 6.2, leaving the patch numbering divergent from 6.0/6.1/7.x. Given the PR body already concludes host MMIO cannot implement the override (CBB `FIREWALL_ERR`, `-EACCES`), the findings look more valuable as a ticket/commit-message record than as a patch in the series.

ymodlin (commented):
Automated review of the POC. The code itself is careful — bounds-checked stream/vc/dt, idle-CSI gate under `source_update`, `nvhost_module_busy()` around the access, original value restored, and default-disabled behind a module param. One note on how it would ship.

PR #572, opened 2026-08-11
[Jetson][D5x][gmsl] feat: receive flat NV12 over UD30

polejoe:
## Depends on

- https://github.com/RealSenseAI-Internal/soc-rs-soc-io-drivers/pull/524
- https://github.com/RealSenseAI-Internal/soc-rs-soc-device-mgr/pull/457

## Summary

- Carry D5x RGB flat NV12 over CSI-2 UD30 while keeping the application ABI as standard `V4L2_PIX_FMT_NV12`.
- Select the camera route with source DT `0x18` and output override DT `0x30`.
- Receive UD30 as an opaque byte surface through an exact private media-bus tuple, one contiguous `Y + UV` allocation, and no post-DMA copy or reformat.
- Restrict FourCC lookup to formats enumerated by the active subdevice so unrelated camera channels cannot select RealSense private tuples.
- Keep visible transport width separate from aligned V4L2 surface stride.

## Host layering

For a visible `W x H` frame, HKR sends `H * 3 / 2` UD30 rows with `WC = W`. The D5xx subdevice owns the private NV12/UD30 media-bus tuple and logical geometry. The declarative opaque-surface entry selects USER1, applies the internal RAW8 descriptor override, and keeps VI `frame_x = W`.

V4L2 separately exposes `bytesperline S = ALIGN(W, 64)` and `sizeimage = S * H * 3 / 2`. VI writes each visible row into an aligned destination stride; the application dequeues standard contiguous NV12 without a post-DMA copy.

The common VI code contains no camera name, stream type, VC/MC, or DTS product branch.

## Isolation

- Standard NV12 tuples from non-RealSense cameras do not match the private media-bus code and remain on the stock path.
- An unsupported forced FourCC request cannot fall back to a private tuple that the active subdevice did not enumerate.
- D4xx streams do not enumerate the D5xx private media-bus code.
- `frame_x` is not overloaded with aligned stride; visible geometry and memory pitch remain separate contracts.

## Rollout compatibility

The HKR changes can merge before this host change. Existing hosts keep their current Depth and IR wire DTs. HKR resolves an input surface from `(stream, CSI output DT)` when that route is unique, so legacy control values do not require dedicated alias rows. Source DT is used only to distinguish multiple surfaces sharing one route, currently RGB NV12 and Bayer10 over UD30.

An old host does not request RGB NV12 over UD30, so the new RGB path remains inactive until this PR is installed.

## Verification

- PR branch contains one commit relative to `d500_gmsl_dev`.
- Full patch series apply on NVIDIA source tags for JP6.0, 6.1, 6.2.1, 6.2.2, 7.0, and 7.1: PASS.
- JP6.2.2 full kernel, modules, D5xx FG24 overlay, package deployment, reboot, and rootfs integrity gates: PASS.
- D5xx GMSL E2E regression on MAX96724 tunnel mode: 19/19 PASS.
- RGB NV12 1280x960@60: PASS.
- RGB NV12 848x480@60: 182 frames, zero drops; `WC=848`, `bytesperline=896`, `sizeimage=645120`.
- RGB NV12 480x270@60: 182 frames, zero drops; `WC=480`, `bytesperline=512`, `sizeimage=207360`.
- RGB NV12 424x240@60: 181 frames, zero drops; `WC=424`, `bytesperline=448`, `sizeimage=161280`.
- Depth, Y8I, YUYV, IMU, Y12I, Bayer10, and three/four-stream cases: PASS.
- SerDes regression: 71 PASS, 15 non-critical warnings, 0 critical errors.

## Explicit non-scope

This PR contains no NVCSI direct-MMIO or DT0x18 override POC. That Option 3 experiment remains isolated in draft PR #571 and is blocked by TrustZone firewall access plus missing shadow/RCE control APIs.

MAX96712 pixel mode is also outside this E2E result. CSI-2 RAW8/UD30 has no 64-byte WC rule, but MAX96712 pixel mode is format/BPP aware and requires paired serializer/deserializer Mode 1 configuration before arbitrary visible WC can be claimed. Current validation is MAX96724 tunnel mode only.


pinquanw (approved):

nvidia-oot/6.2/0014-Add-opaque-byte-surfaces.patch
```diff
@@ -0,0 +1,165 @@
...
++	u8 bpl_width_divisor;
++	u8 height_numerator;
++	u8 height_denominator;
++};
++
++static const struct tegra_vi_surface_override tegra_vi_surface_overrides[] = {
++	{
++		.code = MEDIA_BUS_FMT_RS_NV12_UD30_1X8,
++		.dt_override = TEGRA_IMAGE_DT_RAW8,
++		.bpl_width_divisor = 0,
```
ghu3, line 132: Blocking: bpl_width_divisor=0 leaves the VI descriptor at frame_x = W, while the firmware NV12-to-UD30 DPI byte path emits WC = getStride(FMT_NV12, W), which is 64-byte aligned. For advertised widths 848, 480, and 424, the wire rows are 896, 512, and 448 bytes respectively; the validated 1280 mode hides this mismatch. This can cause packet-width errors, truncated rows, or an undersized userspace buffer. Please make bytesperline, frame_x, and sizeimage consistently use the exact source stride S, and add runtime coverage for a non-64-aligned mode such as 848x480 or 424x240.

kernel/realsense/d5xx.c
```diff
@@ -1604,22 +1622,25 @@ static const struct d5x_format d5x_y_formats_d58x_raw16[] = {
...
 		.n_resolutions = ARRAY_SIZE(d58x_rgb_sizes),
 		.resolutions = d58x_rgb_sizes,
 	}, {
-		/* Flat NV12 surface carried as width x (height * 3 / 2) RAW8. */
-		.data_type = GMSL_CSI_DT_RAW_8,
-		.mbus_code = MEDIA_BUS_FMT_UYYVYY8_0_5X24,
+		/* Flat NV12 bytes use an opaque UD30 carrier. */
+		.source_dt = GMSL_CSI_DT_YUV420_8,
+		.csi_out_dt = GMSL_CSI_DT_USER_1,
+		.mbus_code = MEDIA_BUS_FMT_RS_NV12_UD30_1X8,
```
ghu3, line 1634: Blocking: this shared d5xx.c change advertises the private 0x5008 tuple and programs UD30 for every build, but only the nvidia-oot/6.2 patch set defines the matching VI format and opaque-surface descriptor. The 6.0, 6.1, 7.0, and 7.1 patch sets still install the old UYYVYY8_0_5X24 / RAW8 receiver entry. On those versions fmts_bitmap cannot match this tuple, or VI remains configured for RAW8 while the SerDes sends UD30, so NV12 regresses at runtime even though compile-only CI passes. Please propagate the 0014-0016 changes to every supported stack, or gate the tuple and wire-DT change by target version.

lemon6630 (approved):

PR #574, opened 2026-08-11
D585 gmsl dfu seeed d5xx dt

remibettan:
Same as PR #564 but with the field /sys/bus/i2c/devices/<adapter>-<addr>/of_node/compatible returning _realsense, d5xx_ instead of _intel, d4xx_

### DIFF
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -139,6 +139,10 @@ struct ser_interface {
 #define DS5_DEVICE_TYPE_D45X		6
 #define DS5_DEVICE_TYPE_D43X		5
 #define DS5_DEVICE_TYPE_UNKNOWN		0
+/* FW version major byte identifies the family in recovery, where DEVICE_TYPE
+ * (0x0310) is not served: D58x reports 7 or 8, D4xx reports 5.
+ */
+#define DS5_D58X_FW_MAJOR_MIN		7
 
 /* GVD response payload size per product line */
 #define DS5_GVD_LEN_D4XX		276
@@ -284,6 +288,11 @@ enum ds5_mux_pad {
 #define PIPE_NOT_CONFIGURED	-1
 
 #define DFU_WAIT_RET_LEN 6
+
+/* Deadline for the D58x manifest wait; the manifest can legitimately take
+ * up to ~90 s, so the bound is generous and only guards against a device
+ * stuck in an intermediate manifest state. */
+#define DFU_MANIFEST_TIMEOUT_MS 180000
 
 #define DS5_START_POLL_TIME	10
 #define DS5_START_MAX_TIME	2000
@@ -6622,6 +6631,8 @@ static int ds5_dfu_wait_for_get_dfu_status(struct ds5 *state,
 	u16 status, dfu_state_len = 0x0000;
 	unsigned char dfu_asw_buf[DFU_WAIT_RET_LEN];
 	unsigned int dfu_wr_wait_msec = 0;
+	unsigned long manifest_deadline = jiffies +
+			msecs_to_jiffies(DFU_MANIFEST_TIMEOUT_MS);
 
 	do {
 		ds5_write_with_check(state, 0x5008, 0x0003); // Get Write state
@@ -6653,7 +6664,16 @@ static int ds5_dfu_wait_for_get_dfu_status(struct ds5 *state,
 		dfu_wr_wait_msec = (((unsigned int)dfu_asw_buf[3]) << 16)
 						| (((unsigned int)dfu_asw_buf[2]) << 8)
 						| dfu_asw_buf[1];
-	} while (dfu_asw_buf[4] == dfuDNBUSY && exp_state == dfuDNLOAD_IDLE);
+	/* D58x manifestation runs dfuMANIFEST_SYNC -> dfuMANIFEST ->
+	 * dfuMANIFEST_WAIT_RESET; keep polling through the intermediate
+	 * states until the terminal wait-reset state is reached, bounded by
+	 * manifest_deadline so a stuck device errors out instead of hanging. */
+	} while ((dfu_asw_buf[4] == dfuDNBUSY &&
+		  exp_state == dfuDNLOAD_IDLE) ||
+		 ((dfu_asw_buf[4] == dfuMANIFEST_SYNC ||
+		   dfu_asw_buf[4] == dfuMANIFEST) &&
+		  exp_state == dfuMANIFEST_WAIT_RESET &&
+		  time_before(jiffies, manifest_deadline)));
 
 	if (dfu_asw_buf[4] != exp_state) {
 		dev_notice(&state->client->dev,
@@ -6701,6 +6721,16 @@ static int ds5_dfu_detach(struct ds5 *state)
 			__func__, buf.FW_lastVersion);
 	dev_notice(&state->client->dev, "%s():FW status (%s)\n",
 			__func__, buf.DFU_isLocked ? "locked" : "unlocked");
+	/* Recovery never ran ds5_wait_device_type(), so cached_device_type is
+	 * UNKNOWN; recover the family from the FW major byte so the DFU
+	 * manifest wait picks the D58x terminal state (dfuMANIFEST_WAIT_RESET).
+	 */
+	if (!ret && state->ds5_dev &&
+	    !ds5_is_valid_device_type(
+		    READ_ONCE(state->ds5_dev->cached_device_type)) &&
+	    ((buf.FW_lastVersion >> 24) & 0xff) >= DS5_D58X_FW_MAJOR_MIN)
+		WRITE_ONCE(state->ds5_dev->cached_device_type,
+			   DS5_DEVICE_TYPE_D58X);
 	return ret;
 };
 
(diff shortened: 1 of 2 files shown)


ymodlin:
**[rs-agentic] automated review**

The intended change — switching the four camera nodes in `tegra234-camera-d4xx-overlay-seeed-d5xx.dts` from `intel,d4xx` to `realsense,d5xx` — looks correct: `d4xx_of_match[]` carries both compatibles, and this brings the overlay in line with the other five `*-d5xx.dts` overlays.

**Scope issue:** the PR also carries the full `kernel/realsense/d4xx.c` diff from #564 (+295/-11). The blob hashes are identical (`f5f2dcf7..bcc86303`), so this is #564's DFU manifest/re-probe work, not a change of this PR. Both target `dev`.

Suggest rebasing onto `dev` once #564 lands so this PR is just the 4-line DTS change — otherwise the DFU work gets reviewed and merged twice. No issues found in the DTS change itself.

kernel/realsense/d4xx.c
```diff
@@ -6701,6 +6774,16 @@ static int ds5_dfu_detach(struct ds5 *state)
...
 			__func__, buf.DFU_isLocked ? "locked" : "unlocked");
+	/* Recovery never ran ds5_wait_device_type(), so cached_device_type is
+	 * UNKNOWN; recover the family from the FW major byte so the DFU
+	 * manifest wait picks the D58x terminal state (dfuMANIFEST_WAIT_RESET).
+	 */
+	if (!ret && state->ds5_dev &&
+	    !ds5_is_valid_device_type(
+		    READ_ONCE(state->ds5_dev->cached_device_type)) &&
+	    ((buf.FW_lastVersion >> 24) & 0xff) >= DS5_D58X_FW_MAJOR_MIN)
+		WRITE_ONCE(state->ds5_dev->cached_device_type,
```
ymodlin, line 6785: **BUG** — this stores a *guessed* family into `cached_device_type`, which is the same field `ds5_is_valid_device_type()` and the post-reset readiness path use as the operational-readiness gate (CLAUDE.md: "use `DS5_DEVICE_TYPE` validity as the operational-readiness gate" / "clear cached values for firmware-populated readiness registers before polling readiness").

Consequences:
- The guess is never cleared if DFU fails, so a mis-detect permanently mis-types the camera for the module's lifetime — `ds5_is_d58x()` gates RGB/control register offsets, so this is not confined to DFU.
- It makes `ds5_wait_device_type()` short-circuit on a value FW never reported.

Suggest a dedicated hint field (e.g. `state->dfu_dev.d58x_hint`) consulted only by the manifest-state selection, leaving `cached_device_type` FW-authoritative.

Also please confirm the byte order: nothing else in the driver decodes `FW_lastVersion`, so `>> 24` being the *major* (and not the build) is unverified.

Nit: the comment block above is 4 lines; CLAUDE.md caps in-code comments at 2-3.

kernel/realsense/d4xx.c
```diff
@@ -6653,7 +6717,16 @@ static int ds5_dfu_wait_for_get_dfu_status(struct ds5 *state,
...
 						| (((unsigned int)dfu_asw_buf[2]) << 8)
 						| dfu_asw_buf[1];
-	} while (dfu_asw_buf[4] == dfuDNBUSY && exp_state == dfuDNLOAD_IDLE);
+	/* D58x manifestation runs dfuMANIFEST_SYNC -> dfuMANIFEST ->
+	 * dfuMANIFEST_WAIT_RESET; keep polling through the intermediate
+	 * states until the terminal wait-reset state is reached, bounded by
+	 * manifest_deadline so a stuck device errors out instead of hanging. */
+	} while ((dfu_asw_buf[4] == dfuDNBUSY &&
+		  exp_state == dfuDNLOAD_IDLE) ||
+		 ((dfu_asw_buf[4] == dfuMANIFEST_SYNC ||
```
ymodlin, line 6726: **BUG** — the manifest wait has no guaranteed delay and no I2C fault tolerance:

1. The inner loop only sleeps when the device reports a non-zero `dfu_wr_wait_msec`. If a D58x reports 0 during manifest, this spins back-to-back I2C transactions for the full 180 s.
2. The loop body uses `ds5_write_with_check` / `ds5_read_with_check`, which bail out on the first I2C error. The D58x is rebooting during exactly this window, so a single transient NAK aborts the wait instead of retrying. CLAUDE.md asks for `ds5_read_poll()` in this class of loop ("polling loops expecting transient I2C failures ... DFU timeout checks").
3. On deadline expiry the loop falls into the `!= exp_state` branch, so a timeout is reported as `"Wrong dfu_state"` — please log the timeout distinctly, 180 s of silence followed by a misleading message is hard to triage in the field.

Nit: the comment block is 4 lines; CLAUDE.md caps in-code comments at 2-3.

kernel/realsense/d4xx.c
```diff
@@ -7015,24 +7110,169 @@ static int ds5_v4l_init(struct i2c_client *c, struct ds5 *state)
...
+		 "re-probing %u D585 instances after DFU reboot\n",
+		 ctx->client_count);
+	/* Peers share the primary's SerDes slot, so release the actual primary
+	 * last and recreate it before attaching any peers.
+	 */
+	for (i = 0; i < ctx->client_count; i++) {
+		if (i != ctx->primary_index)
+			device_release_driver(&ctx->clients[i]->dev);
+	}
+	device_release_driver(&primary->dev);
```
ymodlin, line 7163: **BUG (use-after-free)** — `device_release_driver()` on the depth instance runs `ds5_remove()` -> `ds5_chrdev_remove()`, and that helper calls `unregister_chrdev_region()` + `device_destroy()` but **never `cdev_del()`**. The `struct cdev` lives inside `struct ds5` (devm-allocated), so after unbind the kernel still holds a cdev pointing into freed memory, and `ds5_dfu_device_release()`/`_write()` recover `state` via `container_of(inode->i_cdev, struct ds5, dfu_dev.ds5_cdev)`.

Until now unbind only happened on manual unbind / module unload. This PR makes it an automatic runtime event 10 s after every recovery DFU, so any tool that still holds or re-opens `/dev/d4xx-dfu*` in that window hits the freed object.

Suggest adding `cdev_del()` to `ds5_chrdev_remove()`, and rejecting `ds5_dfu_device_open()` while a re-probe is pending.

kernel/realsense/d4xx.c
```diff
@@ -4312,12 +4326,60 @@ static void ds5_init_ds5_dev(struct ds5 *state, struct ds5_dev *ds5_dev)
...
+	ds5_dev->client_count = 0;
+#endif
 	ds5_dev->serdes_setup_complete = false;
 	ds5_dev->cached_device_type = DS5_DEVICE_TYPE_UNKNOWN;
 	mutex_unlock(&ds5_dev->lock);
 	ds5_reset_streaming_flags(ds5_dev);
 }
 
+#ifdef CONFIG_VIDEO_D4XX_SERDES
+static int ds5_track_client(struct ds5 *state)
```
ymodlin, line 4340: `ds5_track_client()` dereferences `state->ds5_dev` unchecked while `ds5_untrack_client()` right below null-checks it. Please make them symmetric — the call site in `ds5_probe()` sits after `ds5_serdes_setup()`, and it is not obvious every success path there leaves `ds5_dev` set.

kernel/realsense/d4xx.c
```diff
@@ -7015,24 +7110,169 @@ static int ds5_v4l_init(struct i2c_client *c, struct ds5 *state)
...
+		if (!ctx && !ds5_dfu_reprobe[i].pending)
+			ctx = &ds5_dfu_reprobe[i];
+	}
+	if (!ctx) {
+		ret = -ENOSPC;
+		goto out;
+	}
+
+	mutex_lock(&ds5_dev->lock);
+	if (ds5_dev->client_count != ARRAY_SIZE(ctx->clients) ||
```
ymodlin, line 7213: **Enhancement** — requiring `client_count == ARRAY_SIZE(ctx->clients)` (exactly 4) means the auto re-probe silently degrades to a `dev_warn` on any overlay that instantiates fewer than 4 d4xx child nodes, or if one sibling failed to probe. `>= 1` with `client_count` as the loop bound would be more robust and still correct.

Also the `[4]` array size is hardcoded in both `struct ds5_dev` and `struct ds5_dfu_reprobe_ctx` — please use a shared named constant so the two cannot drift.

Minor: the `primary_index`/`depth_index` sentinel is `ARRAY_SIZE(ctx->clients)` but is later compared against `ctx->client_count`. Correct only because of the `== 4` guard above; comparing against the sentinel directly would be safer.

ymodlin (commented):
Auto-generated review by the **rs-agentic bot**.

The DFU manifest state-machine fix and the `device_open_count` leak fix on the `dfu_msg` alloc failure are both good. Main concerns are in the new automatic re-probe path: tearing the driver down at runtime makes a latent chardev lifetime bug reachable, and the D58x family guess is stored in the field the readiness logic gates on.

Details inline.

kernel/realsense/d4xx.c
```diff
@@ -7015,24 +7110,169 @@ static int ds5_v4l_init(struct i2c_client *c, struct ds5 *state)
...
 #endif
+	bool auto_reprobe = false;
 	int ret = 0, retry = 10;
 	mutex_lock(&state->lock);
 	state->dfu_dev.device_open_count--;
+#ifdef CONFIG_VIDEO_D4XX_SERDES
+	auto_reprobe = state->dfu_dev.dfu_state_flag == DS5_DFU_DONE &&
+		state->dfu_dev.reprobe_after_dfu && ds5_is_d58x(state);
+	if (auto_reprobe)
+		ds5_dfu_schedule_reprobe(state);
```
ymodlin, line 7271: **BUG** — the return value of `ds5_dfu_schedule_reprobe()` is discarded, but `auto_reprobe` is left `true`. On any scheduling failure (`-ENOSPC` when all slots are pending, `-ENODEV` when `client_count != 4`, `-EBUSY`) the function still takes the early `return 0` below, skipping the "Verify communication" re-read.

Result: the DFU tool sees success, no re-probe ever runs, and `state->fw_version` / `fw_build` are stale — the exact case the caller needs to know about. Please assign the result to `auto_reprobe` so the fall-through path runs when scheduling failed.

(Same issue was raised on #564, which carries this identical commit.)

ymodlin (commented):
**Duplicate-PR warning (rs-agentic bot)** — this PR contains PR #564's two commits verbatim (`05ccddb6`, `32432918`) plus one DT commit, and both PRs target `dev`.

Please either rebase this PR onto #564's branch and keep only `5f848a51`, or close #564. As-is, the same d4xx.c changes will be reviewed and merged twice. The DFU findings I raised here apply equally to #564; my earlier comments there are already resolved against the same lines, so please keep the two threads in sync.

PR #275, opened 2025-06-13
d4xx : Linux v4l2 compat 5.15 to 6.11

Pirouf:
* Support Linux v4l2 from 5.15 to 6.11

PR #576, opened 2026-08-12
[RSDEV-13503][RSDEV-13505][VC4_dual_rgb][D585][gmsl] feat: receive DUALRGB_RIGHT on VC4

polejoe:
## Status

Draft reference implementation for the validated VC1 + VC4 dual-RGB topology. The product architecture is now evaluating a single-VC1 RGB multiplexing model; this PR is kept as the proven VC4 host-channel implementation and as input for future VC5-VC7 work.

## Summary

- register `DUALRGB_RIGHT` as an independent D5x VC4 video/metadata channel
- gate RGB-right admission and the 2C 30 FPS image-profile limit by the D585 product identity
- expose the existing RGB NV12/YUYV profiles through the additional channel
- share one MAX96724 physical tunnel pipe across logical VCs from the same link and roll back each acquired reference on setup/start failure
- keep one FG24 overlay source and declare VC4 as a runtime-probed capability
- initialize each stream configuration cache to an invalid sentinel so the first S_FMT writes an explicit zero output-DT override when required
- provide the full-size D500 GVD control and bound the firmware response copy
- complete metadata before publishing the paired VI video buffer

## Design

Tunnel mode forwards complete CSI packets from one physical GMSL link. VC0-VC4 share one DES tunnel pipe with reference counting; MAX96717/MAX96724 do not perform a per-VC pixel-format conversion in this path.

VC4 is a D5x sensor/channel capability. Existing VC0-VC3 formats, nodes, and control contracts are unchanged. A firmware that does not answer the VC4 probe does not register the additional subdevice.

The cache initialization fix is generic per-stream control behavior: cache values start invalid so the first requested value, including zero, is always sent to firmware.

## Verification

- JP5.0.2 through JP7.1 GitHub build matrix: PASS
- JP6.2.2 D5x FG24 build and deployment: PASS
- single RGB-right NV12 and YUYV `1280x960@30`: PASS
- dual RGB NV12 and YUYV: PASS
- Depth + IR + dual RGB: PASS
- five-stream/five-VC run including IMU: PASS
- independent RGB-right gate-on/gate-off: PASS
- metadata sidecars valid in the covered regression cases
- 2C 60 FPS requests are rejected by the product FPS guard as designed

## Upstream Scope

- D5x stream identity and VC4 topology stay in the D5x sensor/DT integration
- tunnel-pipe reference counting follows generic MAX96724 tunnel semantics
- pixel-mode behavior is unchanged
- config-cache initialization and VI metadata completion ordering are generic correctness fixes

## Related

Related HKR PRs: #525, #458, #275, #359

Parent Story: https://rsjira.realsenseai.com/browse/RSDEV-13489

Jira:

- Tunnel mode: https://rsjira.realsenseai.com/browse/RSDEV-13503
- Host VC4 channel: https://rsjira.realsenseai.com/browse/RSDEV-13505


polejoe:
**Deep review — VC4 host channel, tunnel pipe sharing, VI5 metadata ordering**

One must-fix, several flags:

1. 🔴 **Tunnel pipe refcount leak on error paths (must fix).** In `d5x_configure()`, after `get_available_pipe_id()` succeeds (which now does `st_count++` in tunnel mode), both the `d5x_setup_pipeline()` failure and the DES re-arm re-apply failure paths `return ret` without calling `release_pipe()`. Under the old semantics (`st_count = 0` on release) a leaked count was self-healing; under the new refcount semantics the count is stuck >0 forever, so the tunnel pipe is never truly released after the last stream stops. This is exactly the "refcount symmetric across start, stop and error paths" checklist item — currently not met. Suggested: `goto` cleanup that releases the just-acquired pipe on any post-acquire failure, plus an error-injection test (fail `setup_pipeline` → assert `st_count` returns to 0).
2. 🟡 **Pixel mode + VC4 gives a misleading error.** The pixel-mode branch loops `i < MAX96724_MAX_PIPES (4)` and can never match `vc_id == 4`, so RGB2 in pixel mode fails with `-ENOMEM "No free pipe"`. Recommend an explicit `vc_id >= MAX96724_MAX_PIPES → -EOPNOTSUPP` with a "VC4+ requires tunnel mode" log.
3. 🟡 **`is_rgb && is_rgb2` double-flag discipline.** `strncmp(str, "RGB", 3)` matches "RGB2", so RGB2 nodes have both flags set, and every dispatch site must test `is_rgb2` before `is_rgb` (the diff does this consistently). Any future `if (state->is_rgb)` added elsewhere in this ~8k-line file will silently treat RGB2 as RGB1. Suggest a `d5x_is_rgb1()` helper (`is_rgb && !is_rgb2`) or at minimum a comment on the struct fields stating the ordering rule.
4. 🟡 **Deployment mismatch hazards (document, not code):**
   - Old host driver + new 5ch dual-rgb overlay: the old driver's `strncmp("RGB",3)` also matches `cam-type = "RGB2"` → d5xx@1e is treated as a second RGB1 (writes RGB1 registers, sends `START|RGB`). The 5ch overlay must only ship with this driver or newer — please state the minimum driver/FW pairing in the README.
   - New driver + 5ch overlay + old FW: RGB2 probe reads `0x407E`/`0x4806`, which old FW doesn't serve. Consider a capability probe (FW version gate or a probe-read of the 0x4060 region) before enabling the RGB2 node, or document the minimum FW.
5. 🟡 **VI5 metadata ordering patch is a global VI5 behavior change** — d4xx devices share `vi5_fops.c`. The fix direction is right (complete metadata before `vb2_buffer_done()` on the video buffer, so a userspace requeue can't reset the sequence the metadata path still needs). Please run a d4xx regression as well, and consider whether the patch belongs in the shared 6.0 series per the repo's patch-carrier conventions rather than 6.2-only.
6. 🟡 **5ch overlay shifts all /dev/videoN indices** (VI registers in reverse DTS port order; new port@4 carries metadata). Numeric-index scripts will break; udev aliases stay stable. Worth a README note.
7. 🟢 RGB2 DTS `mode0` is YUY2 1920x1080 while the FW default/E2E profile is NV12 1280x960@60 — relies on dynamic mode update like RGB1; please confirm `serdes_pix_clk_hz = 500 MHz` covers the NV12@60 case.
8. Backward-compat note (for the record): without the 5ch overlay this driver is inert w.r.t. RGB2 — no d5xx@1e node, no new register access — so old-FW + new-driver + 4ch overlay is safe. The HKR-side PRs are likewise safe against this driver's absence (verified in the io-drivers #525 review).

nvidia-oot/files/6.2/drivers/media/i2c/max96724.c
```diff
@@ -1647,7 +1680,15 @@ int max96724_release_pipe(struct device *dev, int pipe_id)
 		return -EINVAL;
 
 	mutex_lock(&priv->lock);
-	priv->pipe[pipe_id].st_count = 0;
+	if (!priv->pipe[pipe_id].st_count) {
+		mutex_unlock(&priv->lock);
+		return -EINVAL;
+	}
+	priv->pipe[pipe_id].st_count--;
```
ymodlin, line 1687: **Potential BUG:** `max96724_release_pipe()` was idempotent (`st_count = 0`); it now refcounts and returns `-EINVAL` when the pipe is already free.

Two consequences worth checking:
- any teardown/error path that releases twice now returns an error instead of being a no-op;
- in tunnel mode all VCs share `PIPE_X`, so a single missed release (e.g. a stream-start that fails after `get_available_pipe_id()`) leaves `st_count > 0` and the tunnel pipe armed for the driver's lifetime.

kernel/realsense/d5xx.c
```diff
@@ -2421,7 +2472,8 @@ static int d5x_configure(struct d5x *state)
 			sensor->pipe_id, pipe_reapply_gen, data_type1, data_type2, vc_id);
 	}
 #else /* Non-SERDES configuration */
-	vc_id = (state->is_depth) ? 0 : (state->is_rgb) ? 1 : (state->is_y8) ? 2 : 3;
+	vc_id = (state->is_depth) ? 0 : (state->is_rgb2) ? 4 :
```
ymodlin: **Potential BUG (latent):** two different sources of truth for the VC in the same function. The SerDes pipe is allocated from `state->g_ctx.dst_vc` (DT `vc-id`, line ~2386), but this line re-derives `vc_id` from `cam-type` and hardcodes `4` for RGB2, and that value is what feeds `md_value` below.

They agree only because the new overlay sets `vc-id = <4>`. If a future overlay picks a different VC for RGB2 the deserializer routing and the FW metadata VC diverge silently. Suggest deriving from `state->g_ctx.dst_vc` instead of hardcoding.

build_all.sh
```diff
@@ -165,14 +165,25 @@ else
...
+                sed -i '/^dtbo-y += tegra234-camera-d5xx-overlay-fg24-4ch.dtbo$/a dtbo-y += tegra234-camera-d5xx-overlay-fg24-5ch-dual-rgb.dtbo' "$JP6_OVERLAY_MAKEFILE"
+            fi
         fi
         make RS_USE_D5XX=$RS_USE_D5XX RS_CSI_LANES=$RS_CSI_LANES dtbs
         cp $BUILD_SRCS/nvidia-oot/device-tree/platform/generic-dts/dtbs/tegra234-camera-d4xx-overlay*.dtbo $TEGRA_KERNEL_OUT/rootfs/boot/ 2>/dev/null || true
         if [[ "$D5XX" == "1" ]]; then
             cp $BUILD_SRCS/nvidia-oot/device-tree/platform/generic-dts/dtbs/tegra234-camera-d5xx-overlay*.dtbo $TEGRA_KERNEL_OUT/rootfs/boot/ 2>/dev/null || true
         fi
         if [[ "$FG24" == "1" ]]; then
+            test -f "$BUILD_SRCS/nvidia-oot/device-tree/platform/generic-dts/dtbs/tegra234-camera-d5xx-overlay-fg24-5ch-dual-rgb.dtbo"
```
ymodlin: **Potential BUG:** bare `test -f` under `set -e` (line 3) — if `tegra234-camera-d5xx-overlay-fg24-5ch-dual-rgb.dtbo` is not produced, `build_all.sh` aborts right here with no message, while every neighbouring `cp` is deliberately tolerant (`2>/dev/null || true`). Reads like leftover debug scaffolding: either drop the line, or make it an explicit failure with a message (`|| { echo "missing dtbo"; exit 1; }`).

hardware/realsense/tegra234-camera-d5xx-overlay-fg24-4ch.dts
```diff
@@ -500,6 +787,243 @@
...
+						 * cam_i2c bus before relying on this overlay.
+						 */
+						d5xx_per1: d5xx@1f {
+							status = "okay";
+							def-addr = <0x10>;
+							reg = <0x1f>;
+							override_reg = <0x1a>;
+							compatible = "realsense,d5xx";
+							use_sensor_mode_id = "true";
+							cam-type = "PER1";
```
ymodlin: **BUG** — `cam-type = "PER1"` (and `PER2`/`PER3`) matches no branch in the `cam-type` parser (`d5xx.c` ~L7768), so `is_depth/is_y8/is_rgb/is_rgb2/is_imu` all stay 0. Consequences:

1. `control_base` falls through to the fallback at ~L7817 → **`D5X_DEPTH_CONTROL_BASE`**. Controls are registered unconditionally for every client (`d5x_ctrl_init()`, L6725-6742) and the ctrl paths use `state->control_base` (L3502 / L3976), so a control write on a PER node lands in the **depth** control block — three extra clients able to disturb the real depth stream.
2. `d5x_set_state_last_set()` else-branch aliases `mux.last_set` to `imu.sensor`; `d5x_get_sensor_name()` reports `"unknown"`.
3. `d5x_state_to_pad()` returns `-1` into a `u32 pad`, which lands on `default: return -EINVAL`. So the "stays registered but enumerates no formats" behaviour the header promises is an accident of the `-1` → `u32` wrap, not a designed path.

Suggest an explicit `PER1..3` case (or a declared "no control block" stream class that skips ctrl registration) before this overlay ships.

build_all.sh
```diff
@@ -165,14 +165,32 @@ else
...
+            test -f "$BUILD_SRCS/nvidia-oot/device-tree/platform/generic-dts/dtbs/tegra234-camera-d5xx-overlay-fg24-8ch.dtbo"
```
ymodlin: Same bare `test -f` under `set -e` flagged on line 189 — now duplicated, so the silent-abort surface doubles while both neighbouring `cp`s stay tolerant (`|| true`).

Worth noting this one gates an overlay that this PR does not verify: the 8ch header itself says the `0x1f`/`0x20`/`0x21` aliases are not yet validated on FG24, and the Verification section covers only the 4ch/5ch cases — yet the `.dtbo` is built and copied to `/boot` unconditionally for `FG24=1`. Either give the check a message or drop it.

hardware/realsense/tegra234-camera-d5xx-overlay-fg24-4ch.dts
```diff
@@ -9,17 +9,68 @@
...
+ *   channels depending on D5XX_CHANNEL_COUNT below. Do not add a new .dts per
+ *   stream combination.
+ *
+ * SUPERSET / OVERLAY
+ *   overlay  - a /plugin/ device tree compiled to .dtbo and merged by the
+ *              bootloader into the base board DTB at boot (fragment@0 /
+ *              __overlay__). It patches nodes into the base tree; it does not
+ *              replace it. Selected by the OVERLAYS line in
+ *              /boot/extlinux/extlinux.conf, so changing it needs a reboot.
+ *   superset - declare the MAXIMUM hardware topology (all 8 VCs) once, and let
```
ymodlin: **CLAUDE.md guideline (comment style)** — this header grows ~55 lines of tutorial prose (what a DT overlay is, the full layer-mapping table, the per-VC cost budget). The rule is 2-3 line comments stating the *why*, with long rationale in the commit message / PR description / `CLAUDE.md`. The VC table and the `D5XX_CHANNEL_COUNT` selector note are worth keeping; the overlay/superset tutorial and cost analysis are not.

Minor: the "+1 SER port node" per-VC cost does not match the file — `ser_a` has no per-VC port nodes at all.

ymodlin (commented):
Re-review of the force-pushed revision (`1a5a6a25` → `4882868a`). New in this revision: the 8-channel superset overlay (VC5-VC7 Perception_1..3), the `D5XX_CHANNEL_COUNT` selector rework, and the build wiring. `d5xx.c` / `max96724.c` are unchanged, so the three earlier comments still stand.

One blocking item: the new `cam-type = "PER1/PER2/PER3"` values are not handled anywhere in `d5xx.c`, and the unmatched path silently falls back to the **depth** control block.

kernel/realsense/d5xx.c
ymodlin: Still open in `4882868a`, and this revision raises the stakes: the 8ch overlay adds VC5-VC7, so the `vc_id = is_depth ? 0 : is_rgb2 ? 4 : rgb1 ? 1 : is_y8 ? 2 : 3` chain now silently resolves **3 (IMU's VC)** for the three PER nodes. Deriving from `state->g_ctx.dst_vc` would cover all 8 VCs in one go.

nvidia-oot/files/6.2/drivers/media/i2c/max96724.c
polejoe: Checked against the current head. Tunnel-mode allocation is reference-counted per physical link pipe. Every failure after allocation now rolls back through `d5x_release_serdes_pipe()`, including initial pipe setup, re-apply after DES re-arm, stream configuration, and STREAMON timeout. The helper ignores an invalid pipe id and clears `pipe_id` only after a successful release, so the outer error path cannot decrement the same reference twice. Normal STOP releases one reference for the stream that acquired it.

kernel/realsense/d5xx.c
polejoe: Fixed in the current head. Both SerDes allocation and HKR metadata programming now use `state->g_ctx.dst_vc` as the single VC source. The old cam-type-derived VC chain is no longer used in the SerDes path.

build_all.sh
polejoe: The 5-channel overlay and its bare artifact check were removed when the PR was narrowed back to the verified VC4 dual-RGB scope. This thread is outdated.

hardware/realsense/tegra234-camera-d5xx-overlay-fg24-4ch.dts
polejoe: The PER1/PER2/PER3 overlay work was removed from this PR. The current patch only adds the verified dual-RGB VC4 channel, so those unclassified control nodes are no longer present.

build_all.sh
polejoe: The unverified 8-channel overlay and its artifact check were removed from this PR. VC5-VC7 remain outside this change.

hardware/realsense/tegra234-camera-d5xx-overlay-fg24-4ch.dts
polejoe: The long 8-channel/PER tutorial block was removed together with that out-of-scope overlay work. The remaining overlay comments are limited to the VC4 dual-RGB contract.

polejoe:
Review findings addressed in `dd2b96b`: D585 product identity gates RGB-right admission and the 2C image profiles at 30 FPS; non-2C products reject RGB-right rather than rejecting VC4 itself; tunnel-pipe references are rolled back on setup, re-apply, configure, and STREAMON failure paths. The JP5.0.2-JP7.1 build matrix passes. The `git diff --check` warnings are unified-diff context markers inside an embedded patch file, not source whitespace; changing them would corrupt the patch. @yalizhanghello Please re-review the current head.

ghu3 (approved):

PR #577, opened 2026-08-12
[D585][GMSL] Configure tunnel CSI lane rates to 1300M Serializer / 1500M Deserializer (Tunnel Mode)

polejoe:
## Follow-up

PR #567 was merged into the already-merged PR #568 head branch, so its changes did not reach `dev`. This PR reapplies the change directly on the latest `dev` with a linear, rebased history.

## Summary

- Configure the D5x FG24/MAX96724 tunnel path for 1300 Mbps/lane from D585 to MAX96717 and 1500 Mbps/lane from MAX96724 to Host, using four lanes on both CSI segments.
- Keep the device-tree lane rates, NVIDIA bandwidth value, and per-format pixel clocks consistent.
- Preserve the legacy MAX96724 2000 Mbps/lane default for overlays without an explicit output rate. The FG24 overlay explicitly selects 1500 Mbps/lane.
- Treat an explicit camera input rate as a board/link property and validate it against firmware-reported capabilities. Invalid explicit values fail initialization instead of silently creating camera/serializer configuration drift.
- D4xx overlays do not request a rate override, so D4xx keeps its existing 1000 Mbps/lane rate and lane count.
- Keep continuous clock enabled and deskew calibration disabled at 1500 Mbps/lane.

The configured GMSL2 forward link is 6 Gbps. For this single-link FG24 topology, a four-lane 1500 Mbps/lane MAX96724 D-PHY output already matches that input-link ceiling; 2000 Mbps/lane adds output-side headroom but does not increase sustained input-limited E2E throughput. MAX96724 supports up to 2.5 Gbps per D-PHY lane.

Official references:

- https://www.analog.com/en/products/MAX96724.html
- https://www.analog.com/media/en/technical-documentation/data-sheets/max96724.pdf

## Validation

- Jetson67, JP 6.2.2 / L4T 36.5 / kernel 5.15.185-tegra, D5x tunnel mode with MAX96717/MAX96724.
- Full clean build and deployment passed. Runtime configuration reported four lanes, 1300 Mbps/lane input, 1500 Mbps/lane output, continuous clock, and deskew disabled.
- Direct V4L2 with wire metadata enabled: Depth Z16 + IR Y8I + RGB YUYV, all 1280x960@60, 5 seconds per cycle, 300 consecutive start/stop cycles.
- 300/300 application passes and 300/300 Host-clean cycles; 0 VI errors, EMBED/PIXEL_RUNAWAY, timeout/recovery, panic/Oops, or Host reboot.
- Host mean good-frame FPS: Depth 60.00, IR 60.00, RGB 60.00. HKR request/done counts matched and HIF done cadence remained approximately 60 FPS.
- SerDes link-integrity, ECC/CRC, decode/idle, pixel-CRC, and memory-ECC counters remained zero.
- JP6.2.2 kernel objects compile successfully, the FG24 overlay builds to DTBO, and checkpatch reports 0 errors / 0 warnings for the final patch.

Original reviewed change: https://github.com/realsenseai/realsense_mipi_platform_driver/pull/567

### DIFF
--- a/hardware/realsense/tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dts
+++ b/hardware/realsense/tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dts
@@ -127,7 +127,7 @@
 				status = "okay";
 				compatible = "nvidia, tegra-camera-platform";
 				num_csi_lanes = <4>;
-				max_lane_speed = <2000000>;
+				max_lane_speed = <1500000>;
 				min_bits_per_pixel = <10>;
 				vi_peak_byte_per_pixel = <2>;
 				vi_bw_margin_pct = <25>;
@@ -334,7 +334,9 @@
 							reg = <0x27>;
 							compatible = "adi,max96724";
 							csi-mode = "2x4";
+							csi-lane-count = <4>;
 							max-src = <1>;
+							csi-lane-rate-mbps = <1500>;
 							link-mask = <0x1>;
 							gmsl-link-speed = <6>;
 						};
@@ -364,6 +366,7 @@
 							compatible = "realsense,d5xx";
 							use_sensor_mode_id = "true";
 							cam-type = "IMU";
+							mipi-lane-rate-mbps = <1300>;
 							nvidia,gmsl-ser-device = <&ser_a>;
 							nvidia,gmsl-dser-device = <&dser>;
 
@@ -392,9 +395,11 @@
 								phy_mode = "DPHY";
 								discontinuous_clk = "no";
 								cil_settletime = "0";
+								deskew_initial_enable = "false";
+								deskew_periodic_enable = "false";
 								mclk_khz = "24000";
-								pix_clk_hz = "1000000000";
-								serdes_pix_clk_hz = "1000000000";
+								pix_clk_hz = "700000000";
+								serdes_pix_clk_hz = "750000000";
 								line_length = "640";
 								embedded_metadata_height = "0";
 							};
@@ -422,6 +427,7 @@
 							compatible = "realsense,d5xx";
 							use_sensor_mode_id = "true";
 							cam-type = "Y8";
+							mipi-lane-rate-mbps = <1300>;
 							nvidia,gmsl-ser-device = <&ser_a>;
 							nvidia,gmsl-dser-device = <&dser>;
 
@@ -451,9 +457,11 @@
 								phy_mode = "DPHY";
 								discontinuous_clk = "no";
 								cil_settletime = "0";
+								deskew_initial_enable = "false";
+								deskew_periodic_enable = "false";
 								mclk_khz = "24000";
-								pix_clk_hz = "1000000000";
-								serdes_pix_clk_hz = "1000000000";
+								pix_clk_hz = "700000000";
+								serdes_pix_clk_hz = "750000000";
 								line_length = "1280";
 								embedded_metadata_height = "1";
 							};
@@ -469,11 +477,11 @@
 								phy_mode = "DPHY";
 								discontinuous_clk = "no";
 								cil_settletime = "0";
-								deskew_initial_enable = "true";
-								deskew_periodic_enable = "true";
+								deskew_initial_enable = "false";
+								deskew_periodic_enable = "false";
 								mclk_khz = "24000";
-								pix_clk_hz = "500000000";
-								serdes_pix_clk_hz = "500000000";
+								pix_clk_hz = "350000000";
+								serdes_pix_clk_hz = "375000000";
 								line_length = "1280";
 								embedded_metadata_height = "1";
 							};
@@ -501,6 +509,7 @@
 							compatible = "realsense,d5xx";
 							use_sensor_mode_id = "true";
 							cam-type = "RGB";
+							mipi-lane-rate-mbps = <1300>;
 							nvidia,gmsl-ser-device = <&ser_a>;
 							nvidia,gmsl-dser-device = <&dser>;
 
@@ -529,11 +538,11 @@
 								phy_mode = "DPHY";
 								discontinuous_clk = "no";
 								cil_settletime = "0";
-								deskew_initial_enable = "true";
-								deskew_periodic_enable = "true";
+								deskew_initial_enable = "false";
+								deskew_periodic_enable = "false";
 								mclk_khz = "24000";
-								pix_clk_hz = "500000000";
-								serdes_pix_clk_hz = "500000000";
+								pix_clk_hz = "350000000";
+								serdes_pix_clk_hz = "375000000";
 								line_length = "1920";
 								embedded_metadata_height = "1";
 							};
@@ -560,6 +569,7 @@
 							compatible = "realsense,d5xx";
 							use_sensor_mode_id = "true";
 							cam-type = "Depth";
+							mipi-lane-rate-mbps = <1300>;
 							nvidia,gmsl-ser-device = <&ser_a>;
 							nvidia,gmsl-dser-device = <&dser>;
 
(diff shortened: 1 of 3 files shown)


ghu3 (dismissed):

pinquanw (dismissed):

ymodlin:
**[rs-agentic] automated review** — external contribution, so this is limited to potential BUGs only. No approval implied.

*(Posted as a PR-level comment because an unsubmitted draft review already exists on this PR and GitHub blocks concurrent inline comments — I did not touch that draft.)*

**1. `nvidia-oot/max96724.c:95` — `MAX96724_CSI_LANE_RATE_DEFAULT_MBPS 1500` regresses the other max96724 board**
The old code wrote `MAX96724_DPLL_2000MBPS` (0x34 = 2000 Mbps) unconditionally for *every* max96724 board. `tegra234-camera-d4xx-overlay-d5xx.dts` does not set the new `csi-lane-rate-mbps` property, so it silently drops 2000 → 1500 Mbps. Either keep 2000 as the default, or set `csi-lane-rate-mbps` in that overlay within this PR.

**2. `tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dts:483` (and 544, 606) — `pix_clk_hz` inconsistent with both link rates**
`serdes_pix_clk_hz = 375000000` is right for 1500 Mbps (375M x 16bpp / 4 lanes). But `pix_clk_hz = 350000000` implies 1400 Mbps, which matches neither the 1300 Mbps serializer nor the 1500 Mbps deserializer in the title. CLAUDE.md gives `325000000` for 4 x 1300 Mbps at 16bpp. The 350M/700M pair looks carried over from the 1400 Mbps pixel-mode change in #570 — please confirm the intended value.

nvidia-oot/max96724.c
```diff
@@ -87,8 +87,13 @@
...
 #define MAX96724_CSI_PHY_EN_ALL		0xF0
 #define MAX96724_CSI_LANE_MAP_DEFAULT	0xE4
-#define MAX96724_LANE_CTRL_4LANE	0xC0
-#define MAX96724_DPLL_2000MBPS		0x34
+#define MAX96724_LANE_CTRL_NUM_LANES(n)	(((n) - 1) << 6)
+#define MAX96724_DPLL_FREQ_OVERRIDE	BIT(5)
+#define MAX96724_CSI_LANE_RATE_MIN_MBPS	100
+#define MAX96724_CSI_LANE_RATE_MAX_MBPS	2500
+#define MAX96724_CSI_LANE_RATE_STEP_MBPS	100
+#define MAX96724_CSI_LANE_RATE_DEFAULT_MBPS	1500
```
ymodlin: **BUG (likely regression)** — the effective default drops from 2000 Mbps (old `MAX96724_DPLL_2000MBPS = 0x34`) to 1500 Mbps for every max96724 DT that does *not* set `csi-lane-rate-mbps`.

`hardware/realsense/tegra234-camera-d4xx-overlay-d5xx.dts` also instantiates `adi,max96724` and does not carry the new property, so its DPLL silently changes without that overlay being touched or validated in this PR.

Either keep the default at 2000 (preserving current behaviour) or add `csi-lane-rate-mbps` to the other overlay in this same change. CLAUDE.md asks for exactly this lockstep for the analogous max96712 `BACKTOP25` rate.

nvidia-oot/max96724.c
```diff
@@ -1370,6 +1383,30 @@ static int max96724_parse_dt(struct max96724 *priv,
 		return -EINVAL;
 	}
 	priv->link_mask = value;
+	priv->csi_lane_count = MAX96724_CSI_LANE_COUNT_DEFAULT;
+	err = of_property_read_u32(node, "csi-lane-count", &value);
```
ymodlin: **Enhancement** — `csi-lane-count` is a no-op knob today: any value other than 4 is rejected with a warning and the default (4) is kept, so the property can never change `MAX96724_LANE_CTRL_NUM_LANES()`. Either drop it until 2-lane is actually supported, or accept 2 as well.

Also: `csi-lane-count`, `csi-lane-rate-mbps` and `mipi-lane-rate-mbps` are all new DT bindings with no documentation — worth a short note in the overlay header comment.

kernel/realsense/d4xx.c
```diff
@@ -6219,24 +6222,50 @@ static int ds5_hw_init(struct i2c_client *c, struct ds5 *state)
...
+				 "invalid DT MIPI lane rate %u, using %u Mbps\n",
+				 dt_lane_rate, default_lane_rate);
+		} else {
+			lane_rate = dt_lane_rate;
+		}
+	}
+	if (mipi_caps_valid &&
+	    (lane_rate < drate_min || lane_rate > drate_max)) {
+		dev_warn(sd->dev,
+			 "MIPI lane rate %u outside FW range %u-%u, using %u Mbps\n",
```
ymodlin: **Logic issue** — when no DT override is present, `lane_rate == default_lane_rate`, so this branch warns and then re-assigns the same value it just rejected. The message reads `"... outside FW range 1000-1100, using 1200 Mbps"` while 1200 is what was rejected. The fallback is not guaranteed to be in range either.

Suggest validating only the DT-supplied value against `drate_min/drate_max` and leaving the compiled-in default alone.

Separately: if FW reports `drate_max < 1300`, the DT's `mipi-lane-rate-mbps = <1300>` is silently dropped to 1200 while the SerDes/overlay stay sized for 1300 — that config drift surfaces only as a `dev_warn`. Consider failing `ds5_hw_init()` on an explicitly-specified but unusable DT rate.

hardware/realsense/tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dts
```diff
@@ -364,6 +366,7 @@
 							compatible = "realsense,d5xx";
 							use_sensor_mode_id = "true";
 							cam-type = "IMU";
+							mipi-lane-rate-mbps = <1300>;
```
ymodlin, line 380: **Guideline (CLAUDE.md)** — `mipi-lane-rate-mbps` is only honoured when `is_d58x` is true, so the same property in a D4xx overlay is parsed and silently ignored. That makes a DT property's meaning depend on the *detected camera model*, which is the pattern CLAUDE.md calls out ("any behavior that depends on the connected camera model must be decided in-driver from the detected device type, never via a DT property", cf. RSDEV-12608).

If the rate is genuinely a board/link fact, honour it for all models (still bounded by the FW caps); if it is a D585-only FW capability, derive it from the detected device type instead of DT.

ymodlin (commented):
Auto-generated review by the **rs-agentic bot**.

Good direction — making the lane rate data-driven instead of hardcoded is the right call, and the `serdes_pix_clk_hz` values are consistent with 1500 Mbps x4 (375M x16bpp, 750M x8bpp) and with `max_lane_speed = <1500000>`.

One likely regression (max96724 default rate) plus a few smaller items inline.

nvidia-oot/max96724.c
polejoe: Valid. Fixed in 2157b5c: the generic MAX96724 default remains 2000 Mbps/lane, while FG24 explicitly selects 1500. For one 6 Gbps GMSL2 link, 4 x 1500 already reaches the input ceiling; 2000 adds no sustained E2E benefit. ADI: https://www.analog.com/en/products/MAX96724.html.

nvidia-oot/max96724.c
polejoe: Addressed in 2157b5c. The current path supports only the validated 2x4/four-lane topology, so non-four-lane values now fail instead of silently falling back. The three lane properties are documented; two-lane support remains separate work.

kernel/realsense/d4xx.c
polejoe: Valid. Fixed in 2157b5c. Only explicit DT overrides are validated; malformed, unavailable, or out-of-range values now fail initialization instead of silently creating camera/serializer rate drift.

hardware/realsense/tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dts
polejoe: Valid. Fixed in 2157b5c. The D58x-only gate was removed: the DT rate is treated as a board/link property and checked against FW capabilities. Existing D4xx overlays have no override and keep their current rate and lane count.

polejoe:
Addressed in `2157b5c`:

- Kept the generic MAX96724 default at 2000 Mbps/lane; FG24 explicitly uses 1500.
- Corrected camera-side `pix_clk_hz` for 1300 Mbps/lane.
- Removed silent DT-rate fallback and the D58x-only property gate.
- Documented and strictly validated the supported four-lane topology.

The branch was rebased onto the latest `dev`; the PR now has a linear two-commit history and no merge commit.

nvidia-oot/max96724.c
ymodlin: change approved

nvidia-oot/max96724.c
ymodlin: change approved

kernel/realsense/d4xx.c

hardware/realsense/tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dts
ymodlin: change approved

ejgoldik (approved):

polejoe:
Correction: this PR was already in the `dev` merge queue and should not have been retargeted.

After the base was changed in error, GitHub merged #577 into `dev_gmsl_merge` as `1521835f74062c06fa713d4d583156fa2fe58429`. A merged pull request cannot have its base restored.

The exact unchanged head commit, `2157b5c488d565534ac1a1d6df4bd2364e5de8e8`, is now reopened as #596 targeting `dev`. The original approving reviewer has been requested again because GitHub does not transfer approvals between pull requests.

`dev_gmsl_merge` will only be used for eligible PRs that have not already entered the `dev` merge queue.

PR #584, opened 2026-08-13
[RSDEV-13639][D58X] Expose GMSL camera controls

lemon6630:
Jira: [RSDEV-13639](https://rsjira.realsenseai.com/browse/RSDEV-13639)

## What changed

- expose Depth-primary Device Mode and active-2C AE Policy V4L2 controls for D58x
- read and write exact one-byte GMSL XU windows at 0x4528 and 0x4530
- remove all Device Mode / 2C AE custom-HWMC commands, helpers, error mapping, and generic-HWMC interception
- expose the shared Auto / Depth Master / RGB Master / Hybrid AE contract as values 0..3
- keep configured and active Device Mode separate until HKR reboot
- require camera idle state for both camera-wide SET operations and activate AE only for the booted 2C mode

## Why

Device Mode and 2C AE now have dedicated HKR GMSL XU bridge endpoints. The Host no longer needs the temporary feature-specific HWMC transport and can use the same one-byte ABI directly.

## Validation

- full JP6.2.1 build passed at the current head
- d4xx.ko: version 1.0.6.3, srcversion 9C4BD01D5823E60AB271AF2, SHA-256 13e6315fa362cfdc5f8a09331d5a152d7630e72b05c7de834c622b8db3da784d
- strict Host delta checkpatch: 0 errors, 0 warnings, 0 checks
- final HKR rebase build passed as BKC 7.58.41077.14027; hardware validation used the patch-equivalent BKC 7.58.41072.14026 stack
- D585 GMSL hardware: 3C -> 2C persistence, AE policies 0/1/2/3 SET/GET, invalid ERANGE, streaming EBUSY
- Depth/Color/IR/IMU regression: 120/120/120/400 frames with no new dmesg error
- final persistent DFU reached dfuMANIFEST-WAIT-RESET, status 0; bench was restored to 3C
- current Host module was deployed on Jetson and passed a final 120-frame Depth stream against firmware 7.58.112.202

Current hardware requires waiting for HKR reboot and reloading d4xx because HKR reset power-cycles GMSL. Automatic teardown/re-probe remains outside this PR.

### DIFF
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -500,6 +500,8 @@ struct ds5_ctrls {
 		struct v4l2_ctrl *query_sub_stream;
 		struct v4l2_ctrl *set_sub_stream;
 		struct v4l2_ctrl *sync_mode;
+		struct v4l2_ctrl *device_mode;
+		struct v4l2_ctrl *dual_rgb_ae_policy;
 		/* RGB-only ISP controls. Only ae_priority needs a stored
 		 * pointer because it must be disabled per-SKU after probe
 		 * (D40X/D401 does not support it). The remaining controls
@@ -665,6 +667,9 @@ struct ds5_dev {
 	/* Pointer to the primary DS5 struct */
 	struct ds5 *ds5_primary;
 	bool serdes_setup_complete;
+	int configured_device_mode;
+	int active_device_mode;
+	bool device_mode_valid;
 
 	bool depth_streaming;
 	bool ir_streaming;
@@ -2567,6 +2572,23 @@ enum ds5_sync_mode {
 
 /* Depth AE mode: single R/W control, maps to librealsense XU selector 0x11 */
 #define DS5_CAMERA_CID_AE_MODE		(DS5_CAMERA_CID_BASE+35)
+#define DS5_CAMERA_CID_DEVICE_MODE	(DS5_CAMERA_CID_BASE + 36)
+#define DS5_CAMERA_CID_DUAL_RGB_AE_POLICY (DS5_CAMERA_CID_BASE + 37)
+
+enum d500_device_mode {
+	D500_DEVICE_MODE_3C = 0,
+	D500_DEVICE_MODE_2C = 1,
+};
+
+enum d500_2c_ae_mode {
+	D500_2C_AE_POLICY_AUTO = 0,
+	D500_2C_AE_POLICY_DEPTH_MASTER = 1,
+	D500_2C_AE_POLICY_RGB_MASTER = 2,
+	D500_2C_AE_POLICY_HYBRID = 3,
+};
+
+#define D500_DEVICE_MODE_XU_BASE		0x4528
+#define D500_DUAL_RGB_AE_XU_BASE	0x4530
 
 /* Auto-exposure algorithm types — mirrors FW ETAeType */
 enum ds5_ae_type {
@@ -2699,6 +2721,133 @@ static int ds5_hwmc_send(struct ds5 *state,
 	ds5_write_with_check(state, DS5_HWMC_EXEC, 0x01); /* execute cmd */
 
 	return 0;
+}
+
+/* Caller must hold state->ds5_dev->lock. */
+static bool d5xx_camera_is_idle_locked(struct ds5 *state)
+{
+	return !state->ds5_dev->depth_streaming &&
+	       !state->ds5_dev->rgb_streaming &&
+	       !state->ds5_dev->ir_streaming &&
+	       !state->ds5_dev->imu_streaming;
+}
+
+static int d5xx_get_device_mode(struct ds5 *state, u32 *mode)
+{
+	u8 value;
+	int ret;
+
+	if (!mode)
+		return -EINVAL;
+
+	ret = ds5_raw_read(state, D500_DEVICE_MODE_XU_BASE,
+			   &value, sizeof(value));
+	if (ret)
+		return ret;
+	if (value > D500_DEVICE_MODE_2C)
+		return -EBADMSG;
+
+	*mode = value;
+	return 0;
+}
+
+static int d5xx_set_device_mode(struct ds5 *state, u32 mode)
+{
+	u8 value;
+	int ret;
+
+	if (mode > D500_DEVICE_MODE_2C)
+		return -EINVAL;
+	value = mode;
+
+	mutex_lock(&state->ds5_dev->lock);
+	if (!d5xx_camera_is_idle_locked(state)) {
+		ret = -EBUSY;
+	} else {
+		ret = ds5_raw_write(state, D500_DEVICE_MODE_XU_BASE,
+				    &value, sizeof(value));
+		if (!ret)
+			state->ds5_dev->configured_device_mode = mode;
+	}
+	mutex_unlock(&state->ds5_dev->lock);
+
+	return ret;
+}
+
+static int d5xx_get_ae_policy(struct ds5 *state, u32 *policy)
+{
+	u8 value;
+	int ret;
+
+	if (!policy)
+		return -EINVAL;
+
+	mutex_lock(&state->ds5_dev->lock);
+	ret = (!state->ds5_dev->device_mode_valid ||
+	       state->ds5_dev->active_device_mode != D500_DEVICE_MODE_2C) ?
+		-EOPNOTSUPP : 0;
+	mutex_unlock(&state->ds5_dev->lock);
+	if (ret)
+		return ret;
+
+	ret = ds5_raw_read(state, D500_DUAL_RGB_AE_XU_BASE,
+			   &value, sizeof(value));
+	if (ret)
+		return ret;
+	if (value > D500_2C_AE_POLICY_HYBRID)
+		return -EBADMSG;
+
+	*policy = value;
+	return 0;
+}
+
+static int d5xx_set_ae_policy(struct ds5 *state, u32 policy)
+{
+	u8 value;
+	int ret;
+
+	if (policy > D500_2C_AE_POLICY_HYBRID)
+		return -EINVAL;
+	value = policy;
+
+	mutex_lock(&state->ds5_dev->lock);
+	ret = (!state->ds5_dev->device_mode_valid ||
+	       state->ds5_dev->active_device_mode != D500_DEVICE_MODE_2C) ?
+		-EOPNOTSUPP : 0;
+	if (!ret && !d5xx_camera_is_idle_locked(state))
+		ret = -EBUSY;
+	if (!ret)
+		ret = ds5_raw_write(state, D500_DUAL_RGB_AE_XU_BASE,
+				    &value, sizeof(value));
+	mutex_unlock(&state->ds5_dev->lock);
+
+	return ret;
+}
+
+static int d5xx_refresh_device_mode(struct ds5 *state, bool boot_mode)
+{
+	u32 mode;
+	int ret;
+
+	ret = d5xx_get_device_mode(state, &mode);
+	mutex_lock(&state->ds5_dev->lock);
+	if (!ret) {
+		state->ds5_dev->configured_device_mode = mode;
+		if (boot_mode)
+			state->ds5_dev->active_device_mode = mode;
+		state->ds5_dev->device_mode_valid = true;
+	} else if (boot_mode) {
+		state->ds5_dev->device_mode_valid = false;
+	}
+	mutex_unlock(&state->ds5_dev->lock);
+
+	if (state->ctrls.dual_rgb_ae_policy) {
+		const bool active = !ret && mode == D500_DEVICE_MODE_2C;
+
+		v4l2_ctrl_activate(state->ctrls.dual_rgb_ae_policy, active);
+	}
+
+	return ret;
 }
 
 static int ds5_set_calibration_data(struct ds5 *state,
@@ -3050,6 +3199,18 @@ static int ds5_hw_reset_with_recovery(struct ds5 *state)
 			dev_warn(&state->client->dev,
 				"%s(): serializer ESYNC %s after HW reset failed (%d)\n",
 				__func__, need_esync ? "enable" : "disable", ret);
+	}
+
+	if (dev_type == DS5_DEVICE_TYPE_D58X) {
+		struct ds5 *owner = state->ds5_dev->ds5_primary ?
+			state->ds5_dev->ds5_primary : state;
+		int mode_ret;
+
+		mode_ret = d5xx_refresh_device_mode(owner, true);
+		if (mode_ret)
+			dev_warn(&state->client->dev,
+				 "%s(): D58x device-mode refresh after HW reset failed (%d)\n",
+				 __func__, mode_ret);
 	}
 
 	WRITE_ONCE(state->ds5_dev->last_reset_jiffies, jiffies);
@@ -3453,6 +3614,18 @@ static int ds5_s_ctrl(struct v4l2_ctrl *ctrl)
 				ret = ds5_set_ser_esync_tunneling(state, need_esync);
 			}
 		}
+		break;
+	case DS5_CAMERA_CID_DEVICE_MODE:
+		if (state->is_depth &&
+		    READ_ONCE(state->ds5_dev->cached_device_type) ==
+			    DS5_DEVICE_TYPE_D58X)
+			ret = d5xx_set_device_mode(state, ctrl->val);
+		break;
+	case DS5_CAMERA_CID_DUAL_RGB_AE_POLICY:
+		if (state->is_depth &&
+		    READ_ONCE(state->ds5_dev->cached_device_type) ==
+			    DS5_DEVICE_TYPE_D58X)
+			ret = d5xx_set_ae_policy(state, ctrl->val);
 		break;
 	case DS5_CAMERA_CID_PWM:
 		if (state->is_depth)
@@ -3899,6 +4072,32 @@ static int ds5_g_volatile_ctrl(struct v4l2_ctrl *ctrl)
 		if (state->is_depth)
 			ds5_read(state, base | DS5_CAMERA_SYNC_MODE, ctrl->p_new.p_u16);
 		break;
+	case DS5_CAMERA_CID_DEVICE_MODE:
+		if (state->is_depth && ctrl->p_new.p_s32 &&
+		    READ_ONCE(state->ds5_dev->cached_device_type) ==
+			    DS5_DEVICE_TYPE_D58X) {
+			u32 mode;
+
+			ret = d5xx_get_device_mode(state, &mode);
+			if (!ret) {
+				*ctrl->p_new.p_s32 = mode;
+				mutex_lock(&state->ds5_dev->lock);
+				state->ds5_dev->configured_device_mode = mode;
+				mutex_unlock(&state->ds5_dev->lock);
+			}
+		}
+		break;
+	case DS5_CAMERA_CID_DUAL_RGB_AE_POLICY:
+		if (state->is_depth && ctrl->p_new.p_s32 &&
+		    READ_ONCE(state->ds5_dev->cached_device_type) ==
+			    DS5_DEVICE_TYPE_D58X) {
+			u32 policy;
+
+			ret = d5xx_get_ae_policy(state, &policy);
+			if (!ret)
+				*ctrl->p_new.p_s32 = policy;
+		}
+		break;
 	case DS5_CAMERA_CID_PWM:
 		if (state->is_depth)
 			ds5_read(state, base | DS5_PWM_FREQUENCY, ctrl->p_new.p_u16);
@@ -4251,6 +4450,42 @@ static const struct v4l2_ctrl_config ds5_ctrl_error_code = {
 	.flags = DS5_READ_ONLY_TELEMETRY_FLAGS,
 };
 
+static const char * const d58x_device_mode_menu[] = {
+	[D500_DEVICE_MODE_3C] = "3C",
+	[D500_DEVICE_MODE_2C] = "2C",
+};
+
+static const struct v4l2_ctrl_config ds5_ctrl_device_mode_d58x = {
+	.ops = &ds5_ctrl_ops,
+	.id = DS5_CAMERA_CID_DEVICE_MODE,
+	.name = "Device Mode",
+	.type = V4L2_CTRL_TYPE_MENU,
+	.min = D500_DEVICE_MODE_3C,
+	.max = D500_DEVICE_MODE_2C,
+	.def = D500_DEVICE_MODE_3C,
+	.qmenu = d58x_device_mode_menu,
+	.flags = V4L2_CTRL_FLAG_VOLATILE | V4L2_CTRL_FLAG_EXECUTE_ON_WRITE,
+};
+
+static const char * const d58x_ae_policy_menu[] = {
+	[D500_2C_AE_POLICY_AUTO] = "Auto",
+	[D500_2C_AE_POLICY_DEPTH_MASTER] = "Depth Master",
+	[D500_2C_AE_POLICY_RGB_MASTER] = "RGB Master",
+	[D500_2C_AE_POLICY_HYBRID] = "Hybrid",
+};
+
+static const struct v4l2_ctrl_config ds5_ctrl_dual_rgb_ae_policy_d58x = {
+	.ops = &ds5_ctrl_ops,
+	.id = DS5_CAMERA_CID_DUAL_RGB_AE_POLICY,
+	.name = "2C AE Policy",
+	.type = V4L2_CTRL_TYPE_MENU,
+	.min = D500_2C_AE_POLICY_AUTO,
+	.max = D500_2C_AE_POLICY_HYBRID,
+	.def = D500_2C_AE_POLICY_DEPTH_MASTER,
+	.qmenu = d58x_ae_policy_menu,
+	.flags = V4L2_CTRL_FLAG_VOLATILE | V4L2_CTRL_FLAG_EXECUTE_ON_WRITE,
+};
+
 static const struct v4l2_ctrl_config ds5_ctrl_pwm = {
 	.ops = &ds5_ctrl_ops,
 	.id = DS5_CAMERA_CID_PWM,
@@ -4317,6 +4552,9 @@ static void ds5_init_ds5_dev(struct ds5 *state, struct ds5_dev *ds5_dev)
 	ds5_dev->ds5_primary = state;
 	ds5_dev->serdes_setup_complete = false;
 	ds5_dev->cached_device_type = DS5_DEVICE_TYPE_UNKNOWN;
+	ds5_dev->configured_device_mode = D500_DEVICE_MODE_3C;
+	ds5_dev->active_device_mode = D500_DEVICE_MODE_3C;
+	ds5_dev->device_mode_valid = false;
 	mutex_unlock(&ds5_dev->lock);
 	ds5_reset_streaming_flags(ds5_dev);
 }
@@ -5222,6 +5460,15 @@ static int ds5_ctrl_init(struct ds5 *state, int sid)
 	if (sid == DEPTH_SID) {
 		ctrls->sync_mode = v4l2_ctrl_new_custom(hdl, &ds5_ctrl_sync_mode, sensor);
 		if (is_d58x) {
+			ctrls->device_mode =
+				v4l2_ctrl_new_custom(hdl, &ds5_ctrl_device_mode_d58x,
+						     sensor);
+			ctrls->dual_rgb_ae_policy =
+				v4l2_ctrl_new_custom(hdl,
+						     &ds5_ctrl_dual_rgb_ae_policy_d58x,
+						     sensor);
+			if (ctrls->dual_rgb_ae_policy)
+				v4l2_ctrl_activate(ctrls->dual_rgb_ae_policy, false);
 			v4l2_ctrl_new_custom(hdl, &ds5_ctrl_visual_preset, sensor);
 			v4l2_ctrl_new_custom(hdl, &ds5_ctrl_soc_pvt_temperature,
 					     sensor);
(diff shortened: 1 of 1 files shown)


ymodlin (commented):
Review of the D58x camera-controls change. Two blocking bugs (a self-deadlock on the HW-reset path and errno rewriting), plus a missing-tests guideline gap per the `add-xu-command` skill. Details inline.

kernel/realsense/d4xx.c
```diff
@@ -2699,6 +2782,250 @@ static int ds5_hwmc_send(struct ds5 *state,
...
+		state->ds5_dev->device_mode_valid = true;
+	} else if (boot_mode) {
+		state->ds5_dev->device_mode_valid = false;
+	}
+	mutex_unlock(&state->ds5_dev->lock);
+
+	if (state->ctrls.dual_rgb_ae_policy) {
+		const bool active = !ret && mode == DS5_D58X_DEVICE_MODE_2C;
+
+		v4l2_ctrl_activate(state->ctrls.dual_rgb_ae_policy, active);
```
ymodlin, line 2847: **BUG (self-deadlock)** — `v4l2_ctrl_activate()` acquires `ctrl->handler->lock`, but this function is reachable from `ds5_s_ctrl()`: `DS5_CAMERA_CID_HW_RESET` (and HWMC_RW opcode `0x20`) → `ds5_hw_reset_with_recovery()` → `ds5_refresh_d58x_device_mode()`. The V4L2 core already holds that handler lock while calling `s_ctrl`, and both `hw_reset` and `dual_rgb_ae_policy` live on `handler_depth` — so a HW reset issued on the depth node deadlocks on itself.

Only the probe caller is safe. Gate the activation (e.g. do it only when `boot_mode` comes from probe, or move it outside the `s_ctrl` path).

kernel/realsense/d4xx.c
```diff
@@ -2591,6 +2626,54 @@ enum DS5_HWMC_ERR {
...
+	case -DS5_FW_ERR_ILLEGAL_SIZE:
+		return -EBADMSG;
+	case DS5_FW_ERR_TIMEOUT:
+	case -DS5_FW_ERR_TIMEOUT:
+		return -ETIMEDOUT;
+	case DS5_FW_ERR_NOT_SUPPORTED:
+	case -DS5_FW_ERR_NOT_SUPPORTED:
+		return -EOPNOTSUPP;
+	case DS5_FW_ERR_SW_NOT_READY:
+	case -DS5_FW_ERR_SW_NOT_READY:
```
ymodlin: **BUG** — translating the *negative* FW codes rewrites genuine Linux errnos. `ds5_hwmc_wait()` returns transport errnos and the existing `enum DS5_HWMC_ERR` values, so `DS5_HWMC_ERR_NODATA` (-21) now silently becomes `-EBUSY`, `-ENODEV` (-19) → `-EOPNOTSUPP`, `-EINTR` (-4) → `-EACCES`. That directly contradicts the "preserve transport errno values" comment just below.

FW ErrorTypes are positive — drop the `-DS5_FW_ERR_*` cases and translate positive values only.

kernel/realsense/d4xx.c
```diff
@@ -2699,6 +2782,250 @@ static int ds5_hwmc_send(struct ds5 *state,
...
+		.header = 0x14,
+		.magic_word = 0xCDAB,
+		.opcode = DS5_HWMC_OPCODE_CUSTOM_CMD,
+		.param1 = subcommand,
+	};
+
+	return ds5_hwmc_get_u32(state, &cmd, value);
+}
+
+static bool ds5_camera_is_idle(struct ds5 *state)
```
ymodlin: **Race** — the idle test is not atomic with the HWMC write that follows it: a stream can start between `ds5_camera_is_idle()` returning true and `ds5_hwmc_set_custom_u32()` landing, so a Device Mode / AE Policy write can still reach FW mid-stream despite the `-EBUSY` guard.

kernel/realsense/d4xx.c
```diff
@@ -3894,8 +4449,50 @@ static int ds5_g_volatile_ctrl(struct v4l2_ctrl *ctrl)
...
-			ds5_read(state, base | DS5_CAMERA_SYNC_MODE, ctrl->p_new.p_u16);
+		if (state->is_depth) {
+			if (READ_ONCE(state->ds5_dev->cached_device_type) ==
+			    DS5_DEVICE_TYPE_D58X && ctrl->p_new.p_u16) {
+				u16 sync_mode;
+
+				ret = ds5_hwmc_get_d58x_sync_mode(state,
+								  &sync_mode);
+				if (!ret) {
+					*ctrl->p_new.p_u16 = sync_mode;
```
ymodlin: **Minor** — `sync_mode` is a `V4L2_CTRL_TYPE_MENU` control, whose payload is `s32`; writing through `p_u16` only sets the low half and leaves the upper 16 bits stale. Use `*ctrl->p_new.p_s32`.

kernel/realsense/d4xx.c
```diff
@@ -2826,6 +3153,177 @@ static int ds5_set_ser_esync_tunneling(struct ds5 *state, bool enable)
 #endif
 }
 
+static int ds5_set_d58x_ser_esync_tunneling(struct ds5 *state, bool enable)
```
ymodlin: **Enhancement** — this is `ds5_set_ser_esync_tunneling()` (~line 3111) minus the fsync block. Consider adding a "skip fsync" argument to the existing helper rather than keeping a second near-identical copy that has to be maintained in step.

kernel/realsense/d4xx.c
```diff
@@ -5218,8 +5874,21 @@ static int ds5_ctrl_init(struct ds5 *state, int sid)
...
+		const struct v4l2_ctrl_config *sync_cfg = &ds5_ctrl_sync_mode;
+
+		if (is_d58x)
+			sync_cfg = &ds5_ctrl_sync_mode_d58x;
+		ctrls->sync_mode = v4l2_ctrl_new_custom(hdl, sync_cfg, sensor);
 		if (is_d58x) {
+			ctrls->device_mode =
+				v4l2_ctrl_new_custom(hdl, &ds5_ctrl_device_mode_d58x,
+						     sensor);
+			ctrls->dual_rgb_ae_policy =
```
ymodlin, line 5466: **Guideline (`.claude/skills/add-xu-command`)** — two new controls land with no tests and no constants mirror. The skill requires `test/v4l2_test/d4xx/constants.py` to mirror `DS5_CAMERA_CID_DEVICE_MODE` / `DS5_CAMERA_CID_DUAL_RGB_AE_POLICY` at the same offsets, and a `Test<ControlName>` class in `test/v4l2_test/tests/test_controls.py` covering enumerated / read / write_legal / set_illegal / default / roundtrip. This PR changes only `d4xx.c`.

ymodlin (commented):
### rs-agentic re-review — `cea9c5c` (Support Hybrid AE policy)

The Hybrid delta itself is consistent: all four bound sites were widened together (`ds5_get_d58x_ae_policy` L2910, `ds5_set_d58x_ae_policy` L2919, `.max` L4900, `qmenu` L4891) — no missed validation site, and the control keeps the required single-CID `VOLATILE | EXECUTE_ON_WRITE` shape.

**Still open from the review on `5f0ed3c` — this push addressed none of them:**

| # | Finding | Status on `cea9c5c` |
|---|---|---|
| 1 | Self-deadlock: `v4l2_ctrl_activate()` called with `handler_depth->lock` already held, via `ds5_s_ctrl` → `ds5_hw_reset_with_recovery()` (L3942 / L3971) → `ds5_refresh_d58x_device_mode()` (L3558) | unchanged |
| 2 | `-DS5_FW_ERR_*` cases (L2646-2668) still rewrite genuine Linux errnos, contradicting the "preserve transport errno" comment in the same `default:` block | unchanged |
| 3 | Idle→write race: `ds5_camera_is_idle()` is not atomic with the following `ds5_hwmc_set_custom_u32()` | unchanged (same pattern now on the AE-policy setter, L2924) |
| 4 | `sync_mode` menu written through `p_new.p_u16` (L4459) — the two new menu controls correctly use `p_s32` (L4473 / L4489), which confirms the inconsistency | unchanged |
| 5 | `ds5_set_ser_esync_tunneling()` near-duplicate helper | unchanged |
| 6 | No `test/v4l2_test/d4xx/constants.py` mirror and no `test_controls.py` coverage (per `.claude/skills/add-xu-command`); this commit widens the menu to **4** values, so the untested surface grew | unchanged — PR still touches only `d4xx.c` |

Not approving; PR is a draft.

kernel/realsense/d4xx.c
```diff
@@ -4249,6 +4847,62 @@ static const struct v4l2_ctrl_config ds5_ctrl_error_code = {
...
+	.type = V4L2_CTRL_TYPE_MENU,
+	.min = DS5_D58X_DEVICE_MODE_3C,
+	.max = DS5_D58X_DEVICE_MODE_2C,
+	.def = DS5_D58X_DEVICE_MODE_3C,
+	.qmenu = d58x_device_mode_menu,
+	.flags = V4L2_CTRL_FLAG_VOLATILE | V4L2_CTRL_FLAG_EXECUTE_ON_WRITE,
+};
+
+static const char * const d58x_ae_policy_menu[] = {
+	[DS5_D58X_AE_POLICY_INDEPENDENT] = "Auto",
```
ymodlin: **Minor (naming)** — value 0 is now exposed to userspace as `"Auto"`, but the enumerator is still `DS5_D58X_AE_POLICY_INDEPENDENT`, so the code and the FW/UVC menu contract disagree. Rename to `DS5_D58X_AE_POLICY_AUTO` (or add a one-line why) so a reader does not take value 0 for independent-AE behaviour.

kernel/realsense/d4xx.c
ymodlin: Correction after re-checking `cea9c5c` — the finding stands, but not for the reason given above. `ds5_hwmc_wait()` collapses every I2C failure to `DS5_HWMC_ERR_LAST`, so `-ENODEV`/`-EINTR` are not actually reachable here; disregard those examples.

The real, reachable problem is the pre-existing `enum DS5_HWMC_ERR` (L2621) — FW codes arrive **negative** on this path (`errorCode` is read raw from `DS5_HWMC_DATA`):

- `-21` is `DS5_HWMC_ERR_NODATA` in that enum but maps to `-DS5_FW_ERR_SW_NOT_READY` -> `-EBUSY` here. Two tables, same number, different meaning.
- `-1` (`ERR_CMD`) and `-6` (`ERR_PARAM`) miss every case and fall into `default:`, which returns them verbatim — userspace sees `-EPERM` / `-ENXIO`.

Worth reconciling the new `DS5_FW_ERR_*` table with `enum DS5_HWMC_ERR` and picking one sign convention before mapping.

kernel/realsense/d4xx.c
lemon6630: Thanks. I rechecked this against the target V4L2 locking contract. v4l2_ctrl_activate() requires the control-handler lock to already be held; it does not acquire that mutex itself. The Linux 5.15 implementation only changes the inactive flag and emits the control event. Calling it from s_ctrl is therefore the intended locked usage, and limiting it to probe would leave the AE control activation stale after a runtime HW reset. No code change here. References: [V4L2 control API](https://docs.kernel.org/driver-api/media/v4l2-controls.html) and [Linux v5.15 implementation](https://github.com/torvalds/linux/blob/v5.15/drivers/media/v4l2-core/v4l2-ctrls-core.c#L1659-L1676).

kernel/realsense/d4xx.c
lemon6630: Fixed in 14bc71e. The firmware path does serialize ErrorTypes as signed negative values, so those cases must remain supported. I removed the conflicting legacy DS5_HWMC_ERR aliases (especially -21=NODATA), made the firmware table/sign contract explicit, added the missing ErrorTypes mappings, retained -63 only as the transport-failure sentinel, and now check the error-payload read itself. Unknown negative results remain preserved; unknown positive firmware statuses map to -EIO.

kernel/realsense/d4xx.c
lemon6630: Fixed in 14bc71e. Device Mode and 2C AE Policy SET now hold the shared per-camera ds5_dev lock across mode/idle validation and the typed HWMC transaction. Stream start updates the shared streaming flag under the same lock before issuing its start command, so it can no longer enter between the idle check and control write. I also checked lock order: the control path is state->lock then ds5_dev->lock, while the stream path only takes ds5_dev->lock.

kernel/realsense/d4xx.c
lemon6630: Fixed in 14bc71e. Both D58x HWMC and legacy register reads now use a local u16 and publish the completed value through *ctrl->p_new.p_s32, matching the V4L2 MENU payload type.

kernel/realsense/d4xx.c
lemon6630: I kept these helpers separate intentionally. The D58x path resolves the physical camera primary owner, is MAX96724-only, and separates serializer ESYNC tunneling from deserializer FSYNC lifecycle so FSYNC can follow stream state. The legacy helper supports the existing MAX96712/MAX96724 product path and couples serializer plus FSYNC programming. Adding a skip-FSYNC mode to that shared legacy helper would broaden the D58x change into the D4xx path without reducing the product-specific branching. No code change for this enhancement.

kernel/realsense/d4xx.c
lemon6630: Fixed in 14bc71e. Added the two CID mirrors, D58x device type/markers, and 13 D58x control tests covering enumeration, ranges/defaults, reads, AE legal/illegal/all-value roundtrips, and Device Mode roundtrip. Persistent Device Mode writes require the explicit --allow-persistent-control-writes option; AE writes only run when the control is active in 2C and restore the original value. The scoped suite collects cleanly; locally all 13 skip because no D58x is attached.

kernel/realsense/d4xx.c
lemon6630: Fixed in 14bc71e: value 0 is now consistently named DS5_D58X_AE_POLICY_AUTO and remains exposed as Auto.

kernel/realsense/d4xx.c
```diff
@@ -2565,6 +2587,20 @@ enum ds5_sync_mode {
 
 /* Depth AE mode: single R/W control, maps to librealsense XU selector 0x11 */
 #define DS5_CAMERA_CID_AE_MODE		(DS5_CAMERA_CID_BASE+35)
+#define DS5_CAMERA_CID_DEVICE_MODE	(DS5_CAMERA_CID_BASE + 36)
+#define DS5_CAMERA_CID_DUAL_RGB_AE_POLICY (DS5_CAMERA_CID_BASE + 37)
+
+enum ds5_d58x_device_mode {
+	DS5_D58X_DEVICE_MODE_3C = 0,
```
yalizhanghello: rename DS5_D58X -> D500

kernel/realsense/d4xx.c
```diff
@@ -2565,6 +2587,20 @@ enum ds5_sync_mode {
...
+	DS5_D58X_DEVICE_MODE_2C = 1,
+};
+
+enum ds5_d58x_ae_policy {
+	DS5_D58X_AE_POLICY_INDEPENDENT = 0,
```
yalizhanghello: D500_2C_AE_Mode

yalizhanghello: DS5_D58X_AE_POLICY_INDEPENDENT  -> D500_2C_AE_POLICY_AUTO

kernel/realsense/d4xx.c
```diff
@@ -449,6 +449,15 @@ static const struct hwm_cmd cmd_hw_reset = {
 	.opcode = 0x20,  /* HW reset opcode */
 };
 
+#define DS5_HWMC_OPCODE_SET_CAM_SYNC	0x69
+#define DS5_HWMC_OPCODE_GET_CAM_SYNC	0x6A
```
yalizhanghello: better to rename DS5 -》D500 for HKR Specific command

yalizhanghello (changes requested):

kernel/realsense/d4xx.c
lemon6630: Updated in `ddd73fe`. The PR-added D58x sync and device-mode enum families now use D500 names. Existing driver-wide DS5 identifiers and product-discovery names remain unchanged to avoid unrelated D4xx churn.

kernel/realsense/d4xx.c
lemon6630: Updated in `dcc74bb`. The enum type is now `d500_2c_ae_mode`, using kernel lowercase type naming while making the D500 2C scope explicit.

lemon6630: Updated in `ddd73fe`. The value is now `D500_2C_AE_POLICY_AUTO`; the remaining policy values use the same `D500_2C_AE_POLICY_*` family.

kernel/realsense/d4xx.c
lemon6630: Updated in `ddd73fe`. The HKR-specific HWMC opcodes and custom subcommands now use `D500_*` names. Shared DS5 driver infrastructure remains unchanged.

kernel/realsense/d4xx.c
```diff
@@ -2548,6 +2563,13 @@ enum ds5_sync_mode {
 	DS5_SYNC_MODE_EXTERNAL = 2,
 };
 
+enum d500_sync_mode {
+	D500_SYNC_MODE_DISABLED = 0,
+	D500_SYNC_MODE_RGB_MASTER_UNSUPPORTED = 1,
```
yalizhanghello: only two mode exposed based on latest Xu Control definition 2 & 3. remove 0 1

yalizhanghello (changes requested):

kernel/realsense/d4xx.c
lemon6630: Fixed in `e038e83`. D58x now exposes only the latest XU-defined values 2 (PWM Master) and 3 (External), with min=2, max=3, default=2. The legacy D4xx control remains unchanged at 0/1/2. Full JP6.2.1 build and the D58x range/read tests pass.

yalizhanghello (approved):

yalizhanghello (commented):
seems okay

kernel/realsense/d4xx.c
ymodlin: You are right — I verified `v4l2_ctrl_activate()` in v5.15/v6.8 `v4l2-ctrls-core.c`: it only does `test_and_set_bit`/`test_and_clear_bit` on `ctrl->flags` plus `send_event()`, and takes no mutex. Calling it from `s_ctrl` with the handler lock held is the intended usage. Withdrawing this finding — no change needed.

kernel/realsense/d4xx.c

kernel/realsense/d4xx.c

kernel/realsense/d4xx.c

kernel/realsense/d4xx.c
ymodlin: change approved

kernel/realsense/d4xx.c

kernel/realsense/d4xx.c
ymodlin: Acknowledged — product-scope separation is a fair reason to keep the two helpers. No change needed.

kernel/realsense/d4xx.c
```diff
@@ -2699,6 +2797,251 @@ static int ds5_hwmc_send(struct ds5 *state,
...
+
+	ret = ds5_get_hwmc(state, response, sizeof(response), &response_len);
+	if (ret)
+		return ret;
+	if (response_len < sizeof(response))
+		return -EBADMSG;
+
+	memcpy(&sync_mode_le, response + 4, sizeof(sync_mode_le));
+	*sync_mode = le16_to_cpu(sync_mode_le);
+	if (*sync_mode < D500_SYNC_MODE_PWM_MASTER ||
```
ymodlin: **BUG** — narrowing the *menu* to 2/3 is right, but applying the same bound to the **read** path turns a legitimate FW state into a hard error. `0` (Disabled) is a real value the FW can report (it was the previous menu default). Now:

- `ds5_g_volatile_ctrl` → userspace `G_CTRL` on "Camera Sync Mode" fails with `-ERANGE` whenever FW sits at 0/1 (librealsense XU read fails too).
- `ds5_adjust_sync_mode_control()` swallows the error and forces the cache to `PWM_MASTER`, so `ds5_dev->sync_mode` silently disagrees with the FW.

Suggest keeping 2..3 enforcement on the write path only, and reporting (or clamping with a warn) on read. Note `test_sync_mode_read` asserts `value in (2, 3)`, so it would fail on such a unit — and it has not run on HW yet.

kernel/realsense/d4xx.c
```diff
@@ -4249,6 +4862,59 @@ static const struct v4l2_ctrl_config ds5_ctrl_error_code = {
 	.flags = DS5_READ_ONLY_TELEMETRY_FLAGS,
 };
 
+static const char * const d58x_sync_mode_menu[] = {
```
ymodlin: **BUG (NULL deref)** — dropping indices 0/1 leaves `d58x_sync_mode_menu[0]` and `[1]` as `NULL` holes (designated initializers). Before this commit both were non-NULL strings, so this is new.

The control is created with `ds5_ctrl_sync_mode_d58x` from the *cached* type (line ~5893), but `ds5_adjust_sync_mode_control()` re-reads `DS5_DEVICE_TYPE` and only falls back to cache when it reads exactly 0. Any other unmapped value hits the `default:` branch (line ~7647):

```c
__v4l2_ctrl_modify_range(state->ctrls.sync_mode, 0, 0, 0, 0);
```

qmenu stays `d58x_sync_mode_menu`, so value 0 is now in range with `menu_skip_mask == 0`, and `std_validate()` does `ctrl->qmenu[0][0]` with no NULL check -> oops. (`v4l2_querymenu()` guards this, `std_validate()` does not.)

Suggest resetting `qmenu`/range to the legacy `sync_mode_menu` in the `default:` branch, or keeping non-NULL placeholder strings at 0/1 with `menu_skip_mask` covering them.

test/v4l2_test/tests/test_controls.py
```diff
@@ -515,6 +528,223 @@ def test_sync_mode_default_after_reset(self, depth_device):
...
+            ):
+                write_int_control(
+                    depth_device,
+                    C.DS5_CAMERA_CID_DUAL_RGB_AE_POLICY,
+                    requested,
+                )
+                readback = read_int_control(
+                    depth_device, C.DS5_CAMERA_CID_DUAL_RGB_AE_POLICY
+                )
+                assert readback == expected
```
ymodlin: **BUG (test)** — this asserts *clamping*, but the control is `V4L2_CTRL_TYPE_MENU`, and `std_validate()` returns `-ERANGE` for out-of-range menu values; only `V4L2_CTRL_TYPE_INTEGER` clamps (`ROUND_TO_RANGE`).

`write_int_control()` -> `device.set_ctrl()` does not swallow errors, so the first write (`-1`) raises `OSError`/`ERANGE` and the test errors out before reaching the assert — it can never pass on HW.

Suggest asserting rejection instead:
```python
with pytest.raises(OSError):
    write_int_control(depth_device, C.DS5_CAMERA_CID_DUAL_RGB_AE_POLICY, requested)
```
and that the value is unchanged afterwards. (The `finally` restore then also becomes a no-op write of `original`.)

kernel/realsense/d4xx.c

<sub>Auto-generated by the **rs-agentic** bot.</sub>

kernel/realsense/d4xx.c

<sub>Auto-generated by the **rs-agentic** bot.</sub>

<sub>Auto-generated by the **rs-agentic** bot.</sub>

kernel/realsense/d4xx.c

<sub>Auto-generated by the **rs-agentic** bot.</sub>

kernel/realsense/d4xx.c

<sub>Auto-generated by the **rs-agentic** bot.</sub>

ymodlin: Scope note: the above approves the **menu narrowing to 2/3** requested here. It does not cover the rest of `e038e83` — 3 open findings on that commit are still pending:

- `d4xx.c:3026` — read path returns `-ERANGE` on FW-reported 0/1
- `d4xx.c:4865` — `default:` branch leaves the D58x qmenu with NULL holes at 0/1
- `test_controls.py:692` — asserts clamping, but MENU controls reject with `-ERANGE`

<sub>Auto-generated by the **rs-agentic** bot.</sub>

wangdi3-intel (approved):

kernel/realsense/d4xx.c
```diff
@@ -449,6 +449,15 @@ static const struct hwm_cmd cmd_hw_reset = {
 	.opcode = 0x20,  /* HW reset opcode */
 };
 
+#define D500_HWMC_OPCODE_SET_CAM_SYNC	0x69
```
ejgoldik: External sync should not be HWMC, it is an XU command. Please see PR https://github.com/realsenseai/realsense_mipi_platform_driver/pull/586

ejgoldik (changes requested):
Remove ext sync HWMC

ymodlin (changes requested):
the sync mode is implemented by XU command and was done by Ehud on another PR.
please remove any code related to sync mode.

kernel/realsense/d4xx.c
lemon6630: Superseded by cd71495. Per the latest scope decision, all PR-introduced D58x sync-mode changes were removed. The final PR no longer carries the 2/3 menu or related read/qmenu behavior; external sync is handled by PR #586.

kernel/realsense/d4xx.c
lemon6630: Addressed by scope removal in cd71495. The D58x HWMC sync GET path, cache, and read validation were removed, so this ERANGE case no longer exists in the PR. External sync is handled by PR #586.

kernel/realsense/d4xx.c
lemon6630: Addressed by scope removal in cd71495. The D58x sync menu/config and qmenu switching were removed, so the NULL-hole path no longer exists in this PR.

test/v4l2_test/tests/test_controls.py
lemon6630: Addressed by scope removal in cd71495. All test/v4l2_test changes were restored to dev. MIPI driver CI is maintained in ds5 fw-tests, so this test no longer exists in the PR.

kernel/realsense/d4xx.c
lemon6630: Addressed in cd71495. Removed all PR-introduced D58x external-sync HWMC opcodes/helpers and the SerDes FSYNC/ESYNC/tunneling lifecycle. PR #586 remains the XU-based solution. The final PR only adds Device Mode and 2C AE Policy.

lemon6630:
@ymodlin @ejgoldik Addressed the latest scope requests in cd71495.

- Removed all PR-introduced D58x sync/HWMC/FSYNC/ESYNC/tunneling changes; PR #586 remains the XU-based external-sync solution.
- Restored every test/v4l2_test file to dev; MIPI driver CI remains in ds5 fw-tests.
- The final PR diff now contains only kernel/realsense/d4xx.c for Device Mode and 2C AE Policy.
- Full clean JP6.2.1 build passed; strict checkpatch reports 0 errors, 0 warnings, and 0 checks.

All related review threads have been updated and resolved. Please re-review the current head.

kernel/realsense/d4xx.c
```diff
@@ -2699,6 +2789,184 @@ static int ds5_hwmc_send(struct ds5 *state,
 	return 0;
 }
 
+static int ds5_hwmc_send_wait(struct ds5 *state, const struct hwm_cmd *cmd)
```
ymodlin: Is this function for d5xx only? if so let's align that if a function is for d5xx then it will not start with ds5 rather with d5xx

kernel/realsense/d4xx.c
```diff
@@ -2699,6 +2789,184 @@ static int ds5_hwmc_send(struct ds5 *state,
...
+/* Caller must hold state->ds5_dev->lock. */
+static bool ds5_camera_is_idle_locked(struct ds5 *state)
+{
+	return !state->ds5_dev->depth_streaming &&
+	       !state->ds5_dev->rgb_streaming &&
+	       !state->ds5_dev->ir_streaming &&
+	       !state->ds5_dev->imu_streaming;
+}
+
+static int ds5_get_d58x_device_mode(struct ds5 *state, u32 *mode)
```
ymodlin: same here if it's for d5xx only let's align to start with d5xx and not ds5

ymodlin (commented):
Looks good to me. @ejgoldik please review and approve

ymodlin (dismissed):

kernel/realsense/d4xx.c
```diff
@@ -2699,6 +2789,184 @@ static int ds5_hwmc_send(struct ds5 *state,
 	return 0;
 }
 
+static int d5xx_hwmc_send_wait(struct ds5 *state, const struct hwm_cmd *cmd)
```
ejgoldik: Why do we need all of these new hwmc apis?
(send_wait, get_u32, get_custom_u32, set_custom_u32)

ejgoldik (changes requested):

ymodlin:
I didnt notice this before but why are you changing the HWMC??? 
Add typed HWMC SET/GET helpers with exact response validation.

Why this is relevant to the PR. the PR title is to expose the GMSL control then you are adding a new way to send HWMC only for d5xx cameras....

ymodlin:
This as well:
Map both signed and unsigned firmware ErrorTypes to Linux errno.
Keep generic HWMC and dedicated V4L2 control state coherent.
How this is related to changing the GMSL controls...

ejgoldik (approved):

PR #585, opened 2026-08-14
[RSDEV-13159][D585][GMSL] Receive tunnel NV12 over User Defined carrier

polejoe:
## Summary

- Receive D58x tunnel-mode RGB NV12 over the selected CSI-2 User Defined carrier.
- Keep the HKR source format as NV12 (source DT 0x18) and use override DT 0x32 only for the wire packet.
- Return the byte-identical surface through the standard V4L2_PIX_FMT_NV12 ABI without application-side conversion.

## Carrier choice

UD32 is used for the current product route. It is handled as an 8-bit User Defined stream, so every flat NV12 byte remains one transported byte. This avoids the 24bpp behavior discussed for UD30 and keeps tunnel and future Pixel Mode routing on the same product carrier.

## Scope

- D58x tunnel mode only; the Pixel Mode format table is unchanged.
- D4xx format selection is unchanged.
- No serializer/deserializer programming or build-system changes.
- The private media-bus tuple maps the incoming User Defined stream to an opaque T_R8 VI surface, uses H * 3 / 2 transport rows, and reports sizeimage = bytesperline * H * 3 / 2.
- Source DT and wire override remain separate. Selecting another RGB format clears the per-request override.

## Dependencies

- HKR device manager: https://github.com/RealSenseAI-Internal/soc-rs-soc-device-mgr/pull/471
- HKR HIF MIPI: https://github.com/RealSenseAI-Internal/soc-rs-soc-io-drivers/pull/548

## Verification

- Rebased onto dev 1.0.6.3.
- JP6.2.1 and JP6.2.2 patch reset/replay: PASS.
- JP6.2.2 full clean Image, NVIDIA OOT modules, DTB/DTBO, and complete rootfs staging build: PASS.
- d4xx.ko vermagic: 5.15.185-tegra.
- Rootfs package SHA256: `8685d30b5b422b89a751281fb53d04861c8d73d80affed935509c6710562562f`.
- Full rootfs deployed to Jetson67; the unified d4xx.ko owns the D58x nodes.
- NV12 1280x720@60: PASS, including a 1,382,400-byte frame-20 dump and real NV12 decode.
- NV12 1280x960@60: 180/180 frames PASS with 79 decoded frames, zero video sequence drops, and 181/181 valid metadata buffers.
- NV12 848x480, 480x270, and 424x240 at 60 fps: 90/90 frames PASS for each, including real NV12 decode and continuous metadata.
- HKR logs confirm source DT 0x18, override DT 0x32, packet DT 0x32, and FMT_NV12.
- Concurrent Depth Z16, IR Y8I, RGB YUYV, and IMU regression: PASS with zero video sequence drops.
- IR Y16I 1600x1300@15 and RGB RW16 1600x1300@25 regression: PASS.
- No NVCSI/VI timeout, CRC, or frame error was reported during the regression.

## Related

- Pixel Mode work remains separate from this tunnel-only PR.

### DIFF
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -520,7 +520,8 @@ struct ds5_format {
 	unsigned int n_resolutions;
 	const struct ds5_resolution *resolutions;
 	u32 mbus_code;
-	u8 data_type;
+	u8 data_type;		/* Source format DT programmed into HKR. */
+	u8 override_data_type;	/* Non-zero CSI packet DT override on the wire. */
 };
 
 struct ds5_sensor {
@@ -1691,7 +1692,35 @@ static const struct ds5_format ds5_rgb_formats_d58x[] = {
 		.resolutions = d58x_calibration_sizes,
 	},
 };
-#define DS5_D58X_RGB_N_FORMATS 2
+
+#define DS5_D58X_NV12_SOURCE_DT		0x18
+#define DS5_D58X_NV12_CARRIER_DT	0x32
+
+/* Keep shared YUYV and calibration entries aligned with the base table. */
+static const struct ds5_format ds5_rgb_formats_d58x_tunnel[] = {
+	{
+		/* First format: default */
+		.data_type = GMSL_CSI_DT_YUV422_8,	/* UYVY */
+		.mbus_code = MEDIA_BUS_FMT_YUYV8_1X16,
+		.n_resolutions = ARRAY_SIZE(d58x_rgb_sizes),
+		.resolutions = d58x_rgb_sizes,
+	}, {
+		/* Flat NV12 bytes use UD32 as an opaque 8-bit CSI carrier. */
+		.data_type = DS5_D58X_NV12_SOURCE_DT,
+		.override_data_type = DS5_D58X_NV12_CARRIER_DT,
+		.mbus_code = MEDIA_BUS_FMT_RS_NV12_UD32_1X8,
+		.n_resolutions = ARRAY_SIZE(d58x_rgb_sizes),
+		.resolutions = d58x_rgb_sizes,
+	}, {
+		/* RGB calibration: unpacked GRBG10, one little-endian
+		 * 10-bit sample per 16-bit container, 2 bytes/pixel.
+		 */
+		.data_type = GMSL_CSI_DT_RAW_16,
+		.mbus_code = MEDIA_BUS_FMT_SGRBG16_1X16,
+		.n_resolutions = ARRAY_SIZE(d58x_calibration_sizes),
+		.resolutions = d58x_calibration_sizes,
+	},
+};
 
 static const struct ds5_variant ds5_variants[] = {
 	[DS5_DS5U] = {
@@ -2162,6 +2191,14 @@ static void ds5_invalidate_sensor(struct ds5 *state, struct ds5_sensor *sensor)
 	sensor->pipe_vc_id = 0xFFFF;
 }
 
+#ifdef CONFIG_VIDEO_D4XX_SERDES
+static u8 ds5_wire_data_type(const struct ds5_format *format)
+{
+	return format->override_data_type ? format->override_data_type :
+		format->data_type;
+}
+#endif
+
 static int ds5_configure(struct ds5 *state)
 {
 	struct ds5_sensor *sensor;
@@ -2190,10 +2227,12 @@ static int ds5_configure(struct ds5 *state)
 		sensor = &state->rgb.sensor;
 		dt_addr = DS5_RGB_STREAM_DT;
 		md_addr = DS5_RGB_STREAM_MD;
-		/* Only CSI-PT needs the FW DT override; leave the YUYV path on the
-		 * FW default so non-passthrough RGB behaviour is unchanged. */
-		override_addr = sensor->config.format->mbus_code ==
-				MEDIA_BUS_FMT_SBGGR8_1X8 ? DS5_RGB_OVERRIDE : 0;
+		/* D58x tunnel formats always program the stream override contract:
+		 * non-zero selects a carrier DT and zero clears any prior carrier.
+		 * The format cache avoids redundant I2C writes. D4xx is unchanged. */
+		override_addr = (ds5_is_d58x(state) && !state->d58x_pixel_mode) ||
+			sensor->config.format->mbus_code == MEDIA_BUS_FMT_SBGGR8_1X8 ?
+			DS5_RGB_OVERRIDE : 0;
 		fps_addr = DS5_RGB_FPS;
 		width_addr = DS5_RGB_RES_WIDTH;
 		height_addr = DS5_RGB_RES_HEIGHT;
@@ -2220,7 +2259,7 @@ static int ds5_configure(struct ds5 *state)
 	md_fmt = (state->metadata_enabled) ? GMSL_CSI_DT_EMBED : 0x00;
 
 #ifdef CONFIG_VIDEO_D4XX_SERDES
-	data_type1 = sensor->config.format->data_type;
+	data_type1 = ds5_wire_data_type(sensor->config.format);
 	data_type2 = md_fmt;
 	is_calib = (state->is_y8 && (data_type1 == GMSL_CSI_DT_RGB_888));
 
@@ -2325,7 +2364,14 @@ static int ds5_configure(struct ds5 *state)
 	}
 
 	if (override_addr != 0) {
-		dt_value = sensor->config.format->data_type;
+		/* D58x tunnel uses an explicit per-format carrier contract, including
+		 * zero to clear a previous override. Preserve the existing D4xx
+		 * depth/IR override values. */
+		if (state->is_rgb && ds5_is_d58x(state) &&
+		    !state->d58x_pixel_mode)
+			dt_value = sensor->config.format->override_data_type;
+		else
+			dt_value = sensor->config.format->data_type;
 		/* RAW8 CSI passthrough: FW runs in CSI-PT mode (0x2E → dt_addr) and
 		 * remaps all wire DTs to RAW8 via override_addr=0x2A.
 		 */
@@ -6409,6 +6455,12 @@ static int ds5_fixed_configuration(struct i2c_client *client, struct ds5 *state)
 
 	sensor = &state->depth.sensor;
 	dev_type = ds5_dev_type(state, dev_type);
+	state->d58x_pixel_mode = false;
+#ifdef CONFIG_VIDEO_D4XX_SERDES
+	state->d58x_pixel_mode = (dev_type == DS5_DEVICE_TYPE_D58X) &&
+				 (state->dser_ops == &max96712_interface);
+#endif
+
 	switch (dev_type) {
 	case DS5_DEVICE_TYPE_D41X:
 		sensor->formats = ds5_depth_formats_d41x;
@@ -6477,8 +6529,13 @@ static int ds5_fixed_configuration(struct i2c_client *client, struct ds5 *state)
 		sensor->n_formats = DS5_RLT_RGB_N_FORMATS;
 		break;
 	case DS5_DEVICE_TYPE_D58X:
-		sensor->formats = ds5_rgb_formats_d58x;
-		sensor->n_formats = DS5_D58X_RGB_N_FORMATS;
+		if (!state->d58x_pixel_mode) {
+			sensor->formats = ds5_rgb_formats_d58x_tunnel;
+			sensor->n_formats = ARRAY_SIZE(ds5_rgb_formats_d58x_tunnel);
+		} else {
+			sensor->formats = ds5_rgb_formats_d58x;
+			sensor->n_formats = ARRAY_SIZE(ds5_rgb_formats_d58x);
+		}
 		break;
 	default:
 		sensor->formats = &ds5_onsemi_rgb_format;
@@ -6487,13 +6544,6 @@ static int ds5_fixed_configuration(struct i2c_client *client, struct ds5 *state)
 	sensor->mux_pad = DS5_MUX_PAD_RGB;
 
 	sensor = &state->imu.sensor;
-
-	state->d58x_pixel_mode = false;
-#ifdef CONFIG_VIDEO_D4XX_SERDES
-	/* D58x on a MAX96712 deserializer runs the serdes link in PIXEL mode. */
-	state->d58x_pixel_mode = (dev_type == DS5_DEVICE_TYPE_D58X) &&
-				 (state->dser_ops == &max96712_interface);
-#endif
 
 	/* For fimware version starting from: 5.16,
 	   IMU will have 32bit axis values.
--- a/nvidia-oot/6.0/0013-Receive-flat-NV12-over-User-Defined.patch
+++ b/nvidia-oot/6.0/0013-Receive-flat-NV12-over-User-Defined.patch
@@ -1,0 +1,192 @@
+From 168b4e389da3ab461252c9d3be806edfe0bba51e Mon Sep 17 00:00:00 2001
+From: zhou <shengyong.zhou@realsenseai.com>
+Date: Mon, 17 Aug 2026 10:25:24 +0800
+Subject: [PATCH] media: tegra: receive flat NV12 over UD32
+
+Add a private media-bus tuple that matches UD32 while DMAing an opaque
+T_R8 surface. Apply the RAW8 receive-parser override only to the current
+capture descriptor, expand the transport height to H*3/2, and expose the
+resulting byte-identical surface through the standard NV12 V4L2 ABI.
+
+Scope NV12 buffer sizing and channel-format lookup to the private tuple,
+while preserving legacy lookup behavior for every other FourCC.
+---
+ .../media/platform/tegra/camera/vi/channel.c  |  9 +++++++++
+ drivers/media/platform/tegra/camera/vi/core.c | 16 ++++++++++++++--
+ .../media/platform/tegra/camera/vi/vi5_fops.c | 12 +++++++
+ .../platform/tegra/camera/vi/vi5_formats.h    | 32 +++++++++++++++++++
+ 4 files changed, 67 insertions(+), 2 deletions(-)
+
+diff --git a/drivers/media/platform/tegra/camera/vi/channel.c b/drivers/media/platform/tegra/camera/vi/channel.c
+index ac5e0e9c..9eee67ba 100644
+--- a/drivers/media/platform/tegra/camera/vi/channel.c
++++ b/drivers/media/platform/tegra/camera/vi/channel.c
+@@ -295,6 +295,9 @@ static void tegra_channel_update_format(struct tegra_channel *chan,
+ 
+ 	if (fourcc == V4L2_PIX_FMT_NV16)
+ 		chan->format.sizeimage *= 2;
++	else if (fourcc == V4L2_PIX_FMT_NV12 &&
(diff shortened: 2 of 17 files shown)


nvidia-oot/6.2/0013-Receive-flat-NV12-over-UD30.patch
```diff
@@ -0,0 +1,162 @@
...
++		chan->format.sizeimage = chan->format.sizeimage * 3 / 2;
+ }
+ 
+ static void tegra_channel_fmts_bitmap_init(struct tegra_channel *chan)
+@@ -2129,8 +2131,10 @@ __tegra_channel_try_format(struct tegra_channel *chan,
+ 				&pix->width, &pix->height, &pix->bytesperline);
+ 	pix->sizeimage = get_aligned_buffer_size(chan,
+ 			pix->bytesperline, pix->height);
+-	if (chan->fmtinfo->fourcc == V4L2_PIX_FMT_NV16)
++	if (vfmt->fourcc == V4L2_PIX_FMT_NV16)
```
ymodlin: **Potential BUG (NULL deref)** — switching from `chan->fmtinfo->fourcc` to `vfmt->fourcc` now dereferences a pointer that the new `test_bit(fmts_bitmap)` gate in `tegra_core_get_format_by_fourcc()` can make NULL. `chan->fmtinfo` was never NULL here; `vfmt` can be. Please confirm this path is NULL-guarded.

nvidia-oot/6.2/0013-Receive-flat-NV12-over-UD30.patch
```diff
@@ -0,0 +1,162 @@
...
+index f0bde35a..33396ad1 100644
+--- a/drivers/media/platform/tegra/camera/vi/core.c
++++ b/drivers/media/platform/tegra/camera/vi/core.c
+@@ -96,7 +96,8 @@ int tegra_core_get_code_by_fourcc(struct tegra_channel *chan,
+ 	unsigned int i;
+ 
+ 	for (i = offset; i < chan->num_video_formats; ++i) {
+-		if (chan->video_formats[i]->fourcc == fourcc)
++		if (chan->video_formats[i]->fourcc == fourcc &&
++		    test_bit(i, chan->fmts_bitmap))
```
ymodlin: **Potential BUG (scope)** — this gate is global: it changes `tegra_core_get_code_by_fourcc()` / `get_format_by_fourcc()` for every VI channel and sensor, not just the D58x tunnel. Any caller that runs before `tegra_channel_fmts_bitmap_init()` populates `fmts_bitmap` now gets NULL/`-EINVAL` where it previously resolved. Worth confirming every call site is post-init.

nvidia-oot/6.2/0013-Receive-flat-NV12-over-UD30.patch
```diff
@@ -0,0 +1,162 @@
...
+
+diff --git a/drivers/media/platform/tegra/camera/vi/channel.c b/drivers/media/platform/tegra/camera/vi/channel.c
+index 83946acb..0bdaba7b 100644
+--- a/drivers/media/platform/tegra/camera/vi/channel.c
++++ b/drivers/media/platform/tegra/camera/vi/channel.c
+@@ -295,6 +295,8 @@ static void tegra_channel_update_format(struct tegra_channel *chan,
+ 
+ 	if (fourcc == V4L2_PIX_FMT_NV16)
+ 		chan->format.sizeimage *= 2;
++	else if (fourcc == V4L2_PIX_FMT_NV12)
```
ymodlin: **Guideline violation (CLAUDE.md, as amended by this very PR)** — *"Never enable a persistent/global override or select it from V4L2 FourCC alone."* This scales `sizeimage` off `V4L2_PIX_FMT_NV12` alone, while `vi5_setup_surface()` correctly keys on `chan->fmtinfo->code`. Suggest using `tegra_vi_get_surface_override(chan->fmtinfo->code)` here too, so the 1.5x stays tied to the private tuple and a future real NV12 sensor can't be mis-sized.

nvidia-oot/6.0/0013-Receive-flat-NV12-over-UD30.patch
```diff
@@ -0,0 +1 @@
+../6.2/0013-Receive-flat-NV12-over-UD30.patch
```
ymodlin: **Guideline violation (CLAUDE.md)** — *"`6.0/` holds the real files; `6.1/`/`6.2/` symlink the byte-identical ones back to `6.0/`"*. Here `6.0/` and `6.1/` symlink **forward** to `6.2/`, inverting the documented convention. Either move the real file to `6.0/`, or update CLAUDE.md (this PR already edits it).

kernel/realsense/d4xx.c
```diff
@@ -110,12 +110,25 @@ struct ser_interface {
...
 #define GMSL_CSI_DT_RAW_8 0x2A
 #define GMSL_CSI_DT_EMBED 0x12
 #endif
 
+#ifndef GMSL_CSI_DT_YUV420_8
+#define GMSL_CSI_DT_YUV420_8 0x18
+#endif
+
+#ifndef GMSL_CSI_DT_USER_1
+#define GMSL_CSI_DT_USER_1 0x30
```
ymodlin: **Enhancement** — `GMSL_CSI_DT_USER_1 0x30` duplicates `GMSL_CSI_DT_UED_U1 0x30`, which `gmsl-link.h` already provides. Also `GMSL_CSI_DT_YUV420_8` is now defined twice (line 113 in the `#else` block plus this unconditional `#ifndef`). Reuse the existing name and drop the duplicate define.

kernel/realsense/d4xx.c
```diff
@@ -110,12 +110,25 @@ struct ser_interface {
...
+#endif
+
+#ifndef MEDIA_BUS_FMT_RS_NV12_UD30_1X8
+#define MEDIA_BUS_FMT_RS_NV12_UD30_1X8 0x5008
```
ymodlin: **Enhancement** — `0x5008` is defined in three independent places (here, `6.2/0013`, `7.0/0013`) with no shared header. Editing one silently desyncs d4xx's advertised mbus code from VI's match table, surfacing as a confusing "format not found". Define it once. Note `0x50xx` is also upstream's vendor-specific range — future collision risk.

kernel/realsense/d4xx.c
```diff
@@ -1691,7 +1707,31 @@ static const struct ds5_format ds5_rgb_formats_d58x[] = {
 		.resolutions = d58x_calibration_sizes,
 	},
 };
-#define DS5_D58X_RGB_N_FORMATS 2
+
+static const struct ds5_format ds5_rgb_formats_d58x_tunnel[] = {
```
ymodlin, line 1700: **Enhancement** — entries 0 (YUYV) and 2 (RGB calibration) are verbatim copies from `ds5_rgb_formats_d58x[]`, so every future D58x RGB format change has to be made twice. Consider deriving the tunnel table from the base one, or add a short keep-in-sync note.

ymodlin (commented):
**rs-agentic automated review** — D58x tunnel NV12 over UD30.

The DT split (`data_type` = HKR source, `override_data_type` = wire carrier) is a clean contract, and `ds5_wire_data_type()` is a no-op for every pre-existing format, so the d4xx blast radius is well contained. `DS5_D58X_RGB_N_FORMATS` → `ARRAY_SIZE` is a good cleanup, and `cached_override_value` invalidating to `0xFFFF` means the "explicit zero clears a stale carrier" contract genuinely holds.

Main concerns are in the VI carrier — two changes are **global** to `tegra_core_get_*_by_fourcc()` rather than scoped to the tunnel path.

3 items to confirm, 4 minor. Not an approval.

nvidia-oot/6.2/0013-Receive-flat-NV12-over-UD30.patch
polejoe: Confirmed this path is already NULL-guarded. `__tegra_channel_try_format()` first looks up the requested format, falls back to `chan->format.pixelformat` if needed, and returns `-EINVAL` if that lookup also fails. The `vfmt->fourcc` size calculation is only reached after those checks, so no code change is needed for this item.

nvidia-oot/6.2/0013-Receive-flat-NV12-over-UD30.patch
polejoe: Fixed in 16ccca4. Channel bitmap filtering is now limited to the new NV12 lookup. An empty bitmap keeps the pre-init lookup behavior, and every non-NV12 FourCC follows the original unconditional lookup path.

nvidia-oot/6.2/0013-Receive-flat-NV12-over-UD30.patch
polejoe: Fixed in 16ccca4. Both active-format sizing and `TRY_FMT` now require the full private tuple: `V4L2_PIX_FMT_NV12` plus `MEDIA_BUS_FMT_RS_NV12_UD30_1X8`. A normal NV12 tuple will not receive the 3/2 size rule.

nvidia-oot/6.0/0013-Receive-flat-NV12-over-UD30.patch
polejoe: Fixed in 16ccca4. JP6.0 now contains the real patch; JP6.1 and JP6.2 are symlinks back to `../6.0/0013-Receive-flat-NV12-over-UD30.patch`, matching the repository convention.

kernel/realsense/d4xx.c
polejoe: Fixed in 16ccca4. The tunnel LUT now reuses `GMSL_CSI_DT_UED_U1`; the duplicate USER_1 and YUV420 definitions were removed. The only local media-bus fallback left is guarded for JP5 headers that predate the shared definition.

kernel/realsense/d4xx.c
polejoe: Fixed in 16ccca4. `MEDIA_BUS_FMT_RS_NV12_UD30_1X8` is defined once in the patched kernel UAPI `media-bus-format.h` for JP6/JP7, and both d4xx and VI consume that definition. JP5 keeps a guarded compatibility fallback because its kernel headers do not contain the new code.

kernel/realsense/d4xx.c
polejoe: Addressed in 16ccca4 with a short keep-in-sync note beside the tunnel table. The entries stay explicit because the tunnel table inserts the NV12 carrier entry between shared D58x formats; deriving a second static initializer from the first C array would make the mapping less direct.

polejoe:
Review follow-up is in 16ccca4.

What changed:
- tied NV12 sizeimage and TRY_FMT handling to the full private NV12/UD30 media-bus tuple
- limited channel bitmap lookup filtering to NV12; pre-init lookup and every non-NV12 FourCC keep the old behavior
- moved the shared media-bus code into the patched kernel UAPI
- restored the JP6.0 real-file / JP6.1+6.2 symlink convention
- reused the existing UD30 constant and added the D58x LUT keep-in-sync note

The NULL-dereference report was checked and does not reproduce in the control flow: both requested and fallback vfmt lookups are tested, with -EINVAL returned before vfmt is dereferenced.

The S5C and AHSV defines visible around the UAPI change are existing context lines, not additions. The only new format define in that hunk is MEDIA_BUS_FMT_RS_NV12_UD30_1X8.

Verification:
- JP6 and JP7 clean patch apply: PASS
- JP6.2.2 kernel and NVIDIA OOT build/package: PASS
- board-side Depth Z16, IR Y8I and IMU GREY direct profiles: PASS
- full post-review board regression paused when the rig was released; no further device test is claimed

polejoe:
Post-review board regression is complete on Jetson67.

- Deployed the complete JP6.2.2 rootfs from this branch; unified `d4xx.ko 1.0.6.2` owns the D58x nodes. Legacy `d5xx.ko` was not used.
- HKR ran an isolated latest-master SRAM image. The stream log confirms `data_type=0x18`, `override_dt=0x30`, `packet_dt=0x30`, `source_fmt=NV12`.
- NV12 1280x960@60: 121 captured / 51 decoded, zero errors and sequence drops. Frame 20 is 1,843,200 bytes with BPL 1280.
- NV12 1280x720@60: 121 captured / 68 decoded, zero errors and sequence drops. Frame 20 is 1,382,400 bytes.
- NV12 with RGB metadata: 61 video frames and 62 valid metadata buffers; zero sequence gaps, parse errors or stream mismatches.
- Regression passed for RGB YUYV, Depth Z16, IR Y8I, IR Y16I, RGB RW16 and IMU GREY.

The earlier probe failure was caused by mixing the legacy `d5xx` overlay/module path with `dev`. Re-testing with the repository's unified D58x tunnel overlay and `d4xx.ko` passed. No SER/DES source or D5x firmware changes were made for this PR.

kernel/realsense/d4xx.c
```diff
@@ -607,6 +622,8 @@ struct ds5 {
 	/* D58x camera behind a PIXEL-mode deserializer - requires strict
 	 * Scheduling and data sizes in HKR side */
 	bool d58x_pixel_mode;
+	/* D58x camera behind MAX96724 in GMSL tunnel mode. */
+	bool d58x_tunnel_mode;
```
ejgoldik: Why do we need a tunnel_mode flag? we already have a dedicated pixel_mode flag. If we are in the D58x and d58x_pixel_mode == 0 then we are in tunnel mode

ymodlin (changes requested):
Please remove the changes from the scripts folder they are not related to NV12 over UD30

ejgoldik (changes requested):
Main issue I worry about is the fact that UD30 is not usable for Pixel mode at all because it is 32 bit. So if we enable this now we will need two different nv12 formats for tunnel and pixel mode. Need to discuss with Evgeny

kernel/realsense/d4xx.c
```diff
@@ -44,6 +44,14 @@
 #include <media/max96717.h>
 #include <media/max96724.h>
 
+#ifndef GMSL_CSI_DT_YUV420_8
+#define GMSL_CSI_DT_YUV420_8 0x18
+#endif
+
```
ejgoldik: Should probably added to media/gmsl-link.h (we patch the rest of the formats there)

ejgoldik (changes requested):

kernel/realsense/d4xx.c
polejoe: Removed the redundant flag. D58x tunnel selection now uses the existing `!state->d58x_pixel_mode` state, and no `d58x_tunnel_mode` member remains.

kernel/realsense/d4xx.c
polejoe: Moved `GMSL_CSI_DT_YUV420_8` into the shared `media/gmsl-link.h` patch carriers. `d4xx.c` now consumes the shared definition.

polejoe:
Updated in `d95400e` based on the latest review:

- Rebuilt the branch as one NV12/UD30 feature commit on `dev`.
- Removed all `scripts/` and build-system changes from the PR.
- Removed the redundant tunnel-mode flag; D58x tunnel mode now uses the existing `!d58x_pixel_mode` state.
- Moved the YUV420 CSI DT definition to the shared `media/gmsl-link.h` patch carriers.
- Kept NV12/UD30 strictly tunnel-only. The D58x pixel-mode format table is unchanged and does not enumerate this format, so the 32-bit UD30 pixel-mode limitation is not exercised by this PR.

The final branch passes a clean JP6.2.2 patch reset/replay and full kernel, NVIDIA OOT module, DTB, and rootfs staging build.

nvidia-oot/6.0/0013-Receive-flat-NV12-over-User-Defined.patch
```diff
@@ -0,0 +1,189 @@
...
++	const struct tegra_vi_surface_override *surface_override =
++		tegra_vi_get_surface_override(chan->fmtinfo->code);
+ 	u32 nvcsi_stream = chan->port[vi_port];
+ 	struct capture_descriptor_memoryinfo *desc_memoryinfo =
+ 		&chan->tegra_vi_channel[vi_port]->
+@@ -413,6 +417,9 @@ static void vi5_setup_surface(struct tegra_channel *chan,
+ 	 */
+ 	if (chan->fmtinfo->fourcc == V4L2_PIX_FMT_Y16I)
+ 		width *= 2;
++	if (surface_override)
```
ev-mp, line 123: This section is the upstream caveat we previously discussed

nvidia-oot/6.0/0013-Receive-flat-NV12-over-UD31.patch
```diff
@@ -0,0 +1,189 @@
...
+ 
+ static void tegra_channel_fmts_bitmap_init(struct tegra_channel *chan)
+@@ -2126,8 +2129,11 @@ __tegra_channel_try_format(struct tegra_channel *chan,
+ 				&pix->width, &pix->height, &pix->bytesperline);
+ 	pix->sizeimage = get_aligned_buffer_size(chan,
+ 			pix->bytesperline, pix->height);
+-	if (chan->fmtinfo->fourcc == V4L2_PIX_FMT_NV16)
++	if (vfmt->fourcc == V4L2_PIX_FMT_NV16)
+ 		pix->sizeimage *= 2;
++	else if (vfmt->fourcc == V4L2_PIX_FMT_NV12 &&
```
ev-mp: This the injection that contaminates vi/nvcsi subsytem with RS-specific logic. Lets continue offline

polejoe:
UD32 tunnel validation is complete on Jetson67 with the full JP6.2.2 rootfs from this branch.

- 1280x720@60 and 1280x960@60 NV12 passed; frame 20 at 1280x720 is 1,382,400 bytes.
- 848x480, 480x270, and 424x240 at 60 fps passed real NV12 decode with continuous metadata.
- HKR logs confirm source DT 0x18, wire override/packet DT 0x32, and FMT_NV12.
- Concurrent Depth Z16, IR Y8I, RGB YUYV, and IMU passed with zero video sequence drops.
- IR Y16I@15 and RGB RW16@25 regression passed.
- No NVCSI/VI timeout, CRC, or frame error was reported during these runs.

The latest commit also fixes the JP7 feature patch hunk count exposed by CI; it changes patch metadata only, not the verified driver logic.

polejoe:
Patch carrier names are now independent of the selected User Defined DT number across JP5, JP6, and JP7. The protocol implementation still explicitly selects DT 0x32.

- JP6.2.2 patch reset/replay with the renamed carriers: PASS
- Full clean Image/OOT/DTB/rootfs build: PASS
- Rootfs SHA256: `8685d30b5b422b89a751281fb53d04861c8d73d80affed935509c6710562562f`
- Full rootfs deployed to Jetson67 and rebooted: PASS
- Unified `d4xx.ko 1.0.6.3` is loaded; all seven D58x video/metadata aliases are present
- System is running with no post-deploy VI/NVCSI error

polejoe:
This PR has been retargeted from `dev` to `dev_gmsl_merge`.

`dev_gmsl_merge` was forked from `dev` at `384f3688a5d3f2fa5a420b9d422dc7feeba52a8c` (version 1.0.6.3). It is the current integration branch for ready-for-review GMSL stack changes. The historical `d500_gmsl_dev` branch is tunnel-only and has diverged too far from `dev` to remain an efficient shared base.

This staging flow keeps each change independently reviewable while surfacing cross-PR and current-`dev` conflicts earlier. It also allows combined regression and full validation before promotion to `dev`, reducing repeated conflict resolution, duplicated validation, and overall merge effort.

PR #589, opened 2026-08-17
Enhance D500 reset recovery for serializer power loss

pinquanw:
## Summary

- serialize reset recovery across the I2C functions of one physical camera
- detect D500 before reset invalidates the cached device type
- after a D500 HW reset, leave the GMSL link untouched for a 5-second HKR/serializer boot window
- asynchronously reproduce the complete dependency lifecycle for the affected deserializer: remove d4xx peers before primaries, release serializer and deserializer devices, then attach deserializer, serializer, Depth, RGB, IR and IMU in order
- cancel pending recovery work safely during module removal
- remove the obsolete runtime dependency on max96724_recover_link
- keep the D400 reset path unchanged
- define the RAW16 CSI data type missing from NVIDIA JP6.2 headers

## Root cause

D58X HW reset causes HKR to power down the remote GMSL serializer. Host-side d4xx instances and SerDes driver state must be rebuilt after HKR restores remote power.

Hardware validation also exposed an HKR-side requirement: current HKR FW logs GMSL power disabled, negotiation state 0 and auto-deinit complete after HW reset, but does not automatically re-enable GPIO_81. The Orin has no HKR UART or reset-GPIO sideband, so host-driver re-probe alone cannot recover a serializer that remains physically powered off. HKR FW must auto-init GMSL after reset, or the platform must provide a UART/GPIO sideband reset.

## Validation

- JP6.2 focused NVIDIA OOT module build with the L4T cross compiler: passed
- checkpatch on the final d4xx diff: 0 errors, 0 warnings
- D500 reset with host-only rebind: expected failure while HKR leaves GMSL power disabled; MAX96724 link A remains 0x16 and times out
- D500 reset plus HKR UART reboot plus automatic driver rebind: passed twice without rebooting Orin
- after recovery: one D585 enumerated, all 4 d4xx I2C functions bound, 7 video nodes
- post-recovery stream probe: 90 framesets, valid Depth and Color frames, exit 0

## Platform requirement

This PR provides the host re-enumeration half of recovery. Autonomous recovery additionally requires HKR FW to restart its GMSL state machine, or a platform sideband capable of rebooting HKR.

### DIFF
--- a/kernel/realsense/d4xx.c
+++ b/kernel/realsense/d4xx.c
@@ -31,6 +31,7 @@
 #include <linux/videodev2.h>
 #include <linux/version.h>
 #include <linux/mutex.h>
+#include <linux/workqueue.h>
 #include <media/media-entity.h>
 #include <media/v4l2-ctrls.h>
 #include <media/v4l2-device.h>
@@ -633,6 +634,10 @@ struct ds5 {
 	struct ds5_dev *ds5_dev; /* pointer to DS5 device struct */
 };
 
+#ifdef CONFIG_VIDEO_D4XX_SERDES
+#define DS5_REENUM_ROLE_COUNT 4
+#endif
+
 struct ds5_dev {
 	struct mutex lock;
 	/*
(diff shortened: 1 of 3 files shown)


pinquanw:
Superseded by #606, which reduces the recovery to a single d4xx.c change and no longer unbinds or modifies the serializer/deserializer drivers.

PR #601, opened 2026-08-19
Host-POC_vc_mux: dev_gmsl: [D5x][GMSL] dual-RGB VC1 mux on Host side

polejoe:
## Summary

Host-side proof of concept for D585 Dual-RGB when Left MC2 and Right MC4 arrive on the same CSI VC1 with their native DT1E unchanged. Two normal V4L2 video nodes and two metadata sidecars remain visible to userspace; applications do not need to understand the VC mux.

## Implementation

- Configure MC2 and MC4 endpoints as a shared-VC topology through device-tree source IDs.
- Keep MC2 as the zero-copy physical VI owner.
- Route each completed frame by the frame-bound GMSL source ID in embedded metadata.
- Deliver MC4 frames to the second standard V4L2 node with one GPCDMA DMA-BUF copy and no CPU full-frame copy.
- Fan out the paired metadata buffer to the matching metadata node.
- Support either video node starting first, independent secondary stop while the owner remains active, format validation, and complete VB2 buffer return on every mux-specific start failure.
- Emit stream-off counters for both sources, dropped/invalid metadata, and DMA errors.

## Validation

- JetPack 6.2.2 full build passed; the deployed Image, DTBO, and tegra-camera module match the local build.
- D585 was verified in 2C Dual-RGB mode with the shared-VC HKR image.
- Secondary-first 1280x960 YUYV at 30 FPS passed: secondary completed 120 frames, then stopped while owner continued to 240 frames; invalid metadata and DMA errors stayed zero.
- The former secondary-first format-mismatch panic path passed: secondary returned EIO cleanly, owner completed 120 frames, and no VB2 warning or panic occurred. A following same-format dual run passed without reboot.
- Existing v4l2_preview.py passed with two RGB video nodes and two metadata nodes:
  - capture-only: RGB 129/129 good, Right 122/122 good, metadata valid with no stream mismatch;
  - YUYV decode: RGB 92/92 good, Right 91/91 good, both application streams PASS.
- HKR stream-off statistics showed independent HIF MIPI stream 0 and stream 1 req/done completion.

## Known POC scope and remaining work

- This patch implements one owner plus one secondary. Generalizing the topology to the product limit of up to four MC sources sharing one VC remains separate work.
- 1600x1300 at 25 FPS is not claimed for the real MC2/MC4 Dual-RGB case: the current HKR image reports an MC2 static graph configuration failure before Host capture can be evaluated.
- On the current Jetson image, generic enumeration creates color-0 for MC4 and color-1 for MC2. Both nodes work, but stable left/right alias semantics still need to be defined; validation used explicit video and metadata node paths.
- GR0P enumeration produces an existing V4L2 Unknown pixelformat warning and is unrelated to this mux path.

## D5x PR set

HKR runtime must be built with the generic `BUILD_HIF_MIPI_MC2VC_MUX=ON` route policy. The option currently selects D585 MC2/MC4 -> VC1 but is not limited to Dual-RGB, VC1, or exactly two MCs.

Depends-On: https://github.com/RealSenseAI-Internal/soc-rs-hkr-apps/pull/297
Depends-On: https://github.com/RealSenseAI-Internal/soc-rs-hkr-test/pull/376
Depends-On: https://github.com/RealSenseAI-Internal/soc-rs-hkr-trunk/pull/135
Depends-On: https://github.com/RealSenseAI-Internal/soc-rs-soc-device-mgr/pull/475
Depends-On: https://github.com/RealSenseAI-Internal/soc-rs-soc-io-drivers/pull/556

### DIFF
--- a/hardware/realsense/tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dts
+++ b/hardware/realsense/tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dts
@@ -94,6 +94,8 @@
 							vc-id = <1>;
 							port-index = <2>;
 							bus-width = <4>;
+							realsense,d5xx-vc-mux-source-id = <2>;
+							realsense,d5xx-vc-mux-owner-channel = <1>;
 							remote-endpoint = <&d5xx_csi_out1>;
 						};
 					};
@@ -133,6 +135,8 @@
 							vc-id = <1>;
 							port-index = <2>;
 							bus-width = <4>;
+							realsense,d5xx-vc-mux-source-id = <4>;
+							realsense,d5xx-vc-mux-owner-channel = <1>;
 							remote-endpoint = <&d5xx_csi_out4>;
 						};
 					};
--- a/hardware/realsense/tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dts
+++ b/hardware/realsense/tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dts
@@ -67,7 +67,7 @@
 		target-path = "/";
 		__overlay__ {
 			tegra-capture-vi {
-				num-channels = <4>;
+				num-channels = <5>;
 				ports {
 					#address-cells = <1>;
 					#size-cells = <0>;
@@ -201,13 +201,22 @@
 							sysfs-device-tree = "/sys/firmware/devicetree/base/bus@0/cam_i2cmux/i2c@1/d5xx@1d";
 						};
 					};
+					cam_module4: module4 {
+						badge = "d5xx_rgb_right";
+						position = "centerright";
+						orientation = "1";
+						cam_module4_drivernode0: drivernode0 {
+							pcl_id = "v4l2_sensor";
+							sysfs-device-tree = "/sys/firmware/devicetree/base/bus@0/cam_i2cmux/i2c@1/d5xx@1e";
+						};
+					};
 				};
 			};
 
 			bus@0 {
 				host1x@13e00000 {
 					nvcsi@15a00000 {
-						num-channels = <4>;
+						num-channels = <5>;
 						#address-cells = <1>;
 						#size-cells = <0>;
 
--- a/hardware/realsense/tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dts
+++ b/hardware/realsense/tegra234-camera-d4xx-overlay-fg24-4ch-d5xx.dts
@@ -14,6 +14,7 @@
  *   VC1 = RGB    (YUY2, 1920x1080 @30fps)
  *   VC2 = IR     (Y8/Y8I/Y12I, 1280x720 @60fps)
  *   VC3 = IMU    (RAW8)
+ *   VC1 = RGB-left and RGB-right on D585 2C shared-VC builds
  *
  * GMSL link:
  *   GMSL2 Tunnel Mode, 6 Gbps
@@ -120,6 +121,19 @@
 							port-index = <2>;
 							bus-width = <4>;
 							remote-endpoint = <&d5xx_csi_out3>;
+						};
+					};
+
+					/* VC1: RGB-right on D585 2C shared-VC builds */
+					port@4 {
+						status = "okay";
+						reg = <4>;
+						d5xx_vi_in4: endpoint {
+							status = "okay";
+							vc-id = <1>;
+							port-index = <2>;
+							bus-width = <4>;
+							remote-endpoint = <&d5xx_csi_out4>;
 						};
 					};
 				};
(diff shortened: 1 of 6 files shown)


polejoe:
POC validation and architecture review update:

- 2C end-to-end preview passed with Depth, IR, RGB Left, and RGB Right exposed as ordinary video nodes. The fanout counters were balanced (`source2=1657`, `source4=1656`) with zero dropped frames, invalid metadata, or DMA errors.
- 3C functional regression passed for Depth + IR + RGB at 640x360@60 for 120 frames, with no drops, capture errors, or metadata gaps.
- However, 3C stream-off still reports the VC-mux owner counters (`source 2=134 source 4=0`). This shows that the static DT topology activates the mux owner path even when the runtime product mode is 3C. Functional regression passes, but the required runtime bypass does not.

The current 849-line fanout patch is a valid feasibility POC, but I do not think its current placement is suitable as the final generic driver design:

1. D5x metadata layout, source parsing, topology state, and lifecycle handling are embedded directly in generic VI files and structures.
2. The model is hard-coded to one secondary channel, while the shared-VC contract allows up to four MC members.
3. `vi5_release_buffer()` waits synchronously for the DMA copy, with a timeout of up to 100 ms, which adds blocking work to the VI frame-completion hot path.
4. `dma_request_channel()` acquires an arbitrary memcpy channel rather than a DT/device-managed resource.
5. Static DT capability is being treated as runtime enablement, which is why the 3C path still performs mux parsing/resource work.

Recommended next revision:

- Keep only a small optional frame-router hook and opaque context in generic Tegra VI: topology bind/unbind, member stream on/off, and owner frame-done.
- Move metadata ABI parsing and runtime 2C capability detection into the D4xx provider.
- Put the generic member array (up to four members), vb2 lifecycle, counters, and asynchronous device-DMA completion in a reusable router component.
- Leave the hook NULL for 3C, non-D5x, and non-mux topologies so the original VI fast path is unchanged.

With two independent standard V4L2 MMAP queues, source identity is only known after capture metadata arrives, so the already-captured owner buffer cannot generally be completed on a different queue without changing the userspace buffer contract. One device-to-device copy for non-owner frames is therefore the practical minimum for the current app-transparent ABI. Strict zero-copy would require a shared/neutral buffer pool, a DMABUF-aware userspace contract, or predictable pre-capture routing.

Two separate HKR issues should not be attributed to this Host fanout:

- Real 1600x1300@25 fails during the HKR IPU/GDF static-graph query because the current PSYS output graph does not contain a 1600x1300 output key. Standalone HifMipi 1600x1300@25 already passes, so this is not a Host, HifMipi, or shared-VC limit.
- The HifMipi shared-profile rejection was triggered by a group build using a new MC2 profile with a stale MC4 profile. The fail-closed check is correct; the coherent profile snapshot/start ordering should be fixed in the HKR group-build transaction.

The PR remains a draft against `dev_gmsl_merge`. Before final review, the open gates are the runtime 2C-only activation, generic hook/provider split, asynchronous copy path, extended lifecycle tests, and the currently failing JetPack 6.2 CI job.

polejoe:
Correction after an isolated 2C RGB-only reproduction:

- The Host requested only RGB Left and RGB Right at 1280x960@30; Depth, IR, and IMU were disabled.
- HKR resolved both profiles as 1280x960@30 and confirmed the shared topology as `MC2:VC1` plus `MC4:VC1`.
- Despite the 30 FPS request, the two HIF logical streams closed at 57.39 and 57.40 FPS, while the Host captured approximately 53 to 57 FPS per node.
- This reproduction did not hit the shared-profile rejection, so the earlier approximately-60-FPS observation must not be attributed to partial-start recovery.

The current working cause is the pre-existing RGB-only product graph in `ACameraSet::buildDualRgbPEIpu()`. That path predates this shared-VC change, selects the OG02B10 full-size sensor mode through a minimum-60-FPS lookup, runs laser-off with no TDM filtering, and passes every PSYS frame downstream. The configured HIF sink FPS is not an upstream cadence limiter. The shared-VC change only changes the effective VC and shared-route lifecycle; it does not duplicate frames.

The mixed Depth/IR + Dual-RGB TDM graph remains different: its sensor cadence is doubled and each HIF logical stream correctly measured approximately 30 FPS in the four-stream run.

Therefore the RGB-only 30-FPS contract violation should be fixed in the D5x product/IPU graph cadence path, not by dropping frames in the Host fanout, HifMipi, or HIF firmware. One `ImagingUnit: Run AIQ fail. ret = -4` message was also present during the isolated run and remains a separate HKR-side startup issue.

polejoe:
Final correction for the 2C RGB-only 30 FPS investigation:

The ~57-60 FPS reported by both logical HIF streams is real task cadence, but the final root cause is more specific than my previous comment. GMSL pre-opens the four 2C group members, with inactive Depth/IR entries marked `start_gated=1`. `ACameraSet::buildPE()` uses `hasDepthStream()` / `hasDualRgbStream()`, which count those gated entries, so an RGB-Left + RGB-Right-only session selects `buildDualRgbSdpPEIpu()` (the mixed Depth+DualRGB graph), not `buildDualRgbPEIpu()`.

The mixed builder converts the active 30 FPS contract to `mFps=60` and configures both RGB PSYS nodes at 60 FPS. Runtime scenario selection, however, ignores gated Depth/IR and correctly selects RGB-only strobe `0x00`. With `processIlluminated=0`, every laser-off frame is non-illuminated, so the illumination filter passes all 60 frames instead of halving them to 30. HIF's configured `fps=30` is profile/bandwidth configuration, not a cadence limiter; each request is emitted, hence the measured ~60 FPS.

This is an HKR `apps/d500` graph/scenario coordination bug, not shared-VC duplication, HIF MIPI/FW behavior, MC2VC arbitration, or Host fan-out. The fix belongs in the HKR product graph/IPU cadence policy. PR #601 must not add a Host-side frame drop to hide it. The normal four-stream `strobe=0xAA` case still measures 30 FPS per output because illumination filtering halves the 60 FPS sensor cadence as intended.

polejoe:
Close this POC for 2 video node fan-out in kernel driver.
dual-rgb host v4l2 keep one aggregate video node, and demux in app/sdk by metadata

PR #603, opened 2026-08-19
dev: [RSDEV-13503][RSDEV-13505][D585-MS4][GMSL] Add tunnel RGB-right on VC4

polejoe:
## Summary

Port the D5x tunnel-mode RGB-right support from merged PR #599 onto `dev`.

- Add the D585 2C RGB-right control path on CSI VC4 while preserving the existing VC0 Depth, VC1 RGB, VC2 IR, and VC3 IMU mapping.

- Gate RGB-right by the D585 2C product PID, cap 2C image profiles at 30 FPS, and append the VC4 VI/NVCSI endpoint only to the FG24 tunnel overlay.

- Allow the MAX96724 Link-A tunnel to share one physical pipe across logical VCs with reference-counted lifetime; the MAX96712 path, pixel-mode overlays, and generic VI/NVCSI implementation are unchanged.

## Integration status

D5x device-side patches are already merged. This host patch was checked on `dev_gmsl_merge`; this PR ports the same commit onto `dev` for integration.
## Compatibility

On D585 3C (PID `0x0c08`), `DUALRGB_RIGHT` remains rejected with `-ENODEV`, preserving the existing Depth/RGB/IR/IMU mapping. On D585 2C, RGB-right is appended on VC4; userspace must resolve stable RealSense links or media entities rather than numeric `/dev/videoN` registration order.

## Validation

- The original PR #599 passed a JP6.2.2 full clean build, rootfs packaging, Jetson67 deployment/reboot, module/DTB hash verification, and 3C live-probe coverage.
- This port cherry-picks `6ca0568` onto the current `dev` tip without conflicts; the resulting diff is restricted to the same four source files and passes `git diff --check`.


kernel/realsense/d4xx.c
```diff
@@ -180,6 +186,14 @@ struct ser_interface {
 #define DS5_RGB_CONTROL_STATUS 	0x402E
 #define DS5_RGB_OVERRIDE		0x403C
 
+#define DS5_DUALRGB_RIGHT_STREAM_DT	0x4060
```
ymodlin: Why DS5?

ymodlin (changes requested):
Every function/define that start is for d5xx only should not start with ds5

ejgoldik (changes requested):
Adding vc ids >= 4 require a change in the general driver design. Currently in the driver the vc_id >> 2 is the link id. Meaning that VC id 4 is actually VC id 0 of the second deserializer link. This PR is premature, we first need to redesign this

PR #606, opened 2026-08-24
Recover D585 prototype after hardware reset

pinquanw:
## Summary
- detect the physical camera PID from GVD during probe and cache it per camera module
- route HW reset opcode `0x20` through Host recovery only for D585 prototype PIDs `0x0C07` and `0x0C08`
- treat Depth/RGB/IR/IMU as four functions of one physical D585 prototype module and invalidate/recover them together
- keep media and V4L2 nodes bound during recovery, avoiding librealsense re-enumeration races
- preserve the existing D400 and final D585 reset behavior and leave `ds5_gmsl_serdes_setup()` unchanged

## SerDes recovery

Both supported D585 prototype combinations use the same target-link recovery contract:
- MAX96717 -> MAX96712 pixel mode
- MAX96717 -> MAX96724 tunnel mode

For either deserializer, recovery:
- identifies the physical GMSL link from the target serializer assigned alias (`0x40 + link`)
- temporarily selects only that link so a reset serializer at default address `0x40` cannot collide with another module
- retrains the target link and restores its serializer alias when required
- restores the configured deserializer link mask on success and failure
- reapplies only the target MAX96717 control and initialization settings
- restores HKR MIPI TX runtime configuration after HKR reboot

The shared deserializer is not power-cycled or globally reinitialized. Other physical camera modules are not reset, reconfigured, unbound, or re-enumerated. Their active links can be briefly interrupted while the target default address is isolated.

## Scope
The serializer power-cycle workaround is enabled only for D585 prototype PIDs `0x0C07` and `0x0C08`. Final D585 PIDs `0x0C04`, `0x0C05`, and `0x0C06` remain on the existing reset path. If GVD PID detection fails, the driver does not assume prototype hardware.

This PR does not include RAW16 compatibility or unrelated build changes, add delayed work, unbind the camera driver, or modify librealsense.

## Required HKR change
Use together with RealSenseAI-Internal/soc-rs-soc-device-mgr#484. HKR handles opcode `0x20`, reboots the selected module, power-cycles its serializer, and restarts its GMSL state machine.

## Validation
- current d4xx ARM64 translation unit compiles successfully with `-Werror`
- MAX96712 patches apply cleanly to JP5.0.2 (`jetson_35.1`), JP5.1.2 (`jetson_35.4.1`), and JP6 (`jetson_36.3`) source baselines
- focused current-diff `checkpatch.pl`: 0 errors, 0 warnings
- `git diff --check` and patch parser checks passed
- the recovery sequence was previously validated on single-camera MAX96712 pixel mode and MAX96724 tunnel mode; the new PID gate was not hardware-tested because no board is currently available
- MAX96712/MAX96724 multi-camera hardware regression remains required

## Commit policy
The PR branch contains one commit. Follow-up updates will amend and force-with-lease this commit.

### DIFF
--- a/kernel/nvidia/5.0.2/0017-Adding-max96712-support-for-D4xx.patch
+++ b/kernel/nvidia/5.0.2/0017-Adding-max96712-support-for-D4xx.patch
@@ -7,8 +7,8 @@ Add MAX96712 deserializer driver support for D4XX cameras.
 
 Signed-off-by: ejgoldik <ehud.joseph.goldik@realsenseai.com>
 ---
- drivers/media/i2c/max96712.c | 1170 +++++++++++++++++++++++++++++++++-
- 1 file changed, 1167 insertions(+), 3 deletions(-)
+ drivers/media/i2c/max96712.c | 1222 +++++++++++++++++++++++++++++++++-
+ 1 file changed, 1219 insertions(+), 3 deletions(-)
 
 diff --git a/drivers/media/i2c/max96712.c b/drivers/media/i2c/max96712.c
 index 8081385..b7dcc99 100644
@@ -479,7 +479,7 @@ index 8081385..b7dcc99 100644
  		i2c_unregister_device(client);
  		client = NULL;
  	}
-@@ -252,13 +635,793 @@ max96712_remove(struct i2c_client *client)
+@@ -252,13 +635,845 @@ max96712_remove(struct i2c_client *client)
  	return 0;
  }
  
@@ -711,10 +711,14 @@ index 8081385..b7dcc99 100644
 +}
 +EXPORT_SYMBOL(max96712_get_ser_pipe_id);
 +
-+static void max96712_reset_oneshot_link_mask(struct max96712 *priv, uint8_t link_mask)
++static int max96712_reset_oneshot_link_mask(struct max96712 *priv, uint8_t link_mask)
 +{
-+	max96712_write_reg(priv, MAX96712_CTRL1_ADDR, link_mask);
++	int err = max96712_write_reg(priv, MAX96712_CTRL1_ADDR, link_mask);
++
++	if (err)
++		return err;
 +	msleep(100);
++	return 0;
 +}
 +
 +void max96712_reset_oneshot(struct device *dev)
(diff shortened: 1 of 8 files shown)


kernel/nvidia/5.0.2/0017-Adding-max96712-support-for-D4xx.patch
```diff
@@ -914,10 +921,92 @@ index 8081385f0..aaa115696 100644
...
++	 */
++	err = max96712_write_reg(priv, MAX96712_REG6_ADDR,
++				   0xF0 | BIT(link_id));
++	if (!err) {
++		msleep(20);
++		err = max96712_reset_oneshot_link_mask(priv, BIT(link_id));
++	}
++
++	if (!err) {
++		/* HKR powers the serializer only after its reboot and GMSL startup.
```
ejgoldik: I will soon (hopefully today) have a PR that adds polling to check if the serializer is available, not by reading it's I2C address, but by checking link status regs on the deserializer side. Please hold for this PR and use its check link locked capability instead

ejgoldik (commented):
I will soon (hopefully today) have a PR that adds polling to check if the serializer is available, not by reading it's I2C address, but by checking link status regs on the deserializer side. Please hold for this PR and use its check link locked capability instead

kernel/nvidia/5.0.2/0017-Adding-max96712-support-for-D4xx.patch
```diff
@@ -1177,6 +1181,64 @@ index 8081385..b7dcc99 100644
...
++	unsigned int link_id;
++	u8 target_addr;
++	int restore_err;
++	int err;
++
++	target_addr = ser_client->addr;
++	if (target_addr < MAX9295_DEFAULT_ADDR ||
++	    target_addr >= MAX9295_DEFAULT_ADDR + MAX96712_MAX_LINKS)
++		return -EINVAL;
++	link_id = target_addr - MAX9295_DEFAULT_ADDR;
```
ejgoldik: Link id should be taken directly from the device tree attribute (Should already be stored in the serializer struct)

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
```diff
@@ -1227,6 +1231,64 @@ index 8e237ee..c1e57e2 100644
...
++
++	restore_err = max96712_write_reg(priv, MAX96712_REG6_ADDR,
++					     0xF0 | priv->link_mask);
++	if (!err)
++		err = restore_err;
++	if (!err && priv->link_multi_vc_pipe[link_id] >= 0) {
++		/* The serializer video path was reset. Reapply this link's complete
++		 * multi-VC mapping on the first post-recovery stream start. */
++		priv->multi_vc_configured[
++			priv->link_multi_vc_pipe[link_id]] = false;
```
ejgoldik: Do we really need to reconfigure the deserializer? It didn't lose its configuration, only the serializer did

ejgoldik (changes requested):

kernel/nvidia/5.0.2/0017-Adding-max96712-support-for-D4xx.patch
pinquanw: Agreed. The physical link should come from the `maxim,gmsl-link-id` DT property rather than being inferred from the serializer alias.

In the current driver this property is already cached by d4xx in `state->gmsl_link`; the MAX96717 private context keeps a `gmsl_link_ctx` pointer, but that context does not contain `maxim,gmsl-link-id`. I have therefore passed the cached DT link explicitly to `recover_link()`, and kept the serializer address only as the alias value that must be restored after the power cycle.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
pinquanw: Agreed, this invalidation is unnecessary. MAX96712 remains powered and its pipe selection, VC mapping, and pipe-enable registers are preserved; only the target serializer loses its configuration. The serializer path is restored by `setup_control()`, `init_settings()`, and the normal per-stream `set_pipe()` call.

`multi_vc_configured` should only be cleared when `max96712_init_settings()` actually resets the deserializer pipe configuration. I removed this block from the recovery path.

ejgoldik (approved):

PR #607, opened 2026-08-24
[RSDSO-21887] enable generic vc mapping

ejgoldik:
- Serializer vc to deserializer vc mapping is now explicitly provided in the device tree
- If no remap table is given, then the default is unity (0 to 0, 1 to 1 etc)
- For multi link device trees, vc mapping table must be provided, otherwise there will be collisions on output VCs

### DIFF
--- a/docs/gmsl-vc-remap.md
+++ b/docs/gmsl-vc-remap.md
@@ -1,0 +1,129 @@
+# GMSL virtual-channel remapping (MAX96712)
+
+Lets a device tree state, per GMSL link, which deserializer CSI virtual channel
+each serializer virtual channel comes out on. Before this, the mapping was
+hardcoded as `dser_vc = link * 4 + ser_vc`, and the link itself was derived from
+the camera's `vc-id` — so VC assignment and link topology could not be chosen
+independently.
+
+Applies to **max96712 only**. max9296 and max96724 are currently single-link in practice
+and keep using the value d4xx passes them unchanged (TODO: max96724 multi-link).
+
+## Device tree
+
+Two properties. The serializer states its link; the deserializer states the map.
+
+```dts
+	dser: max96712@29 {
+		compatible = "maxim,max96712";
+		link-mask = <0x3>;
+		/* <serializer-vc deserializer-vc> */
+		maxim,link0-vc-remap = <0 0>, <1 1>, <2 2>, <3 3>;
+		maxim,link1-vc-remap = <0 4>, <1 5>, <2 6>, <3 7>;
+	};
+
+	ser_a: max96717@40 {
+		maxim,gmsl-link-id = <0>;
+		maxim,gmsl-dser-device = <&dser>;
+	};
+
+	ser_b: max9295_b@41 {
+		maxim,gmsl-link-id = <1>;
+		maxim,gmsl-dser-device = <&dser>;
+	};
+```
+
+The tuple grouping is cosmetic — `<0 0>, <1 1>` and `<0 0 1 1>` compile to the
+same cell array. The driver enforces the stride of 2.
+
+### Rules
+
+- **No table for a link** → unity map, all 8 serializer VCs map to themselves.
+  Usable as-is only on a single-link deserializer.
+- **A table for a link** → it replaces the unity default outright. VCs the table
+  does not list are left *unmapped* and rejected at stream start. Declare exactly
+  the streams the camera uses (today: 4).
+- **A multi-link deserializer must declare a table for every enabled link.** Two
+  links both emitting VC 0-3 would be indistinguishable on the shared CSI output,
+  so the unity default cannot work there. Probe fails with
+  `dser VC <n> mapped more than once`.
+- Deserializer VCs are 5-bit (0-31); serializer VCs are 0-7.
+- `maxim,gmsl-link-id` is optional. Without it d4xx falls back to
+  `vc-id / 4` and logs `no maxim,gmsl-link-id, using link N from vc-id M`,
+  which keeps pre-existing overlays working.
+
+## The deserializer VC appears four times per stream
+
+A remap is only correct if all four agree. Changing the table alone does nothing
+useful — the Tegra side still demultiplexes on the old VC.
+
+| where | consumer |
+|---|---|
+| `maxim,linkN-vc-remap` on the deserializer | max96712 register programming |
+| `ds5_N/ports/port@0/endpoint` `vc-id` | sensor's declared output VC |
+| `ds5_N/gmsl-link/vc-id` | d4xx `g_ctx.dst_vc` |
+| `tegra-capture-vi/ports/port@N/endpoint` `vc-id` | VI channel demux |
+
+The serializer VC never appears in DT. It is the stream index the camera itself
+uses — depth 0, RGB 1, IR 2, IMU 3 — and the driver recovers it by inverting the
+table.
+
+## Driver
+
+`dser_interface` ops take the link explicitly. `vc_id` in these calls is the
+**serializer** VC.
+
+| op | argument |
+|---|---|
+| `get_multi_vc_pipe_id` | `link` |
+| `bind_ser_to_dser_pipe` | `link` |
+| `set_pipe` | `link`, `vc_id` |
+| `reset_oneshot_link` | `link` |
+| `get_ser_vc_id` | `link`, `dser_vc` → serializer VC |
+
+`get_ser_vc_id` is the reverse lookup, and is `NULL` on deserializers that do not
+remap (where the two VCs are the same). d4xx uses it in `ds5_configure()` to
+resolve the serializer VC once per configure.
+
+In the max96712 registers the two VCs are separate fields: the `SRC_n_MAP`
+register carries the incoming serializer VC, `DST_n_MAP` the outgoing one, and
+`MIPI_TX_EXT<n>` carries the high bits of both — destination at bits 4:2, source
+at bits 7:5. They coincide only under the old `link * 4 + ser_vc` layout.
+
+**Camera-side registers use the serializer VC.** `DS5_*_STREAM_MD` is written to
+the camera, so its VC field is the serializer VC, not `dst_vc`. Getting this
+wrong is invisible under an identity map and breaks only the streams that carry
+embedded metadata (depth/RGB/IR, not IMU), because the metadata lands on a VC the
+deserializer has no matching source entry for and the VI channel waits for a line
+that never arrives.
+
+## Limits
+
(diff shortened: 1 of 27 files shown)


docs/gmsl-vc-remap.md
```diff
@@ -0,0 +1,129 @@
...
+| `tegra-capture-vi/ports/port@N/endpoint` `vc-id` | VI channel demux |
+
+The serializer VC never appears in DT. It is the stream index the camera itself
+uses — depth 0, RGB 1, IR 2, IMU 3 — and the driver recovers it by inverting the
+table.
+
+## Driver
+
+`dser_interface` ops take the link explicitly. `vc_id` in these calls is the
+**serializer** VC; the deserializer decides which CSI VC it leaves on.
```
sareluzi: meant "lives"?

Nikolai-L (changes requested):
Looks ok, but honestly I skipped the patches part. If we need to review them I suppose Claude can prepare something more readable.
Anyway a change of such magnitude should pass more then CI/nightly testing and I don't see any test report here.

Nikolai-L (dismissed):

docs/gmsl-vc-remap.md
ejgoldik: To be honest - Not quite sure what claude tried to tell us with this sentence fragment.
Will remove the last half-sentence.

ejgoldik: Fixed

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
```diff
@@ -312,8 +332,106 @@ index 8e237eef..12729a1e 100644
...
++{
++	struct device *dev = &priv->i2c_client->dev;
++	unsigned int link;
++	u32 dser_vc_seen_mask = 0;
++	int map_elems, err, i;
++
++	/* Default unity mapping table */
++	for (link = 0; link < MAX96712_MAX_LINKS; link++) {
++		for (i = 0; i < MAX96712_MAX_SER_VCS; i++) {
++			priv->vc_remap[link][i] = i;
```
sareluzi, line 351: **Back-compat risk (highest-impact item, still non-gating):** the no-table default is now *unity* (ser VC n → dser VC n) for every link, and the new uniqueness check at the end of this function turns that into a hard probe failure.

A pre-existing multi-link max96712 DTB with no `maxim,linkN-vc-remap` (customer overlay, or a stale DTB left in place after a module-only deploy) now fails `max96712_probe` with `dser VC 0 mapped more than once`, d4xx defers forever, and the board comes up with zero `/dev/video*`. d4xx keeps a back-compat fallback for a missing `maxim,gmsl-link-id`; max96712 has none for a missing table.

```c
priv->vc_remap[link][i] = link * MAX96712_MAX_SER_VCS + i;
```
preserves the old `link*4 + ser_vc` behaviour and still lets a table override it. Mirror into `kernel/nvidia/5.0.2/0017` and `kernel/nvidia/5.1.2/0007`.

kernel/nvidia/5.0.2/0017-Adding-max96712-support-for-D4xx.patch
```diff
@@ -258,8 +278,104 @@ index 8081385f0..aaa115696 100644
...
++{
++	struct device *dev = &priv->i2c_client->dev;
++	unsigned int link;
++	u32 dser_vc_seen_mask = 0;
++	int map_elems, err, i;
++
++	/* Default unity mapping table */
++	for (link = 0; link < MAX96712_MAX_LINKS; link++) {
++		for (i = 0; i < MAX96712_MAX_SER_VCS; i++) {
++			priv->vc_remap[link][i] = i;
```
sareluzi, line 297: Same unity-default back-compat concern as `nvidia-oot/6.0/0003` — noting it here so the fix lands in all three carriers byte-identically.

kernel/nvidia/5.1.2/0007-Adding-max96712-support-for-D4xx.patch
```diff
@@ -332,8 +352,104 @@ index 329a10dfd..688f86fb8 100644
...
++{
++	struct device *dev = &priv->i2c_client->dev;
++	unsigned int link;
++	u32 dser_vc_seen_mask = 0;
++	int map_elems, err, i;
++
++	/* Default unity mapping table */
++	for (link = 0; link < MAX96712_MAX_LINKS; link++) {
++		for (i = 0; i < MAX96712_MAX_SER_VCS; i++) {
++			priv->vc_remap[link][i] = i;
```
sareluzi, line 371: Same unity-default back-compat concern as `nvidia-oot/6.0/0003` — third carrier.

kernel/realsense/d4xx.c
```diff
@@ -2242,6 +2253,18 @@ static int ds5_configure(struct ds5 *state)
 	is_calib = (state->is_y8 && (data_type1 == GMSL_CSI_DT_RGB_888));
 
 	vc_id = state->g_ctx.dst_vc;
+	/* vc_id is the deserializer-side VC, from DT and used by the VI graph.
+	 * The serializer and the camera's own registers speak the serializer VC,
+	 * which only the deserializer's remap table can resolve. */
+	ser_vc_id = state->dser_ops->get_ser_vc_id ?
+		state->dser_ops->get_ser_vc_id(state->dser_dev, state->gmsl_link, vc_id) :
+		(int)vc_id;
```
sareluzi, line 2261: The `get_ser_vc_id == NULL` fallback is now raw `vc_id`, where it used to be `vc_id % DS5_MAX_STREAMS`. max9296 provides no `get_ser_vc_id`, so for the two-cameras-on-one-MAX9296 layout the deleted comment documented (dst_vc 4–7) this hands 4–7 to `__max9295_set_pipe()`, which writes `1 << 4 = 0x10` into the 4-bit VC-select field of reg 0x0309 → the pipe selects no VC → no frames. Same value also feeds `serdes_get_ser_pipe_id()` and the `stream_stop()` path.

No in-tree overlay hits this today, which is why it isn't gating, but the invariant that made the raw value safe is gone. Restoring the `% DS5_MAX_STREAMS` fold in the fallback branch keeps the old semantics for deserializers without a remap table.

kernel/realsense/d4xx.c
```diff
@@ -6333,11 +6380,16 @@ static int ds5_mux_s_stream(struct v4l2_subdev *sd, int on)
...
-			state->ser_ops->stream_stop(state->ser_dev,
-						    vc_id % DS5_MAX_STREAMS);
+		 * cleanly instead of wedging the shared pipe. */
+		if (state->ser_ops->stream_stop) {
+			int ser_vc_id = state->dser_ops->get_ser_vc_id ?
+				state->dser_ops->get_ser_vc_id(state->dser_dev,
+							       state->gmsl_link, vc_id) :
+				(int)vc_id;
+
+			if (ser_vc_id >= 0)
```
sareluzi, line 6390: Minor: `stream_stop()` is now silently skipped when the reverse lookup fails. In a `!CONFIG_TEGRA_CAMERA_PLATFORM` build `vc_id` is a stream index rather than a deserializer VC, so on link > 0 the lookup fails every time and the serializer MIPI RX PHY is never re-armed. Worth at least a `dev_warn` so the skip is visible.

sareluzi (commented):
Review of the generic VC-mapping change. **None of the comments below are gating** — the in-tree overlays are all self-consistent and register values are bit-identical for identity tables on link 0, so this can merge and the items can be addressed in a follow-up PR.

What I verified as clean:
- All 16 updated overlays: `maxim,gmsl-link-id` N ↔ `vc-id` 4N..4N+3 ↔ `linkN-vc-remap`, and every link in `link-mask` has a table. No in-tree max96712 overlay with 2+ cameras per deserializer was left table-less.
- JP5 carriers (`5.0.2/0017`, `5.1.2/0007`) got a byte-equivalent delta to `6.0/0003`; all six touched patch files have self-consistent hunk arithmetic and no context-only hunks. max9296 gained the 6-arg `set_pipe` in every generation; max96724 single-source updated.
- `MAX9295_MAX_STREAMS` fully removed, no remaining users. `ext_vc` src bits are 0 for ser VC ≤ 3, so link-0 HW behaviour is unchanged; the `md_value` switch to the serializer VC only changes links > 0, where the old value was wrong or truncated.

One item not anchorable inline: `CLAUDE.md` lines 183 and 195 still document `link = vc_id / MAX9295_MAX_STREAMS` and `reset_oneshot_link(dev, vc_id)`. Those ops take a link now and the macro is gone — the repo's post-patch instruction-hygiene rule wants this fixed alongside.

Suggested follow-up ordering: the `max96712_parse_vc_remap` default (highest impact, must be mirrored byte-identically into all three carriers), then the d4xx fallback VC space, then the CLAUDE.md prose, then HW validation on a 4-camera max96712 rig including one non-identity table from the new doc's swap example.

kernel/realsense/d4xx.c
```diff
@@ -4747,6 +4773,15 @@ static int ds5_board_setup(struct ds5 *state)
 		goto error;
 	}
 
+	/* The GMSL link is a static board fact stated on the serializer node. */
+	if (of_property_read_u32(ser_node, "maxim,gmsl-link-id", &state->gmsl_link))
+		state->gmsl_link = DS5_GMSL_LINK_UNSET;
+	else if (state->gmsl_link >= DS5_MAX_GMSL_LINKS) {
+		dev_err(dev, "illegal maxim,gmsl-link-id %u\n", state->gmsl_link);
+		err = -EINVAL;
+		goto error;
```
sareluzi, line 4782: Nit: this `goto error` leaks the `ser_node` reference — the success path drops it via `of_node_put(ser_node)` a few lines below, but the new range check doesn't.

kernel/realsense/d4xx.c
ejgoldik: % DS5_MAX_STREAMS never worked for max9296 either. max9296 always only outputted the ser vc_id.
An overlay with VC_id > 3 for max9296 is illegal (both before, and now)

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
ejgoldik: Intentionally left out. Preferred a hard probe failure with explicit message rather than a silent output stream collision which is very difficult to debug (a collision would mean frames arriving messed up with very little hint as to why)

sareluzi (approved):

PR #628, opened 2026-08-31
[RSDEV-14228] D401 dual RGB fix exposed res

ejgoldik:
- Expose res 1288x808 for D401 dual RGB
- For this, add alignment arg to the format override macro

This change is breaking from user perspective:
fourcc before: V4L2_PIX_FMT_SBGGR8 — BA81, now: V4L2_PIX_FMT_SBGGR10P — pBAA
advertised size before: 1612 × 808, now: 1288 × 808


kernel/nvidia/5.0.2/0025-Per-format-VI-transport-geometry-and-datatype-override.patch
```diff
@@ -0,0 +1,272 @@
...
++	struct tegra_frac frame_y_scale;
++	u8 frame_x_align;
+ };
+ 
+ #define	TEGRA_VIDEO_FORMAT(VF_CODE, BPP, MBUS_CODE, FRAC_BPP_NUM,	\
+@@ -132,6 +150,27 @@ struct tegra_video_format {
+ 	DESCRIPTION,							\
+ }
+ 
++#define	TEGRA_VIDEO_FORMAT_OVERRIDE(VF_CODE, BPP, MBUS_CODE,		\
```
ev-mp, line 255: Maybe use TEGRA_VIDEO_FORMAT_EXTENDED instead ? Override is just a subset of modifications imposed

kernel/nvidia/5.0.2/0025-Per-format-VI-transport-geometry-and-datatype-override.patch
```diff
@@ -0,0 +1,272 @@
...
++			chan->fmtinfo->frame_y_scale.denominator;
+ }
+ 
+ static void tegra_channel_fmts_bitmap_init(struct tegra_channel *chan)
+@@ -2076,6 +2086,12 @@ __tegra_channel_try_format(struct tegra_channel *chan,
+ 	if (chan->fmtinfo->fourcc == V4L2_PIX_FMT_NV16)
+ 		pix->sizeimage *= 2;
+ 
++	/* Transport rows may exceed V4L2 rows (tegra_camera_core.h). */
++	if (vfmt->frame_y_scale.numerator && vfmt->frame_y_scale.denominator)
```
ev-mp: Defensive programming:
Please add parenthesis to impose strict order of * / operations. Also  - does it need a check for numerator/denominator  > 1 ?
There are two sub-cases:
num/denom <= 0  - invalid UC
num/denom == 1 -> redundant, can be skipped

ev-mp: The clause repeats itself in code in several places

kernel/nvidia/5.0.2/0025-Per-format-VI-transport-geometry-and-datatype-override.patch
```diff
@@ -0,0 +1,272 @@
...
+index c2a9e6f14..7b4391deb 100644
+--- a/drivers/media/platform/tegra/camera/vi/channel.c
++++ b/drivers/media/platform/tegra/camera/vi/channel.c
+@@ -195,6 +195,9 @@ static void tegra_channel_fmt_align(struct tegra_channel *chan,
+ 	fmt_align = (denominator == 1) ? numerator : 1;
+ 	align = lcm(chan->width_align, fmt_align);
+ 	align = align > 0 ? align : 1;
++	/* Keep the stride at least as long as a padded wire line. */
++	if (vfmt->frame_x_align > 1)
++		align = lcm(align, (unsigned int)vfmt->frame_x_align);
```
ev-mp, line 42: Pls add a comment for maintainers

ev-mp (changes requested):
Minor comments. Pls review

kernel/nvidia/5.0.2/0025-Per-format-VI-transport-geometry-and-datatype-override.patch
ejgoldik: What comment in particular?
I thought "/* Keep the stride at least as long as a padded wire line. */" should be enough

kernel/nvidia/5.0.2/0025-Per-format-VI-transport-geometry-and-datatype-override.patch
ejgoldik: I would consider geometry changes a type of override as well, as it overrides the default geometry of the original format (Note that this is also not a new macro in this PR, it was just added to JP5 as well in this PR)

kernel/nvidia/5.0.2/0025-Per-format-VI-transport-geometry-and-datatype-override.patch
ejgoldik: Added helper "static inline u32 tegra_frac_scale(u32 value, const struct tegra_frac *f)" function which should help all these cases.
Note that struct tegra_frac fields are unsigned, so the < 0 case should be fine.

kernel/nvidia/5.0.2/0025-Per-format-VI-transport-geometry-and-datatype-override.patch
ev-mp: ack

ev-mp (approved):

ipilchin1 (approved):

PR #475, opened 2026-06-16
Release 1.0.3.18: merge r/58.2 into master (1.0.2.34 -> 1.0.3.18)

ymodlin:
## Summary

Integration of the **`r/58.2`** maintenance branch (head = tag `v1.0.3.18`, `f1f7ce7`) into `master`, advancing the driver from **1.0.2.34** to **1.0.3.18**.

### Resolution approach

`r/58.2` and the `r/58.1` line already on `master` are parallel release branches that cherry-picked overlapping PRs under different hashes, so a straight merge conflicted in 13 files (incl. `d4xx.c`). Verified before resolving:

- **No r/58.1-only functional changes** — every functional PR on `r/58.1` (#424/#426/#430/#432/#433/#440…) is reimplemented on `r/58.2`; the only r/58.1-unique commits are version bumps (1.0.2.30–34).
- **master's `d4xx.c` is byte-identical to `r/58.1`'s** — master simply carries the r/58.1 release driver.
- **`dev` follows the `r/58.2` architecture** (`serdes_setup_complete`, ESYNC tunneling), not master's superseded `ds5_probe_reset_once` reset path.

Conclusion: `r/58.2` functionally supersedes `r/58.1`. The merge is therefore resolved as a **release promotion** — the merged tree is taken to exactly equal `r/58.2` (recorded as a 2-parent merge commit, `master` + `r/58.2`). This is buildable, matches the tested release, and is consistent with the prior r/58.1→master promotion (#450).

### Major themes included

- **Quad D457 via MAX96712 full-slave mode** (Fangzhu) — RSDSO-20613
- **Seeed Orin Nano** carrier-board support — RSDSO-20614
- **External sync** — on-demand ESYNC GPIO tunneling + TSC-PWM esync (RSDEV-6847, RSDSO-21407)
- **MAX96712 tuning** — lane data rate 2500 Mbps, MFP6 config links B–D, per-link one-shot reset on `set_pipe`
- **Reset / probe robustness** — optimized reset+probe, endless-HW-reset fix, redundant probe-reset removal
- **MIPI DFU recovery** after interrupted FW upgrade — RSDSO-21588
- **Test & CI** — standalone parallel-stream test, quad depth-RGB sync test, auto-increment versioning pipeline, CI-instability fixes

### Notes

- Adds `RELEASE_NOTES_R58.2.md` (cumulative, v1.0.2.27 → v1.0.3.18).
[...]

PR #297, opened 2025-09-15
Fix readme

Kontra2B:
JP6 readme file reviewed

PR #634, opened 2026-09-02
[D5x][GMSL] Add one-node dual-RGB VC mux POC

polejoe:
## Summary

This POC implements the D5x 2C dual-RGB shared-VC path as one aggregate RGB V4L2 node.

- Adds a D5x-only V4L2 menu control: Left, Right, or Frame Alternate.
- Keeps one VI channel, one pixel vb2 queue, and the existing RGB video node.
- Starts the HKR Left and Right logical streams on the same physical VC in Frame Alternate mode.
- Pairs embedded metadata with the exact VI capture descriptor before pixel completion, so userspace can demultiplex by `source_id` without assuming strict LRLR order.
- Adds no full-frame CPU copy or device-to-device copy.
- Leaves existing RGB format tables and the 3C data path unchanged; the control is inactive outside confirmed D5x 2C mode.

## Naming and canonical identity

New D5x-only symbols use the existing `d5x_`/`D5X_` or `d500_`/`D500_` convention. Existing `ds5_`/`DS5_` names are referenced only for shared infrastructure and common register/status definitions.

The HKR PR set uses `Depth/RGB/IMU/Dual-RGB Right/IR = 0/1/2/3/4` for the Host wire command, HKR stream slot, HIF MC, and metadata `source_id`. Dual-RGB is `MC1 + MC3 -> VC1`. VC remains an independent routing property.

## Commit

- `0103f87d` `[D5x][gmsl][feat]: add one-node dual-RGB VC mux`

The PR intentionally contains one commit.

## Dependencies

- HKR shared-VC PR set: build #135, Device Manager #475, Apps #297, IO Drivers #556, and Tests #376.

## Validation

- Base: `dev_gmsl_merge` at `3413a87d`.
- The branch is rebased directly on that base, and local/remote heads both resolve to `0103f87d`.
- JetPack 6.2.2 used a matching Image/module/DTB set; no kernel/module ABI mismatch, Oops, or panic was observed.
- HKR image was built with `BUILD_HIF_MIPI_MC2VC_MUX=ON`; HIF FW is unchanged.
- 2C Left-only and Right-only NV12 `1280x960@30`: PASS at 30 FPS per source.
- 2C Both YUYV `1280x960@30`: PASS, aggregate 180, Left 90, Right 90, zero invalid/unmatched metadata and zero Host drops/errors.
- 2C Both GR16 `1600x1300@25`: PASS for one 1500-frame run, Left 750 and Right 750 at 25 FPS per source, zero invalid/unmatched metadata and zero Host drops/errors.
- 2C Depth + IR + RGB Both + IMU: PASS, RGB Left/Right 93/93.
- NV12 Left/Right content dumps are non-black, non-constant, and have different hashes.
- 3C full-stream NV12 and GR16 regressions passed; the stereo control remained inactive.

## Open lifecycle gap

With the current HKR no-software-VC-serialization image, the first GR16 run completes its Host data path, but HKR can stop MC3 with `req/done=754/753` and reject its detach with `-EBUSY`. An immediate identical reopen then reports that the logical stream is still initialized and produces no frames. The Host `STREAMON` call itself returns, so this failure must be fixed and propagated by the dependent HKR lifecycle path; it is not treated as a passed reopen regression for this PR set.

This POC is independent of PR #601 and does not carry its two-node fan-out implementation.

### DIFF
--- a/nvidia-oot/6.2/0014-Pair-embedded-metadata-per-capture-descriptor.patch
+++ b/nvidia-oot/6.2/0014-Pair-embedded-metadata-per-capture-descriptor.patch
@@ -1,0 +1,197 @@
+From: RealSense <support@realsenseai.com>
+Date: Wed, 19 Aug 2026 11:20:00 +0800
+Subject: [PATCH] Pair embedded metadata per capture descriptor
+
+Allocate one page-aligned embedded-metadata DMA slot per VI capture
+descriptor.  Select the same slot at descriptor setup and buffer release so
+in-flight frames cannot overwrite each other's metadata.  Complete metadata
+with the pixel sequence and timestamp before releasing the pixel vb2 buffer.
+
+The change only adds capture_queue_depth * one-page coherent storage per
+metadata-enabled channel and does not copy pixel data.
+---
+ drivers/media/platform/tegra/camera/vi/channel.c  |  1 +
+ .../platform/tegra/camera/vi/vi5_fops.c           | 54 ++++++++++++++++++-----
+ include/media/mc_common.h                         |  1 +
+ 3 files changed, 46 insertions(+), 10 deletions(-)
+
+diff --git a/drivers/media/platform/tegra/camera/vi/channel.c b/drivers/media/platform/tegra/camera/vi/channel.c
+index ac5e0e9c..a270be19 100644
+--- a/drivers/media/platform/tegra/camera/vi/channel.c
++++ b/drivers/media/platform/tegra/camera/vi/channel.c
+@@ -2935,6 +2935,7 @@ int tegra_channel_cleanup_video_embedded(struct tegra_channel *chan)
+ 		dma_free_coherent(vi_unit_dev,
+ 				chan->emb_buf_size,
+ 				chan->emb_buf_addr, chan->emb_buf);
++		chan->emb_buf_stride = 0;
+ 		chan->emb_buf_size = 0;
+ 	}
+ 
+diff --git a/drivers/media/platform/tegra/camera/vi/vi5_fops.c b/drivers/media/platform/tegra/camera/vi/vi5_fops.c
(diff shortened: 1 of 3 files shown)


nvidia-oot/6.2/0014-Pair-embedded-metadata-per-capture-descriptor.patch
```diff
@@ -0,0 +1,197 @@
+From: RealSense <support@realsenseai.com>
+Date: Wed, 19 Aug 2026 11:20:00 +0800
+Subject: [PATCH] Pair embedded metadata per capture descriptor
+
+Allocate one page-aligned embedded-metadata DMA slot per VI capture
```
lemon6630, line 5: [P1] This per-descriptor storage is the correctness prerequisite for metadata-authoritative demux, but it is installed only from `nvidia-oot/6.2`. `apply_patches.sh` selects one JetPack directory, while the common `d4xx.c` exposes Frame Alternate without a JetPack guard. JP5, 6.0, 6.1, and 7.x can therefore start both RGB sources while retaining the old shared embedded-metadata buffer, allowing in-flight descriptors to overwrite/reassociate metadata and classify pixels as the wrong camera.

Please either port the descriptor patch to every supported VI implementation or compile/runtime-gate Frame Alternate to JP6.2. Port or gate the feature — gpt-5.6-sol, effort high. This needs cross-generation VI/DMA validation.

lemon6630 (changes requested):
Cross-PR review found a merge-blocking portability gap in the metadata-pairing prerequisite. The canonical ID flow and transactional Left/Right startup otherwise look consistent.

Also, `git diff --check` reports extensive whitespace errors in the added NVIDIA patch; please normalize it before merge. Whitespace cleanup — gpt-5.6-luna, effort low. This is mechanical and should not alter patch semantics.

PR #635, opened 2026-09-03
[RSDEV-14092] 2C 3C controls disable

ejgoldik:
Added ds5_v4l2_ctrl_set_disabled so when in 3C mode, the 2C controls will not appear in the v4l2 --list-ctrls menu

This can be used to cleanup device specific ctrls for other usecases as well


sareluzi (dismissed):
LGTM

sareluzi (approved):

PR #638, opened 2026-09-04
[RSDEV-14516][Host][GMSL][D500][Fix] Pair embedded metadata per VI capture descriptor

polejoe:
## Summary

Fix embedded-metadata ownership in the NVIDIA VI5 capture path.

The existing path gives queued video buffers independent pixel DMA addresses, but programs multiple in-flight capture descriptors with one shared embedded-metadata DMA address. A later descriptor can therefore overwrite metadata before an earlier descriptor releases it to userspace.

This change:

- allocates one page-aligned embedded-metadata slot per capture descriptor;
- programs and releases the slot using the same `capture_descr_index`;
- validates the complete allocation geometry with checked arithmetic;
- replaces an existing allocation transactionally and returns real STREAMON errors;
- prepares valid metadata storage before descriptor setup;
- clears the complete DMA state during normal stop and final cleanup.

The correction is folded into the existing embedded-metadata carrier patches so that metadata support is introduced with correct per-descriptor ownership from the start. JP5 support is included following review feedback.

## Scope

This PR intentionally contains:

- no `d4xx.c` changes;
- no new stream ID, register, or V4L2 control;
- no kernel-side Left/Right demux;
- no pixel-frame copy;
- no metadata size-contract change.

The userspace V4L2 ABI is unchanged. Userspace continues to associate video and metadata by sequence and may interpret the metadata payload, including source identity, without kernel-side source parsing.

## Why the carrier diff is larger than the runtime change

This repository stores NVIDIA source changes inside patch files. Folding the correction into an existing carrier changes its hunks and context, so GitHub's patch-of-patch diff is larger than the runtime source delta.

The layout follows the repository's existing patch organization:

- JP5.0.2 keeps its historical split: allocation and descriptor ownership are in `0001`, while release-by-descriptor is adapted in the existing `0004` helper patch;
- JP5.1.2 and JP5.1.6 carry the complete change in their existing `0001` patches;
- JP6.0 owns one `0001` body and JP6.1 links to it;
- JP6.2 has an independent `0001`; repository aliases cover JP6.2.1 and JP6.2.2;
- JP7.0, JP7.1, and JP7.2 retain their existing independent `0001` files.

No new tail repair patch is introduced. Carrier-link cleanup and the separate JP7 metadata-size contract remain outside this PR.

After application, the ownership change is confined to `channel.c`, `vi5_fops.c`, and `mc_common.h`:

| Change | Why it is required |
|---|---|
| Descriptor-indexed DMA address and release lookup | Prevent two in-flight descriptors from owning the same metadata bytes |
| Checked stride and pool-size calculation | Prevent zero-depth and integer-overflow configurations from producing an invalid DMA allocation |
| Transactional pool replacement | Preserve the valid old allocation if a larger replacement cannot be allocated |
| Shared prepare helper | Ensure descriptors receive valid slot geometry while reusing a valid pool when possible |
| Complete state reset on stop/cleanup | Prevent a freed address, handle, or stride from being reused by a later lifecycle |

These checks close one metadata-buffer ownership lifecycle. They do not add routing, product selection, frame-format handling, or pixel processing.

## Root cause and reproduced failure

On the unmodified shared-slot path, a sustained D500 2C one-node RGB Calib `1600x1300@25` run produced 3,443 video buffers and 3,443 sequence-paired metadata buffers, but all metadata payloads were attributed to one source and 1,721 source counters/timestamps were repeated. Correct V4L2 sequence values did not protect the payload because the shared DMA memory had already been overwritten.

The smaller `1280x960` NV12 case did not reproduce the issue. That is a timing result, not evidence that sharing one in-flight metadata address is safe.

Issue details: https://rsjira.realsenseai.com/browse/RSDEV-14516

## Verification

The branch is one commit on current `dev@d5b77a61931a915d254069aae43ca11435b27154`.

Local validation for the added JP5 scope:

- the complete patch series applies to pristine JP5.0.2, JP5.1.2, and JP5.1.6 source baselines;
- full repository builds pass for JP5.0.2, JP5.1.2, and JP5.1.6;
- the JP6.0, JP6.2, JP7.0, JP7.1, and JP7.2 carrier files are byte-identical to the previously reviewed PR head.

The existing JP6.2.2 hardware results remain applicable because the JP6/JP7 implementation was not changed by this update:

- D500 2C RGB Calib `1600x1300@25`: 2,000/2,000 exact video/metadata joins; MC1/MC3 = 1,000/1,000 at 25 FPS per source; zero drop, source, pairing, reuse, or ordering error.
- The immediately following 2C Depth + IR + Dual-RGB + IMU reopen passed: RGB 1,204 exact joins; MC1/MC3 = 602/602 at 30 FPS per source; all stream integrity counters were zero.
- D500 2C RGB-only NV12 `1280x960@60`: 2,405 exact joins; MC1/MC3 = 1,203/1,202 at about 60 FPS per source, about 120 FPS aggregate; all integrity counters were zero.
- D500 3C Depth + IR + RGB + IMU `1280x960@30`: passed 600-frame-class capture for each video stream with exact metadata pairing and zero integrity errors.
- D500 3C RGB-only NV12 and YUYV `1280x960@60`, plus RGB Calib `1600x1300@25`: all passed the requested frame count, FPS, and metadata checks.

One 2C calibration stop observed an HKR-side tail statistic of `req/done=1002/1001`. It remained non-gating: normal close completed and the immediate full-stream reopen passed without stale registration, zero-frame capture, leak, assertion, or new VI error. The acceptance gate is lifecycle reusability and data integrity, not equality of a diagnostic counter at the stop instant.

Companion HKR and validation PRs:

- https://github.com/RealSenseAI-Internal/soc-rs-hkr-trunk/pull/135
- https://github.com/RealSenseAI-Internal/soc-rs-soc-device-mgr/pull/475
- https://github.com/RealSenseAI-Internal/soc-rs-hkr-apps/pull/297
- https://github.com/RealSenseAI-Internal/soc-rs-soc-io-drivers/pull/556
- https://github.com/RealSenseAI-Internal/soc-rs-hkr-test/pull/376

### DIFF
--- a/kernel/nvidia/5.0.2/0001-Porting-driver-patches-to-jetpack-5.0.2.patch
+++ b/kernel/nvidia/5.0.2/0001-Porting-driver-patches-to-jetpack-5.0.2.patch
@@ -1,4 +1,4 @@
-From 9a8deb6e59925a02f2a6c4d7290ee2f160d5212a Mon Sep 17 00:00:00 2001
+From 0ae2028307cd304b258119e8dcf22efd715d96bd Mon Sep 17 00:00:00 2001
 From: Junze Wu <junze.wu@intel.com>
 Date: Tue, 30 Aug 2022 13:24:12 +0800
 Subject: [PATCH] Porting driver patches to jetpack 5.0.2
@@ -10,22 +10,22 @@ Signed-off-by: Junze Wu <junze.wu@intel.com>
  drivers/media/i2c/max9295.c                   | 389 ++++++++++++++++
  drivers/media/i2c/max9296.c                   | 399 ++++++++++++++++
  .../platform/tegra/camera/sensor_common.c     |   4 +
- .../media/platform/tegra/camera/vi/channel.c  | 429 +++++++++++++++++-
+ .../media/platform/tegra/camera/vi/channel.c  | 435 +++++++++++++++++-
  .../media/platform/tegra/camera/vi/graph.c    |  50 +-
  .../platform/tegra/camera/vi/mc_common.c      |  32 ++
  .../platform/tegra/camera/vi/vi4_formats.h    |  32 +-
- .../media/platform/tegra/camera/vi/vi5_fops.c | 136 +-
- .../platform/tegra/camera/vi/vi5_formats.h    |  28 -
+ .../media/platform/tegra/camera/vi/vi5_fops.c | 266 +++++++++--
+ .../platform/tegra/camera/vi/vi5_formats.h    |  30 +-
  .../tegra/camera/tegra_camera_platform.c      |   5 +
  include/media/gmsl-link.h                     |   7 +
  include/media/max9295.h                       |   2 +
  include/media/max9296.h                       |   3 +
- include/media/mc_common.h                     |  25 +
- include/media/tegra_camera_core.h             |  10 +-
- 17 files changed, 1534 insertions(+), 29 deletions(-)
+ include/media/mc_common.h                     |  26 ++
+ include/media/tegra_camera_core.h             |  12 +-
+ 17 files changed, 1643 insertions(+), 70 deletions(-)
 
 diff --git a/drivers/media/i2c/Kconfig b/drivers/media/i2c/Kconfig
-index 8362bc85b..90f5a90e6 100644
+index 8362bc8..873b62e 100644
 --- a/drivers/media/i2c/Kconfig
 +++ b/drivers/media/i2c/Kconfig
 @@ -184,6 +184,24 @@ config NV_VIDEO_AR0234
@@ -54,7 +54,7 @@ index 8362bc85b..90f5a90e6 100644
  
  endif
 diff --git a/drivers/media/i2c/Makefile b/drivers/media/i2c/Makefile
-index af6122bca..0784f0362 100644
+index af6122b..7be6d7e 100644
 --- a/drivers/media/i2c/Makefile
 +++ b/drivers/media/i2c/Makefile
 @@ -23,3 +23,6 @@ obj-$(CONFIG_I2C_IOEXPANDER_DESER_MAX9296) += max9296.o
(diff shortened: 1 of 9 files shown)


ejgoldik (changes requested):
I don't understand this PR.

Dual RGB was supposed to be exactly like regular RGB but twice in throughput. There was not supposed to be new control, nor different handling of metadata. It seems like you send metadata concatenated and then further patch kernel to handle that. None of that is part of the design.

I would not expect any driver changes under the current agreed upon dual RGB design.

kernel/realsense/d4xx.c
```diff
@@ -169,11 +169,13 @@ struct ser_interface {
 #define DS5_DEPTH_STREAM_STATUS		0x1004
 #define DS5_RGB_STREAM_STATUS		0x1008
 #define DS5_IMU_STREAM_STATUS		0x100C
+#define D500_DUAL_RGB_RIGHT_STREAM_STATUS	0x1010
 #define DS5_IR_STREAM_STATUS		0x1014
 
 #define DS5_STREAM_DEPTH		0x0
 #define DS5_STREAM_RGB			0x1
 #define DS5_STREAM_IMU			0x2
+#define D500_STREAM_DUAL_RGB_RIGHT	0x3
```
ejgoldik: No need for new stream ID

kernel/realsense/d4xx.c
```diff
@@ -169,11 +169,13 @@ struct ser_interface {
 #define DS5_DEPTH_STREAM_STATUS		0x1004
 #define DS5_RGB_STREAM_STATUS		0x1008
 #define DS5_IMU_STREAM_STATUS		0x100C
+#define D500_DUAL_RGB_RIGHT_STREAM_STATUS	0x1010
```
ejgoldik: No need for new stream status, as far as the driver is concerned, dual rgb, is rgb

kernel/realsense/d4xx.c
```diff
@@ -196,6 +198,13 @@ struct ser_interface {
 #define DS5_RGB_CONTROL_STATUS 	0x402E
 #define DS5_RGB_OVERRIDE		0x403C
 
+#define D500_DUAL_RGB_RIGHT_STREAM_DT	0x4060
```
ejgoldik: No different control path, dual rgb is rgb

ejgoldik (changes requested):

ejgoldik (dismissed):
LGTM, but we probably should make the same change for JP5

ejgoldik (approved):

PR #641, opened 2026-09-07
[RSDEV-13487] Add 256x144 D58x depth and infrared profiles

pinquanw:
D585 GMSL does not advertise 256x144 for depth and infrared, so RealSense Viewer cannot select that size. Add 256x144 to the D58x depth and IR profile tables using the existing 5/15/30/60/90 fps list. Both tables expose the same size so depth and IR can be selected together.

The camera firmware already has the depth/IR 256x144 path; this change exposes it through the host driver. RGB profile enumeration is unchanged.

Validation: JetPack 6.2 cross-compilation of d4xx.c passed. The final diff adds two profile entries. Hardware streaming validation of depth/IR 256x144 remains outstanding.


ejgoldik (changes requested):
Please squash the commits and add the ticket number

ejgoldik (approved):

PR #646, opened 2026-09-08
[RSDEV-13342] MIPI driver descriptors

ejgoldik:
[RSDEV-13342] MIPI driver descriptors
Read the per-SKU format tables from the camera at probe instead of carrying
them in the driver: a versioned blob at I2C 0x6000, parsed into the V4L2
format tables, with the built-in tables kept as the fallback whenever the
blob is absent or fails validation.

Probe the magic two bytes at a time first and bound every later read by the
declared total size: legacy FW loads a single word for an unmapped register
and its I2C slave wedges if the master clocks past it.


kernel/realsense/d4xx.c
sareluzi, line 1: @ejgoldik - how about putting all the core new descriptors handling in a new c file?

kernel/realsense/d4xx.c
```diff
@@ -1038,6 +1132,503 @@ static int ds5_raw_read(struct ds5 *state, u16 reg, void *val, size_t val_len)
...
+		fmt->override_data_type = ds5_desc_pixfmt_map[slot].wire_dt;
+		fmt->resolutions = res;
+		fmt->n_resolutions = rec.res_count;
+		out->n_formats++;
+	}
+	return out->n_formats ? 0 : -EINVAL;
+}
+
+/* Caller holds ds5_dev->lock. */
+static int ds5_desc_load_locked(struct ds5 *state)
```
sareluzi: this function is a bit too long...

sareluzi: consider small refactor

kernel/realsense/d4xx.c
```diff
@@ -1038,6 +1132,503 @@ static int ds5_raw_read(struct ds5 *state, u16 reg, void *val, size_t val_len)
...
+		 "format descriptor v%u.%u crc 0x%08x: %u/%u/%u/%u formats (depth/ir/rgb/imu)\n",
+		 desc->ver_major, desc->ver_minor, desc->crc,
+		 desc->streams[DS5_STREAM_DEPTH].n_formats,
+		 desc->streams[DS5_STREAM_IR].n_formats,
+		 desc->streams[DS5_STREAM_RGB].n_formats,
+		 desc->streams[DS5_STREAM_IMU].n_formats);
+	return 0;
+}
+
+static void ds5_desc_apply_sensor(struct ds5_sensor *sensor,
```
sareluzi, line 1649: caller responsible to check for NULL ptr?

kernel/realsense/d4xx.c
ejgoldik: Talked to @ymodlin, we will do it in a follow-up as it requires renaming the d4xx.c file (Multi source .ko cannot have one of the sources named like the .ko file)

polejoe:
Reviewed alongside #617. The parser, the fallback and the module-lifetime table list all line up with the design note, and the validation is properly defensive — every range bounds-checked before it feeds a `kcalloc`, duplicate/missing table types rejected, unknown `pixfmt_id` skipped rather than failing the blob. Verified the FW side serves what this parses (236 B, 10 resolutions, depth/IR `(0,8)`, RGB `(0,7)`).

Worth calling out that moving the depth/IR datatype special-case out of `ds5_configure()` and into the tables is what makes the descriptor able to carry it at all — while that knowledge lived inside an `if`, there was nothing for a table to transcribe. It also kills the `// TODO: 0x31 is replaced with 0x1e` comment, which read as if 0x31 were dead. I'd misread it exactly that way while reviewing #617 and filed a bogus finding off it.

The `ds5_desc_pixfmt_map[]` placement for `wire_dt` is right for the reason the doc gives: `0x2E` means CSI-PT on a D401 and RAW16 calibration on a D5xx, so the request byte can't key that table.

---

**`ds5_desc_probe_magic()` treats an I2C error as "no descriptor".** It uses bare `regmap_raw_read`, not `ds5_raw_read`, so there's no retry — and any failure returns false, which becomes `-ENODEV`, which latches `DS5_DESC_ABSENT` for the life of the module and logs *"format descriptors not provided by camera"*. One transient glitch at probe permanently drops a descriptor-capable camera to the static tables, and the log points the investigation the wrong way. The header read in `ds5_desc_read_blob()` and the one in `ds5_desc_recheck()` are bare too.

Retrying doesn't clock any extra bytes, so it doesn't weaken the aperture-probing rule — worth separating read failure from magic mismatch. Related: the retry loop in `ds5_desc_load_locked()` only re-runs on `-EBADMSG`, so a CRC miss gets a second chance but an `-EIO` doesn't.

**Log only carries counts.** `"format descriptor v1.0 crc 0x... : 3/3/1/1 formats"` tells you it parsed, not what it parsed. A `dev_dbg` per format (pixfmt, DT, override) and per resolution (WxH + fps run) would let dmesg alone distinguish a bad blob from a bad parse, which matters because the descriptor itself is opaque hex on the wire and there's no decoder yet.

**Nit:** tab-only line after `goto formats_done;` — checkpatch will flag it.

---

One thing that's now easier and one that isn't. Adding a resolution to an existing SKU is genuinely FW-only, which is the payoff working as intended. Adding a *format* is still three repos — FW enum plus table, `enum rs_pixfmt` plus `ds5_desc_pixfmt_map[]` here, and a `tegra_video_format` row per JetPack carrier. §9 is honest about that, no complaint, just noting the two enums are hand-kept in sync across repos with nothing asserting it.

kernel/realsense/d4xx.c
ejgoldik: Done

kernel/realsense/d4xx.c
ejgoldik: Split into:
ds5_desc_fetch
ds5_desc_build
ds5_desc_publish

kernel/realsense/d4xx.c
```diff
@@ -7258,8 +7887,29 @@ static int ds5_fixed_configuration(struct i2c_client *client, struct ds5 *state)
...
+#ifdef CONFIG_VIDEO_D4XX_SERDES
+	/* D58x on a MAX96712 deserializer runs the serdes link in PIXEL mode. */
+	state->d58x_pixel_mode = (dev_type == DS5_DEVICE_TYPE_D58X) &&
+				 (state->dser_ops == &max96712_interface);
+	/* The FW serves only the formats valid for the serdes link mode, which
+	 * the host alone knows: tell it before asking for the descriptor.
+	 * Non-fatal, as in ds5_hw_init(): FW without this register rejects the
+	 * write, and the descriptor read then simply falls back.
+	 */
+	if (state->d58x_pixel_mode &&
```
ghu3, line 7963: **High — do not load the descriptor after this pixel-mode write fails.** The descriptor's IMU width depends on this write. If it fails, HKR remains on its default tunnel blob (38 bytes), but execution continues into `ds5_desc_apply()` and can accept that blob on a MAX96712 path where the required width is 256 bytes. A transient write failure can therefore publish the wrong geometry instead of taking the safe static pixel-mode table. Please either fail probe or skip descriptor loading and use the static fallback when this write fails.

kernel/realsense/d4xx.c
```diff
@@ -1038,6 +1132,537 @@ static int ds5_raw_read(struct ds5 *state, u16 reg, void *val, size_t val_len)
...
+		res = ds5_desc_build_resolutions(state, desc, n_fps, blob,
+						 tabs[RS_DESC_T_RESOLUTION],
+						 rec.res_first,
+						 rec.res_count);
+		if (IS_ERR(res))
+			return PTR_ERR(res);
+
+		fmt = &out->formats[out->n_formats];
+		fmt->mbus_code = ds5_desc_pixfmt_map[slot].mbus_code;
+		fmt->data_type = rec.src_data_type;
```
ghu3, line 1506: **Medium — validate the format-record invariants before publishing it.** A recognized `pixfmt_id` with `src_data_type == 0` is currently accepted and exposed through V4L2; STREAMON later writes DT zero and fails instead of rejecting the blob and using the built-in tables. The documented `rsvd == 0` rule is also not enforced. Please reject zero source DT and non-zero reserved fields here so malformed descriptors take the intended probe-time fallback path.

ghu3 (commented):
Two additional findings on the current head. The previously raised I2C discovery/retry issue also remains valid and is not duplicated here.

kernel/realsense/d4xx.c
ejgoldik: Fixed in `7f6ae00`.

A `link_mode_configured` flag now gates the descriptor load, so a failed `0x0404` write falls through to the static per-SKU tables instead of accepting a blob built for the other mode:

```c
	if (link_mode_configured && ds5_desc_apply(state))
		goto formats_done;
```

One clarification on the failure modes, since only the second is the bug:

- FW too old to implement `0x0404` also serves no descriptor, so the magic check already rejected and the static pixel-mode table was used. Safe before and after.
- FW implementing both, where the write fails anyway — that is the real hole. The camera stays in tunnel mode and serves the 38-byte IMU record while the link needs 256.

Worth noting `ds5_write()` already retries `DS5_I2C_RETRY_COUNT` times, so this needs a sustained wedge rather than a single glitch. Still worth closing: the retry work in the same round makes the *read* more likely to succeed, which makes the bad-blob outcome more reachable, not less.

kernel/realsense/d4xx.c
ejgoldik: Fixed in `7f6ae00` — both invariants now reject the blob before it reaches V4L2:

```c
		if (!rec.src_data_type || rec.rsvd) {
			dev_warn(&state->client->dev,
				 "format descriptor: format[%u] dt 0x%02x rsvd 0x%02x invalid\n",
				 first + i, rec.src_data_type, rec.rsvd);
			return -EINVAL;
		}
```

Placed *before* the pixfmt lookup deliberately: an unknown `pixfmt_id` is skipped for forward compatibility, but a record still has to be well-formed to be skipped safely.

I went in sceptical of the `rsvd` half, since rejecting on a reserved byte usually breaks forward compatibility. It does not here — the extension mechanism is the TOC`s `entry_size`, so a future field is appended past offset 8 and `rsvd` at offset 3 stays must-be-zero permanently. The design note already committed to it: *"Write it zero, reject it otherwise."* The code just never enforced it.

Checked against the six real blobs (D40x/D41x/D43x/D45x and both D585 link modes): zero records carry `src_data_type == 0` or non-zero `rsvd`, so nothing in the field regresses to the static tables.

ejgoldik:
> Replying to @polejoe: https://github.com/realsenseai/realsense_mipi_platform_driver/pull/646#issuecomment-5679013243

All three taken, in `7f6ae00`.

**`ds5_desc_probe_magic()` treating an I2C error as "no descriptor"** — right, and it was the one with real consequences. It is now `ds5_desc_check_magic()` returning `0` / `-ENODEV` / the errno, and the fetch retries anything that is not a verdict about the descriptor itself:

```c
		/* A verdict about the descriptor itself: a reread cannot change it. */
		if (!ret || ret == -ENODEV || ret == -EPROTONOSUPPORT ||
		    ret == -EINVAL || ret == -ENOMEM)
			break;
```

so a glitch gets a second pass while a legacy camera still short-circuits on the first, and the log no longer blames the camera for a failed read. Same for `ds5_desc_recheck()`, which was reporting "camera now serves a format descriptor" when what actually happened was that the read failed.

Two deliberate splits there. The magic reads stay **bare** `regmap_raw_read`: switching them to `ds5_raw_read` is safe under the over-read rule (it retries at the same length, never clocks extra bytes), but a camera with no descriptor answers from an unmapped register, so every legacy probe would emit `DS5_I2C_RETRY_COUNT` warnings plus an error for an entirely normal condition. The retry belongs at the blob level, where "no descriptor" has already been ruled out. The **header** reads in `ds5_desc_read_blob()` and `ds5_desc_recheck()` did move to `ds5_raw_read`, since those run only after the magic matched.

**Log only carries counts** — added `dev_dbg` per format (stream, pixfmt, request DT, override DT, resolution count) and per resolution (WxH, fps count and first value), ordered format-then-resolutions. I did not dump the framerate pool itself; that needs string assembly in kernel for little gain.

**The tab-only line** — gone, and thanks for it: chasing it showed I had been running checkpatch over the working diff rather than the whole commit. It was an `ERROR: trailing whitespace`, and the commit also carried 6 warnings and 8 checks. All clean now at `--strict`.

On the enums: `mipi_desc_xcheck.py` does catch drift — it rebuilds the driver tables from real blobs and fails on a `pixfmt_id` mismatch. The gap is that nothing runs it. CI wiring is on the follow-up list with the descriptor refresh after HW reset.

kernel/realsense/d4xx.c
ymodlin: change approved

<sub>Auto-generated by the rs-agentic bot.</sub>

kernel/realsense/d4xx.c
ymodlin: change approved

<sub>Auto-generated by the rs-agentic bot.</sub>

CLAUDE.md
```diff
@@ -177,6 +180,9 @@ Each MIPI segment's lane count comes from a **different** DT property, so the ca
...
 - On each HW reset, clear cached values for firmware-populated readiness registers before polling readiness (for example clear `cached_device_type` before waiting for `DS5_DEVICE_TYPE`). Do not let pre-reset cache values short-circuit post-reset readiness checks.
+- **Never clock more I2C bytes than the connected FW is known to load for that register.** D4xx FW answers an unmapped register with a single 16-bit word (and `DS5_HWMC_STATUS`/`_LEN` with 4); the camera's DW I2C slave **wedges — unreachable until power-cycle — when the master reads past what was loaded** (HW-proven: the format-descriptor draft's 16-byte header read at `RS_DESC_BASE` killed a legacy-FW camera). Short reads are safe (the 2-byte `DS5_HWMC_STATUS` poll against a 4-byte load runs on every HWMC). So any register a legacy FW may not implement is probed with 2-byte reads first: `ds5_desc_probe_magic()` reads the two magic words before the 16-byte header, and every later descriptor read is bounded by `total_size`.
```
ymodlin: Stale after this commit's rename: `ds5_desc_probe_magic()` is now `ds5_desc_check_magic()`, and it returns an errno rather than a bool. CLAUDE.md requires docs to track a helper's API contract change.

<sub>Auto-generated by the rs-agentic bot.</sub>

docs/camera-format-descriptor.md
```diff
@@ -0,0 +1,682 @@
...
+| Option | Stateful? | Verdict |
+|---|---|---|
+| **Linear aperture.** 2 KB of read-only register space at a fixed base. | No | **CHOSEN** Chunked reads at any offset, in any order, from any caller. |
+| **Offset register + window.** Write a chunk index, read a fixed window. | Yes | **AVOID** Four subdevs probe the same physical camera concurrently; a shared cursor is a race for no benefit. |
+| **HWMC `GET_DESCRIPTOR`.** Response through the 1 KB HWMC buffer. | Yes | **FALLBACK** Same race, plus HWMC latency. Use only if no contiguous aperture is available. |
+
+```c
+/* two 2-byte probes: legacy FW loads one word for an unmapped register
+ * and its I2C slave wedges if the master clocks past it */
+if (!ds5_desc_probe_magic(state))
```
ymodlin: Same rename, and the snippet's bool contract is stale too — the helper now returns `0` / `-ENODEV` / I2C errno, which is exactly what lets `ds5_desc_fetch()` retry a transient read instead of latching `DS5_DESC_ABSENT`.

<sub>Auto-generated by the rs-agentic bot.</sub>

ymodlin (commented):
Both previous findings are fixed. Two stale doc references left by the `ds5_desc_probe_magic()` rename.

<sub>Auto-generated by the rs-agentic bot.</sub>

kernel/realsense/d4xx.c
ymodlin: change approved

<sub>🤖 auto-generated by the rs-agentic bot</sub>

kernel/realsense/d4xx.c
ymodlin: change approved

<sub>🤖 auto-generated by the rs-agentic bot</sub>

CLAUDE.md
ymodlin: Stale name: this push renamed `ds5_desc_probe_magic()` → `ds5_desc_check_magic()`. The guideline still cites the old name.

<sub>🤖 auto-generated by the rs-agentic bot</sub>

ymodlin (commented):
Re-reviewed the new head (`7f6ae00`). Both findings from the previous pass are fixed correctly:

- **pixel-mode gate** — `link_mode_configured` cleanly skips `ds5_desc_apply()` when the `DS5_MIPI_SERDES_PIXEL_MODE` write fails; `ds5_desc_apply()` has exactly one caller, so the gate is complete.
- **format-record validation** — the `src_data_type == 0` / `rsvd != 0` reject matches the doc contract ("Write it zero, reject it otherwise"), and it sits before the pixfmt lookup without breaking the documented unknown-`pixfmt_id` skip.

Also verified the points raised in @polejoe's review are addressed: `ds5_desc_check_magic()` now returns the errno so a transient I2C failure is retried by `ds5_desc_fetch()` instead of latching `DS5_DESC_ABSENT`; the header reads moved to `ds5_raw_read()` (safe — both are behind a confirmed magic); per-format/per-resolution `dev_dbg` added; tab-only line removed.

No new bugs found in this push. Checked the new `dev_dbg` in `ds5_desc_build_resolutions()` — `desc->framerates[fps_first]` is in bounds because the preceding check rejects `fps_first + fps_count > n_fps` with `fps_count >= 1`. Only doc/comment nits below.

<sub>🤖 auto-generated by the rs-agentic bot</sub>

docs/camera-format-descriptor.md
ymodlin: Same rename — snippet still calls `ds5_desc_probe_magic()` and tests it as a bool; it now returns an errno (`0` / `-ENODEV` / I2C error).

<sub>🤖 auto-generated by the rs-agentic bot</sub>

kernel/realsense/d4xx.c
```diff
@@ -1038,6 +1133,573 @@ static int ds5_raw_read(struct ds5 *state, u16 reg, void *val, size_t val_len)
...
+
+	for (i = 0; i < count; i++) {
+		struct rs_desc_format rec;
+		struct ds5_resolution *res;
+		struct ds5_format *fmt;
+		int slot;
+
+		ds5_desc_rec(blob, ftoc, first + i, &rec, sizeof(rec));
+		/* DT 0 is illegal; rsvd must-be-zero.
+		 * newer fields ,if any, arrive via a longer entry_size.
```
ymodlin: Nit: `fields ,if any,` → `fields, if any,`; capitalize `Newer`.

<sub>🤖 auto-generated by the rs-agentic bot</sub>

CLAUDE.md
```diff
@@ -177,6 +180,9 @@ Each MIPI segment's lane count comes from a **different** DT property, so the ca
...
+- **Tell the camera its link context before asking for the format descriptor.** The descriptor carries no host-context flags: the one mode-dependent entry (D58x IMU record width, 38 B tunnel / 256 B pixel) is resolved by the FW serving the table for the mode it was told. `ds5_fixed_configuration()` therefore writes `DS5_MIPI_SERDES_PIXEL_MODE` right where `d58x_pixel_mode` is decided, before `ds5_desc_apply()`; `ds5_hw_init()` stays the MIPI-config owner and re-writes it after a D585 HW reset, which is why `ds5_desc_recheck()` runs **after** that `ds5_hw_init()` call in `ds5_hw_reset_with_recovery()` — comparing earlier would see the FW's post-reboot tunnel-mode blob and report a false change.
```
ymodlin: **Medium — the rationale here only holds for D585 *prototype* PIDs, so production D585 in pixel mode will log a false "descriptor changed" on every HW reset.**

This note says `ds5_hw_init()` "re-writes it after a D585 HW reset, which is why `ds5_desc_recheck()` runs **after** that `ds5_hw_init()` call". But in `ds5_hw_reset_with_recovery()` that call is guarded by `d585_proto_reset`:

```c
bool d585_proto_reset = d585_product_id == D585_2C_PROTO_PID ||
                        d585_product_id == D585_3C_PROTO_PID;
...
if (d585_proto_reset) {
        ret = ds5_hw_init(state->client, state);
```

On a production D585 behind a MAX96712, `d585_proto_reset` is false, so `ds5_hw_init()` does **not** run, `DS5_MIPI_SERDES_PIXEL_MODE` is not re-written after the FW reboot, and `ds5_desc_recheck()` reads exactly the post-reboot tunnel-mode blob this note says the ordering exists to avoid. Result: a spurious `format descriptor changed across reset (crc ... -> ...); reload d4xx to apply` on every HW reset on the most common D585 configuration — and that message is an operator-facing instruction that is wrong.

Secondary point, worth stating even though it predates this PR: the same gap means production D585 FW is genuinely left in tunnel mode after a reset while the driver still believes `d58x_pixel_mode`. This PR is the first thing that makes that divergence visible, which is useful — but the message should say what it actually is.

Suggested fix: re-write `DS5_MIPI_SERDES_PIXEL_MODE` whenever `state->d58x_pixel_mode` is set before `ds5_desc_recheck()`, not only under `d585_proto_reset`. Then the comparison is apples-to-apples and this note becomes true as written.

kernel/realsense/d4xx.c
```diff
@@ -1038,6 +1133,573 @@ static int ds5_raw_read(struct ds5 *state, u16 reg, void *val, size_t val_len)
...
+	if (present) {
+		if (ds5_raw_read(state, RS_DESC_BASE, &header, sizeof(header)))
+			return;
+		crc = header.crc32;
+	}
+
+	mutex_lock(&dd->lock);
+	if (dd->desc_state == DS5_DESC_READY && crc != dd->desc->crc)
+		dev_warn(&state->client->dev,
+			 "format descriptor changed across reset (crc 0x%08x -> 0x%08x); reload d4xx to apply\n",
```
ymodlin, line 1711: **Medium — see my note on `CLAUDE.md:184`: this warning will fire on every HW reset on a production D585 + MAX96712**, because the `ds5_hw_init()` re-write that is supposed to restore pixel mode before this check only runs for prototype PIDs.

Separate, smaller wording issue in this same block: when the descriptor has disappeared entirely (`present == false`), `crc` stays `0` and this branch reports `crc 0x1234abcd -> 0x00000000`, which reads like a corrupted blob rather than "the camera stopped serving one". The inverse case (`DS5_DESC_ABSENT && present`) gets proper prose; this one deserves the same:

```c
if (dd->desc_state == DS5_DESC_READY && !present)
        dev_warn(..., "camera no longer serves a format descriptor; reload d4xx to apply\n");
else if (dd->desc_state == DS5_DESC_READY && crc != dd->desc->crc)
        dev_warn(..., "format descriptor changed across reset (crc 0x%08x -> 0x%08x); ...");
```

kernel/realsense/d4xx.c
```diff
@@ -1038,6 +1133,573 @@ static int ds5_raw_read(struct ds5 *state, u16 reg, void *val, size_t val_len)
...
+
+/* True when the camera's descriptor supplied every sensor's format table. */
+static bool ds5_desc_apply(struct ds5 *state)
+{
+	struct ds5_dev *dd = state->ds5_dev;
+	const struct ds5_desc *desc = NULL;
+
+	mutex_lock(&dd->lock);
+	if (dd->desc_state == DS5_DESC_UNKNOWN)
+		ds5_desc_load_locked(state);
```
ymodlin, line 1668: **Medium — this runs up to a few KB of I2C while holding the global slot mutex, against the convention already written down in `CLAUDE.md`.**

`dd->lock` here is `ds5_inited[i].lock` — a module-global slot lock that `ds5_serdes_setup()` and `ds5_release_slot()` also scan under. `ds5_desc_load_locked()` -> `ds5_desc_fetch()` does the 2-byte magic probe, the 16-byte header, and up to eight 256-byte chunk reads, and `ds5_desc_fetch()` may repeat the whole sequence on a CRC miss or a transient bus error. At 400 kHz that is on the order of 50-100 ms of bus time per attempt with the slot lock held, serialising against another camera's slot scan during a concurrent probe.

`CLAUDE.md` states the rule this violates: *"For sibling-health checks, snapshot pointers/flags under lock and perform I2C probing after unlocking."*

Suggested shape — keep the lock for the state transition only:

```c
mutex_lock(&dd->lock);
snapshot = dd->desc_state;
mutex_unlock(&dd->lock);

if (snapshot == DS5_DESC_UNKNOWN)
        new_desc = ds5_desc_load(state);   /* no lock held */

mutex_lock(&dd->lock);
if (dd->desc_state == DS5_DESC_UNKNOWN)
        /* publish new_desc */
else
        /* lost the race: drop new_desc, use dd->desc */
desc = (dd->desc_state == DS5_DESC_READY) ? dd->desc : NULL;
mutex_unlock(&dd->lock);
```

The loser of the race can simply be freed — this is the one point where freeing a `ds5_desc` is safe, because no `sensor->formats` points into it yet.

To be clear about what this is not: there is no deadlock today. Nothing on the regmap/I2C path re-enters `dd->lock`, and `ds5_desc_all_lock` nests strictly inside `dd->lock` at the single site in `ds5_desc_publish()`. This is about hold time and about not leaving a counter-example to a documented convention in the tree.

kernel/realsense/d4xx.c
```diff
@@ -1038,6 +1133,573 @@ static int ds5_raw_read(struct ds5 *state, u16 reg, void *val, size_t val_len)
...
+	u16 *framerates;
+	u32 crc;
+	u8 ver_major;
+	u8 ver_minor;
+};
+
+/* A sibling subdev can outlive the primary that owns the slot (sysfs unbind),
+ * so parsed tables are never freed while the module is loaded.
+ */
+static LIST_HEAD(ds5_desc_all);
```
ymodlin, line 1162: **Medium — `ds5_desc_all` grows without bound across unbind/bind cycles.**

I agree with the never-free-while-loaded rationale for the *sibling outlives the primary* case. The problem is its interaction with slot re-init: `ds5_init_ds5_dev()` clears `desc = NULL; desc_state = DS5_DESC_UNKNOWN`, so the next probe re-fetches, re-parses and `list_add()`s a **new** `ds5_desc`. The previous one is unreachable from that moment but stays on the list until `module_exit`.

So a sysfs unbind/bind loop leaks roughly one parsed descriptor per cycle — the `ds5_desc`, up to four `formats` arrays, one `ds5_resolution` array per format, and the framerate pool; call it 1-2 KB a cycle. Not a crash, but unbounded and invisible, and re-probe loops are exactly what the stability tests in `test/` exercise.

Two ways out, either is fine:
1. **Refcount it.** A `kref` on `struct ds5_desc`, one reference per `ds5_sensor` that points into it, released in subdev teardown. That frees on the last sibling going away — precisely the invariant the comment above describes.
2. **Reuse instead of reparse.** Treat `ds5_desc_all` as a cache keyed on `(crc, ver_major, ver_minor)` and have `ds5_desc_publish()` look for a match before adding. A re-probe of the same camera on the same FW then reattaches to the existing tables and allocates nothing — and it removes the redundant re-fetch on every rebind as a bonus.

Option 2 is the smaller change and fits the module-lifetime model you already have.

kernel/realsense/d4xx.c
```diff
@@ -1038,6 +1133,573 @@ static int ds5_raw_read(struct ds5 *state, u16 reg, void *val, size_t val_len)
...
+		if (!rec.src_data_type || rec.rsvd) {
+			dev_warn(&state->client->dev,
+				 "format descriptor: format[%u] dt 0x%02x rsvd 0x%02x invalid\n",
+				 first + i, rec.src_data_type, rec.rsvd);
+			return -EINVAL;
+		}
+		slot = ds5_desc_pixfmt_slot(rec.pixfmt_id);
+		if (slot < 0) {
+			dev_warn(&state->client->dev,
+				 "format descriptor: pixfmt %u unknown, skipped\n",
```
ymodlin, line 1489: **Medium — skipping an unrecognised pixfmt silently re-assigns the node's default format.**

The forward-compat skip itself is right. What is not visible is that it can quietly break a contract the docs commit to — `docs/camera-format-descriptor.md:151`:

> *"The first record of a range is the default, exactly as in the driver's own tables, so record order carries it."*

`ds5_sensor_format_init()` takes `sensor->formats[0]` and `formats[0].resolutions[0]` as the node default. If a future FW puts a pixfmt this driver does not know at index 0 of a stream's range, that record is `continue`d and **index 1 becomes the default** — e.g. an IR node defaulting to Y8I instead of Y8 — with nothing in the log tying the change to the skip. Userspace that relies on the format at open (and the pytest suite does) would change behaviour on a FW update, under a `dev_warn` that reads like a harmless forward-compat notice.

Minimum ask: make the skip of the *first* record of a stream loud, since that is the one with a side effect.

```c
if (slot < 0) {
        dev_warn(&state->client->dev,
                 i == 0 ? "format descriptor: stream %u: unknown pixfmt %u at index 0, default format shifts to the next entry\n"
                        : "format descriptor: stream %u: pixfmt %u unknown, skipped\n",
                 st->stream_id, rec.pixfmt_id);
        continue;
}
```

If you would rather FW could not move a default at all, the stricter alternative is to reject the blob when index 0 of any stream is unrecognised — that keeps the documented contract absolute at some cost in forward compatibility. Your call which way; the current silent shift is the one option that matches neither the doc nor the old behaviour.

---

**Nit, on the new `rsvd`/`src_data_type` check just above (`7f6ae00`):** it sits *before* `ds5_desc_pixfmt_slot()`, so it also applies to records this driver is otherwise willing to skip. A future FW that gives `rsvd` a meaning inside a new pixel format would therefore take down the entire descriptor for every older driver, rather than having that one record skipped. Strict must-be-zero is a defensible contract and the doc states it — just worth being deliberate that it is stricter than the skip-unknown-pixfmt policy sitting three lines below it.

kernel/realsense/d4xx.c
```diff
@@ -7258,8 +7925,33 @@ static int ds5_fixed_configuration(struct i2c_client *client, struct ds5 *state)
...
+				 (state->dser_ops == &max96712_interface);
+	/* The FW serves only the formats valid for the serdes link mode, which
+	 * the host alone knows: tell it before asking for the descriptor.
+	 * Non-fatal, as in ds5_hw_init(): FW without this register rejects the
+	 * write, and the descriptor read then simply falls back.
+	 */
+	if (state->d58x_pixel_mode &&
+	    ds5_write(state, DS5_MIPI_SERDES_PIXEL_MODE, 1)) {
+		dev_warn(&client->dev,
+			 "FW has no serdes_pixel_mode support (reg 0x0404)\n");
```
ymodlin, line 7966: **Nit — this makes the "no serdes_pixel_mode support" warning fire 8x per camera on older FW.**

`ds5_fixed_configuration()` and `ds5_hw_init()` are each called once per i2c client, and there are four clients per camera (Depth / Y8 / RGB / IMU). With this addition, a D585 on a MAX96712 running FW without `0x0404` emits the identical warning eight times during probe. The write is idempotent so there is no functional problem — it is log noise that will get mistaken for a retry loop.

`dev_warn_once()` would do, or gate the second write on a flag in `ds5_dev`.

kernel/realsense/d4xx.c
```diff
@@ -8566,7 +9252,21 @@ static struct i2c_driver ds5_i2c_driver = {
...
+{
+	return i2c_add_driver(&ds5_i2c_driver);
+}
+
+static void __exit ds5_module_exit(void)
+{
+	struct ds5_desc *desc, *tmp;
+
+	i2c_del_driver(&ds5_i2c_driver);
+	list_for_each_entry_safe(desc, tmp, &ds5_desc_all, node)
```
ymodlin, line 9288: **Nit — walking `ds5_desc_all` here without `ds5_desc_all_lock` is safe but asymmetric.**

It is correct as written: `i2c_del_driver()` has already returned, so no probe can still be adding to the list. But this is the only access to `ds5_desc_all` that skips the lock, which leaves a future reader to reconstruct the invariant. Either take the mutex (it costs nothing on the unload path) or add a one-line comment saying the `i2c_del_driver()` above is what makes the bare walk safe.

CLAUDE.md
ymodlin: **Nit — function name drift.** This note (and `docs/camera-format-descriptor.md:548`) refers to `ds5_desc_probe_magic()`; the function is called `ds5_desc_check_magic()`. Worth fixing in both places while the name is still fresh — these instruction files are what future readers grep.

ymodlin (dismissed):
## Review — [RSDEV-13342] MIPI driver descriptors

Reviewed the whole diff: descriptor fetch/validate/parse, the `data_type`/`override_data_type` split in the static tables, the `ds5_configure()` simplification, and the probe/reset wiring.

**Verdict: approve.** No blockers. The parser's bounds checking is solid — I walked every FW-controlled index (`header_size`/`total_size`, TOC `offset`/`entry_size`/`n_entries`, `fmt_first`/`res_first`/`fps_first` plus their counts) and could not construct a blob that reads outside the `kzalloc(total)` buffer; the `u16` field widths keep every `off + n * stride` computation inside `u32`. The stride-tolerant `ds5_desc_rec()` and the reject-and-fall-back-to-static posture are the right shape for this.

I also checked that the `ds5_configure()` rewrite is behaviour-preserving for the static path, since that is what every shipping FW will take:
- depth: `n_formats = 1`, so the old `is_depth && dt != 0 -> 0x31` only ever hit Z16, which now carries `DS5_FW_DT_Z16` directly. Equivalent.
- IR: the old `is_y8 && dt == YUV422_8 -> 0x32` only matched the Y8I entries, which now carry `DS5_FW_DT_R8L8` directly. Y8 / Y12I / RAW16 / CSI-PT were untouched then and are untouched now. Equivalent.

Severity scale used below:

| Level | Meaning |
|---|---|
| **Blocker** | Must be fixed before merge — breaks a supported configuration, or is a memory-safety / locking defect, or regresses shipping behaviour. |
| **High** | Should be fixed before merge — wrong behaviour in a plausible field scenario, but the trigger is narrow or not on a currently shipping path. |
| **Medium** | Fix in this PR or an agreed follow-up — correctness/robustness gap, project-convention violation, or misleading operator-facing output. |
| **Nit** | Style, naming, docs, log hygiene. Author's call. |

Found: **0 Blocker, 0 High, 4 Medium, 3 Nit.** Reviewed at `7f6ae00`, i.e. including the fixes for @ghu3's two findings — both look right to me: the `link_mode_configured` gate correctly falls back to the static tables rather than trusting a blob built for the wrong link mode, and validating `src_data_type != 0` / `rsvd == 0` before publishing is the right place for it.

Approving because none of the Medium items break a supported configuration today — no shipping FW serves the blob, so every camera in the field takes the static path, which I verified is unchanged. Please still address them (or file the follow-up) before this rides a FW that actually serves the descriptor.

CI note: the 10 `build` jobs were still pending when I reviewed; this approval assumes they go green.

CLAUDE.md
ymodlin: This is should fix I agree

kernel/realsense/d4xx.c
ymodlin: No need to fix

kernel/realsense/d4xx.c
ymodlin: No need to fix

CLAUDE.md
ejgoldik: Fixed in `73e2bb3`, with one correction to the framing.

The gap is real and the analysis of it is right: `d58x_pixel_mode` is set from `dev_type == DS5_DEVICE_TYPE_D58X && dser_ops == &max96712_interface`, with no PID involved, while the `ds5_hw_init()` that re-asserts `0x0404` is gated on `d585_proto_reset`. The two disagree by construction. `ds5_hw_reset_with_recovery()` now re-asserts the mode on the `else` branch:

```c
	} else if (state->d58x_pixel_mode &&
		   ds5_write(state, DS5_MIPI_SERDES_PIXEL_MODE, 1)) {
```

**The correction: nothing logs this today, and nothing can.** There is no production D585 — every existing part is a prototype PID, so `d585_proto_reset` is always true, `ds5_hw_init()` always runs, and the comparison is already apples-to-apples. The new `else` branch needs a D58X whose PID is neither prototype, which is currently the empty set, so it is unreachable on any hardware that exists. "Production D585 in pixel mode will log a false descriptor changed on every HW reset" overstates it — no such camera exists to log it.

That does not make it not worth fixing; it makes it worth fixing *cheaply now* rather than rediscovering it later, because the failure mode when a production PID lands is silent (FW in tunnel mode against a pixel-mode link) plus an operator-facing instruction that is wrong. But it does mean this cannot be validated on a rig, so it ships untested by necessity.

Your secondary point is the durable one and I have recorded it as such: the underlying issue is that `ds5_hw_init()` was the sole re-assert of `0x0404` while `d58x_pixel_mode` is decided independently of PID. CLAUDE.md now documents both the new write site and, explicitly, that the branch is dead until a non-prototype PID appears — so a future reader does not assume it is exercised, and knows it is the first path to test when production hardware arrives.

kernel/realsense/d4xx.c
ejgoldik: Fixed in `73e2bb3` — you were right, and this is the one of the pair that acts on hardware we actually have.

I went in expecting to refute it, since "never free while loaded" reads as deliberate. It is deliberate, but only for the sibling-outlives-primary case; the leak is the unexamined interaction with slot re-init, which is a different situation. The chain holds exactly as you describe: `ds5_init_ds5_dev()` clears `desc`/`desc_state`, the next probe re-fetches and re-parses, `list_add()`s a fresh `ds5_desc`, and the previous one is unreachable from that moment with nothing to remove it.

Took option 2. `ds5_desc_publish()` now looks for a match before adding:

```c
	mutex_lock(&ds5_desc_all_lock);
	list_for_each_entry(d, &ds5_desc_all, node) {
		if (d->crc == desc->crc && d->ver_major == desc->ver_major &&
		    d->ver_minor == desc->ver_minor) {
			pub = d;
			break;
		}
	}
	if (!pub) {
		list_add(&desc->node, &ds5_desc_all);
		pub = desc;
		desc = NULL;
	}
	mutex_unlock(&ds5_desc_all_lock);
	ds5_desc_free(desc);
```

Three things that make the key safe: the CRC covers the whole table body (`[header_size, total_size)`), so an identical blob is an identical parse — the parse depends on nothing else but `ds5_desc_pixfmt_map`, which is const. The D585 tunnel and pixel blobs differ in the IMU record width, so they have different CRCs and stay separate entries rather than aliasing. And `ds5_desc_free()` never touches the list, so freeing the just-built copy that lost the race is safe.

The bonus you predicted is real: a rebind on unchanged FW now allocates nothing at all.

CLAUDE.md
ejgoldik: Fixed in `73e2bb3`, in both places, and for the contract as well as the name: `ds5_desc_check_magic()` now returns `0` / `-ENODEV` (magic mismatch, i.e. no descriptor) / an I2C errno, and CLAUDE.md records why it deliberately keeps bare `regmap_raw_read` — a camera serving no descriptor answers from an unmapped register, so retrying there would be pure log noise on every legacy probe. Separating read failure from magic mismatch is what lets `ds5_desc_fetch()` retry a transient read instead of latching `DS5_DESC_ABSENT` for the life of the module.

Swept both files for any other stale `ds5_desc_*` name while I was in there; none left.

ejgoldik: Fixed in `73e2bb3`, in both places, and for the contract as well as the name: `ds5_desc_check_magic()` now returns `0` / `-ENODEV` (magic mismatch, i.e. no descriptor) / an I2C errno, and CLAUDE.md records why it deliberately keeps bare `regmap_raw_read` — a camera serving no descriptor answers from an unmapped register, so retrying there would be pure log noise on every legacy probe. Separating read failure from magic mismatch is what lets `ds5_desc_fetch()` retry a transient read instead of latching `DS5_DESC_ABSENT` for the life of the module.

Swept both files for any other stale `ds5_desc_*` name while I was in there; none left.

ejgoldik: Fixed in `73e2bb3`, in both places, and for the contract as well as the name: `ds5_desc_check_magic()` now returns `0` / `-ENODEV` (magic mismatch, i.e. no descriptor) / an I2C errno, and CLAUDE.md records why it deliberately keeps bare `regmap_raw_read` — a camera serving no descriptor answers from an unmapped register, so retrying there would be pure log noise on every legacy probe. Separating read failure from magic mismatch is what lets `ds5_desc_fetch()` retry a transient read instead of latching `DS5_DESC_ABSENT` for the life of the module.

Swept both files for any other stale `ds5_desc_*` name while I was in there; none left.

docs/camera-format-descriptor.md
ejgoldik: Fixed in `73e2bb3` — the snippet now shows the errno form rather than the bool, with the reason it matters inline:

```c
/* two 2-byte probes: legacy FW loads one word for an unmapped register
 * and its I2C slave wedges if the master clocks past it.
 * Returns 0 / -ENODEV (no descriptor) / an I2C errno, so a transient
 * read failure can be retried instead of latching the static tables. */
ret = ds5_desc_check_magic(state);
if (ret < 0)
	return ret;
```

ejgoldik: Fixed in `73e2bb3` — the snippet now shows the errno form rather than the bool, with the reason it matters inline:

```c
/* two 2-byte probes: legacy FW loads one word for an unmapped register
 * and its I2C slave wedges if the master clocks past it.
 * Returns 0 / -ENODEV (no descriptor) / an I2C errno, so a transient
 * read failure can be retried instead of latching the static tables. */
ret = ds5_desc_check_magic(state);
if (ret < 0)
	return ret;
```

kernel/realsense/d4xx.c
ejgoldik: Fixed in `73e2bb3` — now `Newer fields, if any, arrive via a longer entry_size.`

kernel/realsense/d4xx.c
ejgoldik: Fixed in `73e2bb3` — `ds5_hw_reset_with_recovery()` now re-asserts `DS5_MIPI_SERDES_PIXEL_MODE` on the `else` branch whenever `d58x_pixel_mode` is set, so the post-reset comparison is apples-to-apples. Full reasoning on the `CLAUDE.md:184` thread, including the correction that no production D585 exists yet, so neither this warning nor the underlying tunnel-mode divergence can occur on current hardware — the branch is unreachable until a non-prototype PID appears, and CLAUDE.md says so explicitly.

kernel/realsense/d4xx.c
ejgoldik: If a format is skipped because it is unknown, and it was supposed to be the default, then using the next one is the expected behaviour. Won't fix

CLAUDE.md
ymodlin: change approved

<sub>🤖 auto-generated by the rs-agentic bot</sub>

docs/camera-format-descriptor.md
ymodlin: change approved

<sub>🤖 auto-generated by the rs-agentic bot</sub>

CLAUDE.md
ymodlin: change approved

<sub>🤖 auto-generated by the rs-agentic bot</sub>

docs/camera-format-descriptor.md
ymodlin: change approved

<sub>🤖 auto-generated by the rs-agentic bot</sub>

kernel/realsense/d4xx.c

<sub>🤖 auto-generated by the rs-agentic bot</sub>

CLAUDE.md
ymodlin: change approved

<sub>🤖 auto-generated by the rs-agentic bot</sub>

kernel/realsense/d4xx.c
ymodlin: change approved — the `(crc, ver_major, ver_minor)` cache in `ds5_desc_publish()` closes it: a rebind now reattaches to the existing tables and frees the reparsed copy, so the list stops growing per unbind/bind cycle. CRC covers the TOCs and all tables (`crc32_le` over `blob + header_size`), so the key is sound; `sensor->formats` is `const`, so sharing one `ds5_desc` between two cameras on identical FW is safe.

<sub>🤖 auto-generated by the rs-agentic bot</sub>

CLAUDE.md
ymodlin: change approved — the `else if` re-asserts `DS5_MIPI_SERDES_PIXEL_MODE` before `ds5_desc_recheck()`, so the comparison is apples-to-apples on a production PID.

One correction to the new note: **"that branch is unreachable today" is not quite true.** `d585_proto_reset` is derived from `ds5_dev->d585_product_id`, which is 0 until the probe-time GVD read succeeds — and that read is non-fatal (`"cannot cache D585 PID from GVD"`). If it fails on all four clients, a *prototype* D585 takes the `else` branch: pixel mode is restored but `ds5_hw_init()` is skipped, so `DS5_MIPI_LANE_NUMS`/`DS5_MIPI_LANE_DATARATE` are not reprogrammed after the HKR reboot. That gating predates this PR — only the doc claim is new. Suggest softening to "unreachable on current hardware once the PID is cached".

<sub>🤖 auto-generated by the rs-agentic bot</sub>

kernel/realsense/d4xx.c
ymodlin: Main point: change approved — the production-D585 pixel-mode re-write lands before `ds5_desc_recheck()`, so the spurious "descriptor changed" no longer fires.

Secondary point is still open: when `present == false`, `crc` stays 0 and the first branch still prints `crc 0x1234abcd -> 0x00000000`, which reads as corruption rather than "the camera stopped serving one".

<sub>🤖 auto-generated by the rs-agentic bot</sub>

kernel/realsense/d4xx.c
ymodlin: Still open in `73e2bb3` — `ds5_desc_apply()` continues to call `ds5_desc_load_locked()` (magic probe + header + chunked reads, retried once) with `dd->lock` held. Fine to defer, but please say so rather than leave it silent against the `CLAUDE.md` snapshot-then-probe rule.

<sub>🤖 auto-generated by the rs-agentic bot</sub>

kernel/realsense/d4xx.c
ymodlin: Still open in `73e2bb3` — an unknown `pixfmt_id` at index 0 of a stream range is still `continue`d silently, so the node default shifts to the next record. Minimum ask remains: make the index-0 skip say so in the warning.

<sub>🤖 auto-generated by the rs-agentic bot</sub>

ymodlin (approved):

PR #648, opened 2026-09-11
fix building from source package for jp7

mengyui:
sync hardware/ source from server.
move device tree to hardware folder for jp7
build dtbo from dtso files
add missing format mapping, align to 7.0/7.1


ymodlin (changes requested):
Thanks for taking on the JP7 source-package flow — the `dtbo-source` fallback in `Makefile.generic` is clean, the `dtbo-y` lists match the 2 tegra234 / 8 tegra264 `.dtso` files exactly, and the `0009-Fix-vi5-yuv422` symlink for 7.2 is a correct alignment fix.

Two blockers before this can merge.

### 1. The overlay-build move is only half migrated — the git/CI flow breaks

This PR deletes `kernel/kernel-noble-src/7.{0,1,2}/0004-add-d4xx-camera-overlays.patch`, which is what currently makes the `.dtbo` appear in the kernel tree. Only `apply_patches_ext.sh` was updated to the new nv-public overlay mechanism; the git flow was not:

- `apply_patches.sh:129-130` still links the `.dtso` into `${BUILD_SRCS}/$KERNEL_DIR/arch/arm64/boot/dts/nvidia/` — now dead files, since nothing lists them there any more. Only `apply_patches_ext.sh` copies them into the nv-public `overlay/` dirs, so in the git flow the new `dtbo-y +=` entries in `hardware/nvidia/{t23x,t264}/nv-public/overlay/Makefile` have **no source file** → `No rule to make target ....dtso`.
- `build_all.sh:130` still does `cp $BUILD_SRCS/$KERNEL_DIR/arch/arm64/boot/dts/nvidia/tegra2[36]4-camera-d4xx-overlay*.dtbo ...`. Nothing produces those any more → the `cp` fails under `set -e` and the JP7.0/7.1/7.2 builds abort.
- On 7.x, `build_all.sh` never invokes `make dtbs` at all, and the `nvidia-dtbs` target in our own top-level `nvidia-oot/Makefile:180` only builds `nvidia-oot/device-tree/platform/generic-dts` (the JP6 route). So our build never descends into the nv-public overlay dirs this PR just populated.

Evidence that the old mechanism is the live one: dev's JP7.0/7.1/7.2 builds are green today with the `cp` at `build_all.sh:130`, which only works because `0004` puts the `.dtbo` in the kernel tree.

Either keep `0004` for the git flow and add the nv-public route only for the source-package flow, or migrate `apply_patches.sh` (copy the `.dtso` into the two `overlay/` dirs) **and** `build_all.sh` (build + collect from the new output path) in this same change.

### 2. CI has never run on this PR

All ten build workflows are in `action_required` (awaiting approval); only the Aikido scan ran. Since this touches the shared `7.x` patch dir and `scripts/sources_7.x` for 7.0/7.1/7.2 at once, JP7.0/7.1/7.2 must be green before merge — and 5.x/6.x too, because the new kernel-noble symlink hunk in `apply_patches_ext.sh:36-39` is not gated on `JETPACK_VERSION == 7.x` and therefore also executes on 5.x/6.x.

Two things the CI run specifically needs to prove: that the new `7.x` overlay-Makefile patches apply cleanly at all three tags (`git apply` has no fuzz, and the same patch dir is shared by R38.2 / R38.4 / R39.2), and that the three newly added repos in `scripts/sources_7.x` all carry `jetson_38.2` / `jetson_38.4` tags — `sync-sources.sh` hard-fails with "Couldn't find tag" otherwise.

I have a handful of smaller non-blocking notes (unstaged `t264` / `build/nvidia-public` trees in `apply_patches.sh`, the ungated symlink, stale `CLAUDE.md:90`, pre-existing `sources_7.x` workspaces and the cache tarball missing the new repos) — happy to add them once the two above are settled.

PR #652, opened 2026-09-14
[RSDEV-14616] Use BA10 exclusively for D500 RGB calibration

polejoe:
## Summary
Use BA10 (V4L2 SGRBG10) as the sole D500 RGB calibration format in both 3C and 2C. Replace the GR16 output row rather than adding an alias.

Tracked on [RSDEV-14616]

## Design
- Replace GR16 with BA10 on the existing D500 SGRBG16_1X16 calibration transport row in the shared JP6 and JP7 patch carriers.
- Resolve FourCC through the channel's advertised format bitmap first, with the original lookup as fallback. Other sensors retain the stock RAW10 BA10 row.
- Reuse the existing byte-preserving calibration descriptor settings. No changes to d4xx firmware requests, SerDes programming, VI capture scheduling or NVCSI capture logic.

The HKR output remains RAW16. The two lookup helpers must distinguish this transport from the stock RAW10 BA10 row; no additional driver or capture-path changes are needed.

Applications requesting GR16 must switch to BA10. The internal SGRBG16_1X16 media-bus code still describes the RAW16 transport, not a userspace GR16 format.

SDK integration: https://github.com/realsenseai/librealsense/pull/15668

## Validation
- Patch application checks against NVIDIA JP6.0, JP6.2.2, JP7.0 and JP7.1 source tags.
- Compiled lookup tests verify BA10-only calibration, no GR16 row, stock RAW10 selection, shared-FourCC ordering, offset, empty bitmap and unknown formats.
- The BA10-only revision passed a complete JP6.2.2 kernel/OOT build. Its matching Image, complete module tree, DTB and overlay were deployed with backup and pre-reboot ABI/boot-integrity checks.
- The IR Y16I descriptor is unchanged in both patch carriers. Local preview regression checks preserve Y16I decoding and legacy/current byte-order handling.
- The BA10-only revision boots successfully on Jetson67 with matching Image/modules and no detected module ABI errors. RGB enumerates BA10 1600x1300 without GR16; IR still enumerates Y16I 1600x1300.
- manually confirmed preview works with the latest BA10-only host revision in Both pixel and tunnel mode. Ready for review. Full IR calibration and 2C/3C regression coverage is not claimed.


nvidia-oot/6.2/0013-Tegra-video-format-extended.patch
```diff
@@ -154,6 +162,37 @@ index 487ea19..7f4e64d 100644
...
  				dev_warn(vi->dev,
  					"corr_err: discarding frame %d, flags: %d, "
+diff --git a/drivers/media/platform/tegra/camera/vi/core.c b/drivers/media/platform/tegra/camera/vi/core.c
+--- a/drivers/media/platform/tegra/camera/vi/core.c
++++ b/drivers/media/platform/tegra/camera/vi/core.c
+@@ -96,6 +96,14 @@ int tegra_core_get_code_by_fourcc(struct tegra_channel *chan,
+ {
+ 	unsigned int i;
+ 
++	/* A memory format may have several wire encodings. Prefer the
```
ejgoldik: Per Evgenis request: Please add a comment that we are switching from a one to one fourcc -> mbus code to a one to many mapping, requiring matching both the fourcc and the mbus code. And that is required so SGRBG10 can be used for two different formats

ejgoldik (changes requested):
Please add the comment in the patches, otherwise looks good

nvidia-oot/6.2/0013-Tegra-video-format-extended.patch
polejoe: Added the requested comments in both the JP6 and JP7 patch carriers. They now explain that BA10 is a one-to-many FourCC mapping: the stock row uses SGRBG10_1X10/RAW10, while D58x calibration uses SGRBG16_1X16/RAW16 with one 10-bit sample per 16-bit LE word. The lookup comments also explain why fmts_bitmap must select the channel-advertised media-bus row before the legacy first-match fallback, including the descriptor fields affected by that choice.

ejgoldik (approved):
Please verify on pixel mode to make sure

PR #654, opened 2026-09-14
[RSDEV-14617]:[D58x] Cache GVD and accelerate stream polling

pinquanw:
## Summary
- cache the immutable D58x GVD response per physical camera and reuse it across subdevices
- serve both native and HWMC_RW GVD requests from the shared cache
- invalidate the cache after a hardware reset
- poll stream transitions at 2 ms initially, then retain the existing increasing backoff

## Validation
- `./build_all.sh 6.2.1`
- full Jetson rootfs deploy; d4xx 1.0.6.15 loaded on D585 Prototype
- 10-cycle combined Depth/IR/Color/IMU streaming test
- steady video first-frame delay improved from about 1.22 s to 0.98-1.00 s before the paired FW optimization
- observed stream transitions settle in 4-16 ms for video and 112-152 ms for the required IMU one-shot path


ymodlin (changes requested):
This PR is still not approved by design.

PR #658, opened 2026-09-16
[RSDEV-14843][D585][GMSL] Expose composite DPP controls

lemon6630:
## Summary
- replace the two scalar decimation CIDs with one `U32[2]` composite CID: `[enable, magnitude]`
- replace the four scalar temporal CIDs with one `U32[4]` composite CID: `[enable, alpha_x1000, delta, persistency]`
- reuse the MinZ full-block DPP XU path for atomic get/set and keep the existing CID base slots
- align the D585 decimation magnitude with the USB/FW fixed value of 2

## Scope
This PR changes only the MIPI/GMSL host driver. Firmware already exposes decimation and temporal as 38-byte DPP XU windows. Librealsense consumption is tracked separately by RSDEV-14661.

## Validation
- JP6.2.1 AArch64 `d4xx.c` compile with `-Werror`
- checkpatch: 0 errors, 0 warnings, 0 checks
- focused composite-CID contract assertions
- `git diff --check`

Jira: https://rsjira.realsenseai.com/browse/RSDEV-14843


kernel/realsense/d4xx.c
```diff
@@ -5741,6 +5709,36 @@ static const struct v4l2_ctrl_config d500_ctrl_minz = {
...
 
+static const struct v4l2_ctrl_config d500_ctrl_decimation = {
+	.ops = &ds5_ctrl_ops,
+	.id = D500_CAMERA_CID_DECIMATION,
+	.name = "D500 Decimation Configuration",
+	.type = V4L2_CTRL_TYPE_U32,
+	.min = 0,
+	.max = U32_MAX,
+	.step = 1,
+	.def = 0,
```
ymodlin, line 6565: **BUG** — auto generated by the rs-agentic bot

`.def = 0` fills *every* element of the array, so the advertised default is `[0, 0]`. But `d500_decimation_min[1] == 2`, so `d500_dpp_params_valid()` rejects the control's own default with `-EINVAL`. Any userspace reset-to-default (`v4l2-ctl -L` then write back) fails permanently.

Same at line 5735: temporal default `[0,0,0,0]` vs `d500_temporal_min[2] == 1` (delta).

Since `.def` is scalar for array types, per-element defaults need `v4l2_ctrl_new_custom()` + explicit init of `ctrl->p_def.p_u32[]`, or the min arrays must admit 0.

kernel/realsense/d4xx.c
```diff
@@ -2692,6 +2688,34 @@ static const u32 d500_minz_max[D500_DPP_XU_MINZ_PARAM_COUNT] = {
...
+static const u32 d500_minz_step[D500_DPP_XU_MINZ_PARAM_COUNT] = {
+	1, 1, 1, 1, 1, 1, 1,
+};
+
+static const u32 d500_decimation_min[D500_DPP_XU_DECIMATION_PARAM_COUNT] = {
+	0, 2,
+};
+
+static const u32 d500_decimation_max[D500_DPP_XU_DECIMATION_PARAM_COUNT] = {
+	1, 2,
```
ymodlin, line 3374: **BUG** — auto generated by the rs-agentic bot

`min[1] == max[1] == 2` pins magnitude to exactly 2. Because validation is all-or-nothing, a client sending the *previous* default (`magnitude = 4`) now gets `-EINVAL` for the whole block — the `enable` bit is silently dropped too.

If the FW value really is fixed, better to ignore/overwrite `params[1]` with 2 on write rather than fail the set, so an `enable`-only request still succeeds.

kernel/realsense/d4xx.c
```diff
@@ -2692,6 +2688,34 @@ static const u32 d500_minz_max[D500_DPP_XU_MINZ_PARAM_COUNT] = {
...
+static const u32 d500_temporal_min[D500_DPP_XU_TEMPORAL_PARAM_COUNT] = {
+	0, 0, 1, 0,
+};
+
+static const u32 d500_temporal_max[D500_DPP_XU_TEMPORAL_PARAM_COUNT] = {
+	1, 1000, 100, 8,
+};
+
+static const u32 d500_temporal_step[D500_DPP_XU_TEMPORAL_PARAM_COUNT] = {
+	1, 10, 1, 1,
```
ymodlin, line 3394: **BUG (regression)** — auto generated by the rs-agentic bot

The old scalar CID declared `step = 10` in `v4l2_ctrl_config`, and the V4L2 core *rounds* to step in `std_validate()`. The composite path advertises `.step = 1` and instead **hard-rejects** off-step values here.

So `alpha_x1000 = 405` used to be accepted (rounded to 400) and now returns `-EINVAL`. Recommend rounding to step instead of rejecting, to keep the existing userspace contract.

kernel/realsense/d4xx.c
```diff
@@ -2747,70 +2773,55 @@ static const struct d500_dpp_ctrl_desc d500_minz_desc = {
...
+				  const u32 *params)
 {
 	u8 i;
 
-	for (i = 0; i < D500_DPP_XU_MINZ_PARAM_COUNT; i++)
-		if (params[i] < d500_minz_min[i] ||
-		    params[i] > d500_minz_max[i])
+	for (i = 0; i < desc->param_count; i++)
+		if (params[i] < desc->min[i] || params[i] > desc->max[i] ||
+		    (params[i] - desc->min[i]) % desc->step[i])
```
ymodlin, line 3506: Enhancement — auto generated by the rs-agentic bot

`desc->step` is dereferenced unconditionally; a future desc that omits `.step` gives a NULL deref (and a `0` step gives div-by-zero) in kernel context. Cheap guard: `desc->step[i] && (...) % desc->step[i]`.

Also, since `QUERY_EXT_CTRL` advertises `0..U32_MAX step 1`, every rejection here is an opaque `-EINVAL` to userspace — a `dev_dbg()` naming the offending index/value would make this debuggable on a rig.

kernel/realsense/d4xx.c
```diff
@@ -3631,42 +3642,28 @@ static int ds5_s_ctrl(struct v4l2_ctrl *ctrl)
...
 			goto unlock;
 		}
-		if (!d500_minz_params_valid(ctrl->p_new.p_u32)) {
+		if (!d500_dpp_params_valid(dpp_desc, ctrl->p_new.p_u32)) {
 			ret = -EINVAL;
 			goto unlock;
 		}
-		for (i = 0; i < D500_DPP_XU_MINZ_PARAM_COUNT; i++)
+		for (i = 0; i < dpp_desc->param_count; i++)
 			control.params[i] = cpu_to_le32(ctrl->p_new.p_u32[i]);
```
ymodlin, line 4496: **Possible BUG** — auto generated by the rs-agentic bot

The write now always pushes the full 38-byte `struct d500_dpp_xu_control`, with `params[param_count..7]` zeroed by the designated initializer. The previous decimation/temporal path did a read-modify-write and preserved those trailing dwords.

For decimation that is 24 bytes of the XU window blindly zeroed. Fine only if FW ignores everything past `param_count` — worth confirming on the pending HW round-trip before merge.

kernel/realsense/d4xx.c
```diff
@@ -2653,13 +2653,9 @@ enum ds5_ae_type {
...
-#define D500_CAMERA_CID_MINZ			(DS5_CAMERA_CID_BASE + 44)
-#define D500_CAMERA_CID_DECIMATION_ENABLE	(DS5_CAMERA_CID_BASE + 49)
-#define D500_CAMERA_CID_DECIMATION_MAGNITUDE	(DS5_CAMERA_CID_BASE + 50)
-#define D500_CAMERA_CID_TEMPORAL_ENABLE		(DS5_CAMERA_CID_BASE + 51)
-#define D500_CAMERA_CID_TEMPORAL_ALPHA		(DS5_CAMERA_CID_BASE + 52)
-#define D500_CAMERA_CID_TEMPORAL_DELTA		(DS5_CAMERA_CID_BASE + 53)
-#define D500_CAMERA_CID_TEMPORAL_PERSISTENCY	(DS5_CAMERA_CID_BASE + 54)
+#define D500_CAMERA_CID_MINZ		(DS5_CAMERA_CID_BASE + 44)
+#define D500_CAMERA_CID_DECIMATION	(DS5_CAMERA_CID_BASE + 49)
+#define D500_CAMERA_CID_TEMPORAL		(DS5_CAMERA_CID_BASE + 51)
```
ymodlin: Nit — auto generated by the rs-agentic bot

Double tab; misaligns with the two `#define`s above it.

ymodlin (commented):
### rs-agentic bot review — auto generated

Reviewed the composite DPP refactor against `CLAUDE.md`. The consolidation itself is right: **§V4L2 control conventions** requires a USB-XU-backed control to be a *single* read/write CID with `VOLATILE | EXECUTE_ON_WRITE`, and the new `D500_CAMERA_CID_DECIMATION` / `D500_CAMERA_CID_TEMPORAL` composites satisfy that, where the old 6 scalar CIDs did not. Dropping the read-modify-write in favour of one atomic full-block write is also the correct shape.

Found **3 likely bugs** and a few smaller items, marked inline. The main one: the advertised `.def = 0` is outside the driver's own per-element minimums for both new controls, so the default value can never actually be written.

Note per §V4L2 control conventions: the selector still needs a matching `xu_to_cid()` entry in librealsense — noted as tracked by RSDEV-14661.

Not approving (review only, no commits pushed).

kernel/realsense/d4xx.c
lemon6630: Fixed in b28d75b. Added per-slot FW defaults through custom V4L2 type operations: MinZ `[0,0,1,0,126,1,0]`, Decimation `[0,2]`, and Temporal `[0,400,20,3]`. This deliberately does not write `ctrl->p_def.p_u32[]`: in the target V4L2 core, `p_def` storage for this type is only one element and the standard U32 initializer still uses the scalar `.def` for every array slot. The custom initializer also makes `G_EXT_CTRLS` with `WHICH_DEF_VAL` return the complete per-slot defaults.

kernel/realsense/d4xx.c
lemon6630: Fixed in b28d75b. The per-slot validator now clamps values to the descriptor range before `s_ctrl`, matching the old scalar V4L2 validation behavior. A legacy magnitude of 4 (or an array default of 0) is normalized to the D585 FW-supported value 2, so the enable update is still applied.

kernel/realsense/d4xx.c
lemon6630: Fixed in b28d75b. Per-slot validation now rounds to the nearest declared step with `DIV_ROUND_CLOSEST` instead of rejecting off-step values. On the target V4L2 implementation the scalar `std_validate()` adds half a step before division, so 405 rounds to 410 rather than 400; the composite path now preserves that behavior.

kernel/realsense/d4xx.c
lemon6630: Fixed in b28d75b. Both the custom validator and the final defensive validation check descriptor pointers, element bounds, and nonzero steps before division. User-provided values are now clamped/rounded rather than rejected, so range/step inputs no longer produce an opaque `-EINVAL`; that remains only for a malformed internal descriptor.

kernel/realsense/d4xx.c
lemon6630: Fixed in b28d75b.

kernel/realsense/d4xx.c
lemon6630: Confirmed against HKR FW. `HostProxyGmsl.cpp` always receives the 38-byte payload, but `setDecimation()` consumes only `parameters[0..1]` and `setTemporal()` only `parameters[0..3]`; entries beyond `parameter_count` are ignored. The public librealsense DPP structs also specify that reserved entries MUST be zero on SET. Keeping the designated zero initialization is therefore the intended ABI behavior and avoids the old read-modify-write.

kernel/realsense/d4xx.c
```diff
@@ -2747,73 +2786,114 @@ static const struct d500_dpp_ctrl_desc d500_minz_desc = {
...
+	value = clamp(ptr.p_u32[idx], desc->min[idx], desc->max[idx]);
+	value = desc->min[idx] +
+		DIV_ROUND_CLOSEST(value - desc->min[idx], desc->step[idx]) *
+		desc->step[idx];
+	ptr.p_u32[idx] = min(value, desc->max[idx]);
+	return 0;
+}
+
+static const struct v4l2_ctrl_type_ops d500_dpp_ctrl_type_ops = {
+	.equal = d500_dpp_ctrl_equal,
```
ymodlin, line 3621: **BUG** — auto generated by the rs-agentic bot

`struct v4l2_ctrl_type_ops` is not stable across the JetPacks this repo builds (JP5/JP6 = 5.10/5.15, JP7 = 6.8/noble):

```
5.15: bool (*equal)(ctrl, u32 idx, ptr1, ptr2);   6.8: bool (*equal)(ctrl, ptr1, ptr2);
5.15: void (*init)(ctrl, u32 idx, ptr);           6.8: void (*init)(ctrl, u32 from_idx, ptr);
5.15: int  (*validate)(ctrl, u32 idx, ptr);       6.8: int  (*validate)(ctrl, ptr);
```

Two consequences on JP7:

- `.equal` / `.validate` become incompatible pointer assignments → build break under `-Werror`.
- `.init` still compiles, but 6.8 calls it **once** (`ctrl->type_ops->init(ctrl, 0, ctrl->p_cur)` in `v4l2-ctrls-core.c`) and expects it to fill `[from_idx, ctrl->elems)`. Writing only `ptr.p_u32[idx]` leaves elements 1..N-1 at 0 — decimation default becomes `[0, 0]`, temporal `[0, 0, 0, 0]`, MinZ downscale 0 — i.e. exactly the defect this commit set out to fix, silently reintroduced.

Validation listed JP6.2.1 only. Needs `LINUX_VERSION_CODE` guards like the rest of the file (cf. lines 2433 / 2489) plus a JP7 compile.

kernel/realsense/d4xx.c
ymodlin: `change approved` — per-element defaults via a custom `.init` are the right fix on 5.15.

Residual (non-blocking): `QUERY_EXT_CTRL` still advertises `default_value = 0` from the scalar `.def`, so a userspace reset-to-default writes zeros and clamps to min (temporal alpha → 0, not 400) instead of restoring `d500_temporal_def[]`. See also the new comment on the kernel-6.8 `.init` semantics.

— auto generated by the rs-agentic bot

kernel/realsense/d4xx.c
ymodlin: `change approved` — `.validate` now clamps, so `magnitude = 4` lands on 2 and the `enable` bit is no longer dropped.

— auto generated by the rs-agentic bot

kernel/realsense/d4xx.c
ymodlin: `change approved` — `DIV_ROUND_CLOSEST` against `desc->step[idx]` restores the old rounding contract (`405` → `400`).

— auto generated by the rs-agentic bot

kernel/realsense/d4xx.c
ymodlin: `change approved` — NULL-`desc` and zero-`step` guards are now in both `d500_dpp_params_valid()` and `d500_dpp_ctrl_validate()`. The `dev_dbg()` suggestion is moot now that validate clamps instead of rejecting.

— auto generated by the rs-agentic bot

kernel/realsense/d4xx.c
ymodlin: Still open — unchanged in `b28d75b`. `d500_dpp_xu_write()` still pushes the full 38-byte block with `params[param_count..7]` zeroed. Please confirm on the pending HW round-trip that FW ignores the trailing dwords.

— auto generated by the rs-agentic bot

kernel/realsense/d4xx.c
ymodlin: `change approved`

— auto generated by the rs-agentic bot

kernel/realsense/d4xx.c
lemon6630: Confirmed on D585 hardware at commit 10d6ddc (JetPack 6.2.1, FW 8.58.157.30). Both composite SET/GET round-trips succeeded with the full 38-byte payload and zeroed trailing parameters: decimation [1,4] was accepted and normalized/read back as [1,2], and temporal [1,405,20,3] as [1,410,20,3]. The original values were restored afterward. This confirms the FW ignores the reserved trailing dwords as specified.

kernel/realsense/d4xx.c
ymodlin: `change approved`

— auto generated by the rs-agentic bot

lemon6630: Fixed in 10d6ddc. The callbacks are now version-gated at Linux 6.8: JP5/JP6 keep the indexed signatures, while JP7 uses the array-wide signatures. The 6.8 init wrapper fills every element in [from_idx, ctrl->elems), validate processes [0, ctrl->new_elems), and equal compares all ctrl->elems. Verified by compiling d4xx.o with -Werror against NVIDIA jetson_38.2 Linux 6.8 sources and the JP7 toolchain; the JP6.2.1 full build and D585 hardware round-trip also pass.

lemon6630:
Revalidated after rebasing onto `dev` (`bc4e836`). The PR contains one commit (`1671083`) and remains scoped to decimation/temporal; MinZ code and behavior are unchanged from `dev`.

Validation:
- JetPack 6.2 full clean build: PASS
- JetPack 6.2.1 full build: PASS
- JetPack 7.0 / Linux 6.8 `d4xx.o` build with `-Werror`: PASS
- Jetson 5.15.148-tegra: deployed the rebased module and verified composite GET/SET round trips
- Decimation `[1,4]` normalized to `[1,2]`
- Temporal `[1,405,20,3]` normalized to `[1,410,20,3]`
- MinZ default remains `[0,0,0,0,0,0,0]`
- Device controls restored to `[0,2]` and `[0,400,20,3]`
- `git diff --check`: PASS
- All 8 review threads resolved; no new unresolved comments before push
- JP 6.2 integration rootfs generated and archive integrity verified

kernel/realsense/d4xx.c
ymodlin: `change approved` — HW round-trip on D585/FW 8.58.157.30 settles it: FW ignores the reserved trailing dwords, so the full-block write with zeroed `params[param_count..7]` is correct. Thread closed.

— auto generated by the rs-agentic bot

kernel/realsense/d4xx.c
ymodlin: `change approved` — re-verified against the real cores, since my earlier approval on this thread predated `10d6ddc`.

- v5.15 `v4l2-ctrls-core.c`: `type_ops->init(ctrl, idx, p_cur/p_new)` per element, `validate(ctrl, idx, ...)` per element, `equal(ctrl, idx, ...)` — matches the `#else` branch.
- v6.8: `init(ctrl, 0, ctrl->p_cur)` called **once** then `cur_to_new()`, and `v4l2_ctrl_type_op_validate()` loops `new_elems` itself — matches the `>= 6.8` branch filling `[from_idx, elems)` and iterating `new_elems`.
- `DIV_ROUND_CLOSEST` matches `ROUND_TO_RANGE`'s add-half-step-then-truncate (405/step 10 → 410 in both).

Nit, non-blocking: the signature change actually landed in v6.4 (v6.6 already has the array-wide form), so the `6.8` boundary would misbuild on a 6.0–6.7 kernel. No such target in this repo (5.10 / 5.15 / 6.8), so it is correct as written.

— auto generated by the rs-agentic bot

ymodlin:
### rs-agentic review — push `6e34c576` (synchronize)

**No new code in this push.** `6e34c576` is a squash of `b28d75b` + `10d6ddc`; tree `681e50d` is byte-identical to the previously reviewed `10d6ddc` (`git diff 10d6ddc 6e34c576` is empty). Only the commit message changed.

All 7 review threads are resolved. The two that still had an unanswered author reply are now answered:
- 38-byte full-block write — `change approved` (HW round-trip confirms FW ignores the reserved trailing dwords)
- Linux 6.8 `v4l2_ctrl_type_ops` gating — `change approved`, re-verified against the v5.15 and v6.8 cores

Re-scanned the full diff for new defects: none found. Two non-blocking residuals already on record:
- `QUERY_EXT_CTRL` still advertises `0..U32_MAX step 1 def 0` from the scalar config; a userspace reset-to-default writes zeros and clamps to min rather than restoring `d500_*_def[]`.
- `d500_dpp_params_valid()` would reject an off-grid `max` after `.validate` clamps to it. No current descriptor has an off-grid `max`, so it is latent only.

Still a draft, and `behind` dev — rebase before merge. Not approving, per instruction.

— auto generated by the rs-agentic bot

kernel/realsense/d4xx.c
```diff
@@ -2747,73 +2786,169 @@ static const struct d500_dpp_ctrl_desc d500_minz_desc = {
...
+		return &d500_decimation_desc;
+	case D500_CAMERA_CID_TEMPORAL:
+		return &d500_temporal_desc;
 	default:
-		return -ENOENT;
+		return NULL;
 	}
 }
 
+/* V4L2 U32 arrays otherwise apply one scalar default/range to every slot. */
```
ymodlin, line 3523: Guideline (CLAUDE.md → "Post-patch instruction hygiene") — auto generated by the rs-agentic bot

This PR establishes a reusable convention that CLAUDE.md's **V4L2 control conventions** section does not cover: per-slot range/default for a `V4L2_CTRL_TYPE_U32` array needs custom `type_ops` (`.init` + `.validate`), because `v4l2_ctrl_config.min/max/step/def` are scalar and apply to *every* slot — and the callback signatures must be `LINUX_VERSION_CODE`-gated at 6.8. CLAUDE.md requires writing a newly-exposed convention down rather than leaving it implicit in code.

Please add it to CLAUDE.md in this PR — Sonnet, effort low (doc-only, no code risk).

ymodlin:
### rs-agentic review — `review_requested` re-review @ `6e34c576`

**No new code since my last pass.** Head is still `6e34c576` (1 commit); this is a re-review request, not a push. All 7 prior threads remain resolved.

**Independent re-verification** (this round I checked the claims against the real kernel sources rather than re-reading the diff):
- `struct v4l2_ctrl_type_ops` — v5.10 and v5.15 both use the indexed form, v6.8 uses the array-wide form (`equal(ctrl, p1, p2)`, `init(ctrl, from_idx, ptr)`, `validate(ctrl, ptr)`). The `#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 8, 0)` split matches all three JetPack generations this repo builds.
- `def_to_user()` loops `elems` calling `init(ctrl, idx, p_new)` on 5.15, and calls `init(ctrl, 0, p_new)` once on 6.8 — both satisfied by the wrappers, so `G_EXT_CTRLS` + `WHICH_DEF_VAL` does return the full per-slot defaults on 5.15 *and* 6.8. Author's claim confirmed.
- `v4l2_ctrl_new()`: 5.15 inits `p_cur`/`p_new` per element; 6.8 does `init(ctrl, 0, p_cur)` + `cur_to_new()`. Both correct here.
- `DIV_ROUND_CLOSEST` vs `ROUND_TO_RANGE` (add half-step, clamp, truncate): identical for the shipped descriptors.
- Descriptor tables vs the old scalar CIDs: enable `0..1/0`, alpha `0..1000 step 10 /400`, delta `1..100/20`, persistency `0..8/3` all preserved; only decimation magnitude intentionally narrows `2..4 def 4` → `2..2 def 2`.
- No duplicate `DS5_CAMERA_CID_BASE + N` slots, no stale `param_index` references, `param_count` ≤ `params[8]` for all three descriptors.

**New findings: 1** (guideline, non-blocking) — CLAUDE.md is not updated with the array-control `type_ops` convention this PR introduces. Inline comment posted.

**Residuals already on record** (both non-blocking, no action needed now):
- `QUERY_EXT_CTRL` still reports the scalar `default_value = 0`; compound clients should use `WHICH_DEF_VAL`, which is correct.
- `d500_dpp_params_valid()` would reject an off-grid `max` after `.validate` clamps to it — latent only, no current descriptor has one.

**Merge blockers:** branch is `behind` dev — rebase first. CI: 6 builds green, 5 still pending (incl. JP7) at time of writing.

Not approving, per instruction.

Recommended next actions:
- Rebase onto `dev` and re-run CI — any tier / inherit session; mechanical.
- Add the array-control convention to CLAUDE.md — Sonnet, effort low; doc-only.
- Re-check the 5 pending builds before merge — any tier / inherit session; pass/fail read only.

— auto generated by the rs-agentic bot

kernel/realsense/d4xx.c
ymodlin: `change approved` — the CLAUDE.md bullet added in `6b85a80` covers it: custom `v4l2_ctrl_type_ops` for heterogeneous `U32` array slots, `.init` for per-slot defaults, `.validate` for per-slot clamp/round, signatures gated at 6.8. Accurate, and correctly scoped to "this repo's supported kernels" (the upstream signature change actually landed in 6.4, but 5.10/5.15/6.8 are the only targets here).

— auto generated by the rs-agentic bot

kernel/realsense/d4xx.c
ymodlin: For the record: the **MinZ half of this fix was withdrawn** in `874d569` (carried into `6b85a80`). MinZ is now byte-identical to `dev` — no `.type_ops`, `d500_minz_def[]`/`d500_minz_step[]` deleted, and the hand-rolled `d500_minz_params_valid()` hard-reject restored. Confirmed: no MinZ line appears as `+`/`-` in the PR diff vs base.

The decimation/temporal half (per-slot defaults via `.init` + `desc->def`) stands, so this PR's own controls keep the fix. Consequence: the MinZ reset-to-default defect reverts to pre-existing `dev` behavior — scalar `.def = 0` fills all 7 slots, but `d500_minz_min[2] == 1` (downscale ratio), so writing the advertised default returns `-EINVAL`.

Not a regression from this PR and out of its narrowed scope — flagging only so it is not lost with the scope cut. Worth a separate ticket.

— auto generated by the rs-agentic bot

ymodlin:
### rs-agentic review — pushes `874d569` → `6b85a80` (synchronize)

Two pushes landed during this pass. Reviewed both against the previously approved `6e34c576`.

**`874d569` — rebase onto `dev` (`dc2ce3e`) + scope cut to decimation/temporal.** No new logic. The entire delta vs `6e34c576` is MinZ being removed from the shared composite path: `.type_ops` dropped from `d500_ctrl_minz`, `d500_minz_def[]`/`d500_minz_step[]` deleted, `D500_CAMERA_CID_MINZ` removed from `d500_dpp_ctrl_desc()`, and the hand-rolled MinZ blocks restored in `ds5_s_ctrl`/`ds5_g_volatile_ctrl`. Verified this is an exact revert, not a rewrite — no MinZ line appears as `+`/`-` in the diff vs base, so MinZ (including its locking: inside `state->lock`, `goto unlock`) is byte-identical to `dev`. The decimation/temporal composite path is unchanged from the version already approved.

**`6b85a80` — CLAUDE.md only.** `d4xx.c` is untouched between the two pushes. Adds the array-control `type_ops` convention to **V4L2 control conventions**, closing the one open thread.

**New findings: 0.** All 8 threads are now resolved or answered.

Re-verified this round:
- Compile integrity of the revert — no dangling refs to the deleted `d500_minz_step`/`d500_minz_def`; `d500_minz_desc`/`d500_minz_min`/`_max` all still used.
- `d500_dpp_ctrl_desc()` no longer returning MinZ is safe — MinZ no longer installs the type ops, so `init_elem`/`validate_elem` are never reached with a NULL desc for it.
- Rounding contract still holds on the shipped descriptors: temporal alpha `405` → `410`, decimation magnitude `4` → `2`, and both results are on-grid for the final `d500_dpp_params_valid()` check.
- **No CID collision with the incoming `dev` commit `bc4e836`** (+781 in `d4xx.c`): it still carries the scalar CIDs at `+49..+54`, which this PR replaces with composites at `+49`/`+51`, freeing `+50`/`+52`/`+53`/`+54`. No other CID slot moved.

Carried over, both non-blocking and unchanged:
- `QUERY_EXT_CTRL` still advertises the scalar `default_value = 0`; compound clients must use `G_EXT_CTRLS` + `WHICH_DEF_VAL`, which returns the correct per-slot defaults.
- `d500_dpp_params_valid()` would reject an off-grid `max` after `.validate` clamps to it — latent only; no shipped descriptor has one.

One item the scope cut leaves behind, noted on the original thread: MinZ's reset-to-default still fails on `dev` (`.def = 0` on all 7 slots vs `d500_minz_min[2] == 1` → `-EINVAL`). Pre-existing, not this PR's regression, but it loses the fix it briefly had here.

**Merge blockers:** branch is `behind` again — `dev` moved to `bc4e836` after this push. GitHub reports no textual conflict, but `bc4e836` rewrites large parts of `d4xx.c`, so rebase and re-run CI. Checks at time of writing: Aikido green, 10 `build` jobs still in progress.

Not approving, per instruction.

Recommended next actions:
- Rebase onto `bc4e836` and re-run CI — Sonnet, effort medium; textually clean per GitHub, but `bc4e836` touches `d4xx.c` heavily, so the merge result needs a real read rather than a blind `--continue`.
- Confirm the 10 pending `build` jobs (incl. JP7/6.8) go green — any tier / inherit session; pass/fail read only.
- Open a follow-up ticket for the MinZ per-slot default defect on `dev` — Sonnet, effort low; ticket prose only, the diagnosis is already written up above.

— auto generated by the rs-agentic bot

kernel/realsense/d4xx.c
ymodlin: `change approved` — CLAUDE.md line 138 now records the convention (custom `type_ops` for heterogeneous `U32` array slots, `.init` for per-slot defaults, `.validate` for per-slot clamp/round, gated at Linux 6.8).

— auto generated by the rs-agentic bot

ymodlin:
### rs-agentic review — push `6b85a80` (synchronize)

Re-reviewed the delta since my last pass (`6e34c57`). Two things changed: a **rebase onto `dc2ce3e`** and a **scope narrowing — MinZ is now completely untouched by this PR**. `git diff dc2ce3e..6b85a80` is 456 lines and confined to decimation/temporal + one CLAUDE.md line.

**What's new in this push**
- `d500_minz_desc` no longer carries `.min/.max/.step/.def`, `d500_ctrl_minz` no longer uses `.type_ops`, `d500_minz_step[]`/`d500_minz_def[]` are gone, and `d500_minz_params_valid()` + the dedicated MinZ `s_ctrl`/`g_volatile_ctrl` blocks are back. Verified against `dc2ce3e`: **every MinZ line in the PR head is byte-identical to `dev`** — no behavior change, and no stale references (`param_index`, `d500_minz_step`, `d500_minz_def`, the old six `*_ENABLE/_MAGNITUDE/_ALPHA/...` CIDs are all gone; `+49` and `+51` are the only CID slots used, no collisions).
- CLAUDE.md line 138 adds the array-control `type_ops` convention — closes my one open thread (`change approved`, thread resolved).

**New findings: 0.** The composite decimation/temporal path is unchanged from the version I verified against the v5.15 and v6.8 `v4l2-ctrls-core.c` sources last round: the `6.8` type-ops gate, the `DIV_ROUND_CLOSEST` rounding contract, and the descriptor ranges (`enable 0..1/0`, `alpha 0..1000 step 10/400`, `delta 1..100/20`, `persistency 0..8/3`, decimation magnitude intentionally `2..2 def 2`) all still hold. All 8 review threads are now resolved.

**One note, not a request for change in this PR.** Narrowing back to `dev`'s MinZ restores a pre-existing `dev` defect that the previous head had incidentally fixed: `d500_ctrl_minz` advertises scalar `.def = 0` for all 7 slots, but `d500_minz_min[D500_MINZ_DOWNSCALE_RATIO] == 1`, so `d500_minz_params_valid()` rejects the control's own advertised default — a userspace reset-to-default on MinZ returns `-EINVAL`. It is **not** a regression introduced here (it is `dev`'s behavior today) and narrowing the PR was the right call, but the `d500_dpp_ctrl_type_ops` machinery this PR adds is now exactly what MinZ would need. Worth a follow-up ticket rather than reopening this PR's scope.

**Merge blockers:** branch is still `behind` — head sits on `dc2ce3e` while `dev` is at `bc4e836` ([RSDEV-13342] MIPI driver descriptors, +752 in `d4xx.c`). Its `d4xx.c` hunks and its CLAUDE.md hunks (lines ~95 and ~180) do not overlap this PR's regions, so the rebase should be mechanical — but re-run CI after it. CI at time of writing: Aikido pass, 10 builds still pending.

Not approving, per instruction.

— auto generated by the rs-agentic bot

ejgoldik (approved):

ymodlin:
`change approved` — re-reviewed `16710831` (force-push at 09:31Z).

This push is a **pure rebase** of `6b85a80` onto the new `dev` (`bc4e836`, RSDEV-13342). Verified by interdiff: the added/removed lines are byte-identical to the previously reviewed head; only hunk offsets moved (~680 lines inserted earlier in `d4xx.c`). No new code to review, and all 8 prior threads stay resolved.

Rebase-interaction checks, all clean:
- `dc2ce3ed..bc4e836` touches only format tables / probe / descriptor parsing — no overlap with `ds5_s_ctrl`, `ds5_g_volatile_ctrl` or `ds5_ctrl_init`.
- Dev added no new CIDs, so CID base slots 49/51 are still uncontested; no stale refs to the six removed scalar CIDs remain in-tree.
- The CLAUDE.md bullet re-landed correctly under **V4L2 control conventions** (line 138) despite dev rewriting that file.
- MinZ is still byte-identical to `dev`, so the `874d569` scope cut survived the rebase.
- No conflict markers, no duplicate symbol definitions, `git diff --check` clean.

Dev's new I2C guideline ("never clock more bytes than FW loads") does not apply here: the 38-byte DPP XU windows at `0x4540`/`0x4568` are FW-implemented and D585-gated, and were already read by the pre-existing scalar path — no new exposure.

— auto generated by the rs-agentic bot

PR #660, opened 2026-09-16
[GMSL] Sync camera format descriptors into dev_gmsl_merge

polejoe:
## Summary

- Sync the `dev` delta after the metadata snapshot already ported by PR #653.
- Bring the camera-provided MIPI format/profile descriptor parser from `dev` commit `bc4e836` into `dev_gmsl_merge`.
- Include the intervening RGB calibration BA10 update and driver version `1.0.6.16`.
- Keep the descriptor parser's validated built-in-table fallback for cameras that do not provide a valid blob.

## Port boundary

- Integration base: `dev_gmsl_merge` at `202cf66`.
- Source range: `dev` after `900eff5` through `bc4e836`.
- PR #653 already ports the preceding metadata descriptor snapshot, so this change does not apply it again.
- The aggregate patch-id matches the source `dev` range exactly.

## Validation

- JetPack 6.2.2 patch stack reset and reapplied successfully.
- Generated `d4xx.c` contains the camera descriptor reader at I2C `0x6000` and the validated fallback path.
- Full clean JP6.2.2 build passed with no compiler or linker errors.
- `d4xx.ko`, `max96717.ko`, and `max96724.ko` all report `5.15.185-tegra` vermagic.
- The complete package contains the matching Image, Nano Super DTB, FG24 D5x overlay, and module tree.


kernel/realsense/d4xx.c
```diff
@@ -7258,8 +7948,33 @@ static int ds5_fixed_configuration(struct i2c_client *client, struct ds5 *state)
...
+	/* The FW serves only the formats valid for the serdes link mode, which
+	 * the host alone knows: tell it before asking for the descriptor.
+	 * Non-fatal, as in ds5_hw_init(): FW without this register rejects the
+	 * write, and the descriptor read then simply falls back.
+	 */
+	if (state->d58x_pixel_mode &&
+	    ds5_write(state, DS5_MIPI_SERDES_PIXEL_MODE, 1)) {
+		dev_warn(&client->dev,
+			 "FW has no serdes_pixel_mode support (reg 0x0404)\n");
+		link_mode_configured = false;
```
ymodlin, line 7967: 🤖 **rs-agentic bot — auto-generated**

**BUG:** `DS5_MIPI_SERDES_PIXEL_MODE` (0x0404) is only ever written `1` — here, in `ds5_hw_init()` and in `ds5_hw_reset_with_recovery()`. It is never written `0`.

So a tunnel-mode D58x is never *told* its mode, yet `link_mode_configured` stays `true` at L7974 and the descriptor is accepted. On a `rmmod`/`modprobe` without a camera power-cycle, 0x0404 is still `1` from the previous session → FW serves the pixel-mode IMU record (256 B) against a tunnel link. That is exactly the mismatch the `link_mode_configured` guard exists to prevent.

Suggest writing `0` explicitly on the `!d58x_pixel_mode` path, with the same non-fatal handling.

kernel/realsense/d4xx.c
```diff
@@ -1038,6 +1133,589 @@ static int ds5_raw_read(struct ds5 *state, u16 reg, void *val, size_t val_len)
...
+	const struct rs_desc_toc *ftoc = tabs[RS_DESC_T_FORMAT];
+	struct ds5_desc_stream *out;
+	unsigned int first = st->fmt_first;
+	unsigned int count = st->fmt_count;
+	unsigned int i;
+
+	if (!ds5_desc_stream_valid(st->stream_id)) {
+		dev_warn(&state->client->dev,
+			 "format descriptor: stream id %u unknown\n", st->stream_id);
+		return -EINVAL;
```
ymodlin, line 1455: 🤖 **rs-agentic bot — auto-generated**

**BUG (forward compat):** an unknown `stream_id` returns `-EINVAL`, which propagates through `ds5_desc_build()`'s `goto err` and discards the **entire** descriptor → every camera silently falls back to the static tables.

This contradicts the versioning contract stated elsewhere in the same parser: unknown *table types* are skipped by design (L1341), unknown *pixfmts* are skipped (L1489), and record strides are designed so newer FW can append fields. A FW that adds one new stream id would brick the descriptor for every older driver.

Same concern one level down: a stream whose formats are *all* unknown pixfmts hits `return out->n_formats ? 0 : -EINVAL;` and rejects everything, despite the per-format `skipped` path.

Suggest skipping unknown stream ids and letting the existing "stream %u missing" completeness check in `ds5_desc_build()` be the single rejection point.

ymodlin (commented):
### rs-agentic bot review — auto-generated 🤖

Scope note: this PR targets `dev_gmsl_merge`, not `dev`. Reviewing it anyway since it is a `dev` → `dev_gmsl_merge` port; flagging in case the branch filter was meant to skip it.

Port itself looks faithful — the `ds5_configure()` DT rework is behaviour-preserving (`0x31`/`0x32` moved from inline forcing into `.data_type`, wire DT still `0x1E` via `ds5_wire_data_type()`), and the descriptor parser's bounds/CRC checks and module-lifetime table ownership hold up. 4 comments below: 2 BUG, 1 verify, 1 CLAUDE.md hygiene.

**Recommended follow-ups** (per CLAUDE.md "Model/effort recommendation"):
- Fix the `0x0404` tunnel-mode write — Sonnet, effort medium; small edit but it decides which IMU record width the FW serves.
- Loosen unknown-stream-id rejection — Sonnet, effort medium; forward-compat contract, no HW risk.
- Verify the `fmts_bitmap` lookup ordering on a rig — Opus, effort high; a wrong row silently mis-programs bpp/DT and only shows as capture corruption.
- CLAUDE.md L247–248 update — Sonnet, effort low; mechanical.

nvidia-oot/6.2/0013-Tegra-video-format-extended.patch
```diff
@@ -154,11 +162,51 @@ index 487ea19..7f4e64d 100644
...
+ 
++	/* A FourCC does not uniquely identify a table row. BA10 is shared by
++	 * the stock SGRBG10_1X10/RAW10 row and D58x calibration's
++	 * SGRBG16_1X16/RAW16 row. The subdevice advertises the applicable
++	 * media-bus code in fmts_bitmap, so select that row before using the
++	 * legacy first-match fallback.
++	 */
++	for (i = offset; i < chan->num_video_formats; ++i) {
++		if (chan->video_formats[i]->fourcc == fourcc &&
++		    test_bit(i, chan->fmts_bitmap))
```
ymodlin, line 180: 🤖 **rs-agentic bot — auto-generated**

**Please verify:** `chan->fmts_bitmap` is populated by `tegra_channel_fmts_bitmap_init()` only after the subdev binds. Any `tegra_core_get_code_by_fourcc()` / `tegra_core_get_format_by_fourcc()` call made before that (channel init, default-format setup) sees an all-zero bitmap and falls through to the legacy first-match — which now resolves `BA10` to the stock `SGRBG10_1X10`/RAW10 row, i.e. wrong `bpp`, wrong CSI DT and no override for the D58x calibration surface.

If that ordering can happen, the fallback needs to be bitmap-aware (or the row needs a distinct fourcc). Worth confirming on the rig rather than by inspection — the failure mode is silent capture corruption, not an error.

Same hunk in `nvidia-oot/7.0/0013-…patch`.

CLAUDE.md
```diff
@@ -177,6 +180,9 @@ Each MIPI segment's lane count comes from a **different** DT property, so the ca
...
+- Parsed format-descriptor tables (`struct ds5_desc`, module-wide list `ds5_desc_all`) are **module-lifetime** — never freed on slot re-init — because a sibling subdev can outlive the primary that owns the `ds5_dev` slot (sysfs unbind) while its `sensor->formats` still points into them. `ds5_dev->desc` is written once under `lock` before any subdev registers; afterwards only the `sensor->formats` pointers into the tables are used, so nothing on a stream path touches the lock. Whether embedded metadata is captured stays a **host** decision (DT `embedded_metadata_height` → `metadata_enabled` → `DS5_*_STREAM_MD`); the descriptor does not carry it.
```
ymodlin, line 185: 🤖 **rs-agentic bot — auto-generated**

**Guideline violation** — CLAUDE.md "Post-patch instruction hygiene": *"Check for stale architectural claims and remove or correct them immediately."*

This PR flips the RGB-calib fourcc `GR16` → `BA10` and makes the fourcc lookup subdev-filtered, but leaves two claims in this same file stale:

- **L247:** *"the RGB calib surface is the `V4L2_PIX_FMT_SGRBG16` ('GR16', upstream fourcc) override row"*, and *"duplicate **fourccs** are not [fine]: `tegra_core_get_format_by_fourcc()` is first-match and not subdev-filtered, which is why `SGRBG10` was rejected — the RAW10 row already owns it."* Both are now inverted by this change.
- **L248:** *"Both D58x calib streams (GR16 and Y16I) are captured via the override"* — GR16 no longer exists.

Please update both in this PR.

PR #661, opened 2026-09-16
Fix dfu infinite loop (re-push of #359 by @adanila-cti)

ymodlin:
Re-push of #359 (by @adanila-cti) from an org-owned branch, at the identical head commit `25f0067`.

## Why a new PR

#359 comes from a fork and its author is not on the Jenkins GHPRB whitelist, so the required
`D4xx_Kernel_Module_Jetson_JP6` check can never be produced on it: `MIPI_driver_pr_monitor` fires
the synthetic webhook correctly (builds 4614 and 4616 both logged `[ready] PR #359 ... HTTP 200`),
but GHPRB drops it and comments *"Can one of the admins verify this patch?"* instead of building.
The `ok to test` / `add to whitelist` phrases can't help either — GHPRB's phrase handling needs the
`issue_comment` webhook, and that hook is disabled on this repo. No GitHub Actions workflow produces
that context on `pull_request` (`merge-queue-gate.yml` is `on: merge_group` only), so an org-owned
branch is the only path to a real driver-CI run.

## Contents

Unchanged from #359 — the original commit by Alex Danila &lt;adanila@connecttech.com&gt; is preserved:

```
 kernel/realsense/d4xx.c | 4 ++--
```

Both DFU bus-clock loops in `ds5_dfu_device_open()` and `ds5_dfu_device_release()` re-evaluated
`i2c_parent_is_i2c_adapter(state->client->adapter)` inside the loop body instead of walking the
chain, so `parent` never advanced and the loop spun forever on any nested I2C adapter. The fix
passes `parent` so the walk terminates.

Credit to @adanila-cti. Closes #359 on merge.


ipilchin1 (approved):

kernel/realsense/d4xx.c
```diff
@@ -8625,7 +8625,7 @@ static int ds5_dfu_device_open(struct inode *inode, struct file *file)
 #if defined(CONFIG_TEGRA_CAMERA_PLATFORM) && LINUX_VERSION_CODE < KERNEL_VERSION(6, 8, 0)
 	/* get i2c controller and set dfu bus clock rate */
 	while (parent && i2c_parent_is_i2c_adapter(parent))
-		parent = i2c_parent_is_i2c_adapter(state->client->adapter);
+		parent = i2c_parent_is_i2c_adapter(parent);
```
ymodlin, line 8628: 🤖 *rs-agentic bot (auto-generated)*

Fix confirmed — the walk terminates.

Non-blocking, pre-existing: seeding from `i2c_parent_is_i2c_adapter(client->adapter)` skips one level, so a DT where the client sits directly on the root controller gives `parent == NULL` and the DFU clock is never applied. Seeding with the adapter itself covers both topologies, drops the `!parent` branch, and avoids calling the helper twice per iteration:

```c
struct i2c_adapter *parent = state->client->adapter;

while (i2c_parent_is_i2c_adapter(parent))
	parent = i2c_parent_is_i2c_adapter(parent);
```

kernel/realsense/d4xx.c
```diff
@@ -8839,7 +8839,7 @@ static int ds5_dfu_device_release(struct inode *inode, struct file *file)
 #if defined(CONFIG_TEGRA_CAMERA_PLATFORM) && LINUX_VERSION_CODE < KERNEL_VERSION(6, 8, 0)
 	/* get i2c controller and restore bus clock rate */
 	while (parent && i2c_parent_is_i2c_adapter(parent))
-		parent = i2c_parent_is_i2c_adapter(state->client->adapter);
+		parent = i2c_parent_is_i2c_adapter(parent);
 	if (!parent) {
```
ymodlin, line 8843: 🤖 *rs-agentic bot (auto-generated)*

Non-blocking, pre-existing: this early return also skips the `DS5_FW_VERSION` / `DS5_FW_BUILD` re-read below, so on a no-mux DT (`parent == NULL`) the cached FW version is never refreshed after a DFU session. Same root cause as the seeding note in `ds5_dfu_device_open()`.

ymodlin (commented):
🤖 *Auto-generated review by the **rs-agentic bot** — not a human review.*

**Verdict: the fix is correct.** Both loops now advance `parent` up the adapter chain and terminate at the root controller. The `#if ... LINUX_VERSION_CODE < KERNEL_VERSION(6, 8, 0)` guard is preserved, so JP7/noble is unaffected (CLAUDE.md → *Kernel ABI / module compatibility*). No new bug found in the diff.

Scope note: on the shipped Jetson overlays the client sits one level below a `pca9548`, so the loop body never executed and the hang only reproduced on nested-mux carriers — consistent with the PR description.

Two non-blocking notes inline (both pre-existing, not introduced here).

Out-of-diff bug, FYI only — `ds5_dfu_device_open()` increments `device_open_count` before the `devm_kzalloc()` of `dfu_msg`; on `-ENOMEM` it returns without decrementing, and VFS does not call `release()` for a failed `open()`, so `/dev/d4xx-dfu*` stays `-EBUSY` until module reload. Separate fix.

ymodlin (commented):
🤖 *rs-agentic bot — superseded.*

Duplicate of the review immediately above (same run, two workers). The substantive findings live there; my inline copies have been removed to keep the thread clean.

Independent conclusion was identical: **fix is correct**, both occurrences covered, no CLAUDE.md violation, version guard preserved. Not approving.

PR #531, opened 2026-07-28
Expose GMSL RGB scalar V4L2 controls

lemon6630:
Firmware dependencies:

- https://github.com/RealSenseAI-Internal/soc-rs-soc-io-drivers/pull/478
- https://github.com/RealSenseAI-Internal/soc-rs-soc-device-mgr/pull/414

## What changed

- Expose Saturation, Sharpness, White Balance Temperature, signed Hue, Auto White Balance, Power Line Frequency, and Exposure Auto Priority on the RGB V4L2 subdevice.
- Use RGB direct registers `0x4230..0x423C`.
- Add SET and volatile GET handling, including signed 16-bit Hue conversion.
- Register the requested ranges, defaults, White Balance step 10, and Power Line menu.
- Mark every control volatile and execute-on-write, matching Brightness, Contrast, and Gamma.

## Why

The D5xx tunnel Host Driver did not expose the seven RGB scalars implemented by the firmware direct-register path, so applications could not access them through standard V4L2 CIDs.

## Impact

The RGB subdevice provides the standard Linux control surface with live firmware readback. Existing Depth controls and non-RGB nodes are unchanged.

## Validation

- `./build_all.sh --clean --d5xx --fg 6.2.1`: PASS
- Deployed on Orin NX, kernel `5.15.148-tegra`
- Loaded module version `1.0.3.9`; GMSL Link A locked
- Safety Camera 1.2 firmware probe `8.58.132.126`
- Seven valid V4L2 SET/GET operations and direct-register bytes passed
- 13 raw invalid firmware writes preserved the previous values

RGB buffers currently carry an existing VI error flag with repeated `corr_err`; this PR claims control/register validation, not resolution of that separate capture-quality issue.

PR #662, opened 2026-09-17
[RSDEV-13850]: Request a 550.4 MHz VI clock floor during D5xx streaming on JP6.2

pinquanw:
RealSense video + IMU capture can run with a low VI clock under the throughput-only clock policy. Request 550.4 MHz while an active D5xx stream is present on JetPack 6.2.x; release the request when the last D5xx stream stops.

Tracked on [RSDEV-13850](https://rsjira.realsenseai.com/browse/RSDEV-13850).

- Match only `realsense,d5xx` DT sensor nodes. D4xx-only capture retains the throughput policy; mixed D4xx/D5xx capture shares the VI clock until the last D5xx stream stops.
- Compute `max(throughput_rate, floor)` directly inside the existing camera clock policy under `device_list_mutex`. Only entries allocated and owned by this module are converted to its private wrapper. No accessor recovers private state from a caller-supplied public struct pointer; `vi5_set_rate()` remains upstream behavior.
- Preserve the public `tegra_camera_dev_info` layout and existing exported signatures. One new sensor-registration export carries device identity; the private wrapper stores only eligibility.
- Pass the request to the existing `clk_set_rate()` path, which delegates supported-rate selection to the clock provider. Keep any higher throughput vote.
- Load-time opt-out: `options tegra_camera_platform d5xx_vi_min_rate=0`. Default is `550400000` Hz, read-only after load. The sysfs parameter reports the request, not the actual hardware frequency.

The patch covers the 6.2 stack and its existing aliases. It does not change scheduling priorities or pacing. Failed-start handling remains upstream: partially started subdevices must be safely stopped before releasing votes.

Validation (earlier ABI checks retained; latest follow-up recompiles the changed camera policy and channel objects):
- Regenerated the NVIDIA patch and verified clean application to the preceding R36.5 patch-stack baseline.
- Cross-compiled camera policy, channel and VI objects for ARM64 against the configured 5.15.185 kernel, GCC 11.3.
- Recomputed export CRCs with kernel `genksyms`: all 35 existing camera/channel exports match the same baseline; only the sensor-registration export is added.
- ASan/UBSan production-function harness passed: intrusive-list traversal in non-array order using the extracted private wrapper, an additional nonzero-offset layout, forward-declaration extraction checks, PG-floor ordering, null/unrelated/D4xx/D5xx identity, idle, mixed streams, partial/last close, opt-out, higher throughput vote, clock-set error propagation, and a by-value public-info copy passed into the policy callback.
- Test script imports safely under pytest collection arguments; argument parsing, source reads, compilation and execution occur only from `main()`.
- Applied C source whitespace checks passed. The existing CLAUDE.md ownership invariant was updated.

No hardware deployment or new FPS measurement is claimed. Provider rounding, stream/start/stop/restart clock readback and Viewer FPS remain target validation. Rebuild and deploy the NVIDIA module set together; the channel module depends on the new registration export. The 550.4 MHz value comes from manual Orin Nano experiments, not a universal hardware minimum.

### RSDEV-13850 integration qualification — 2026-09-18

The complete current integration has **not** been hardware-qualified. Historical 180-second results from Apps `aac7cacf` / IO `c6843ad9` and manual VI 550.4 MHz do not establish a pass for the current XU/merged revisions. Individual tests above retain their stated scope; this qualification requirement concerns the combined D585 milestone, not whether this PR is independently usable.

Firmware serialization/shaping needs Apps #339 + IO #623 + Device Manager #531 + HIF FW #3 + matching prebuilt #95. Paired delivery additionally needs Host #663 **and SDK #15680**. JP6.2 automatic clock mitigation is Host #662; the Viewer scenario additionally needs SDK #15692. Installing only the seven firmware/host PRs does not enable or decode pairs.

[IO #623](https://github.com/RealSenseAI-Internal/soc-rs-soc-io-drivers/pull/623) records the exact nine-PR candidate commits, deployment identity requirements, XU/start/stop/restart checks and acceptance contract. For four 960p60 streams: 180-second run averages of >=198 Hz per sensor when paired, >=100 Hz when legacy, with Depth/IR retaining 60 FPS and the previously accepted RGB-specific tolerance of >=59.35 FPS. Report all 5-second windows and source-counter gaps separately. The 5 ms check is at HKR pair formation; it is not a proven SDK delivery-latency bound or per-window rate guarantee. No new integrated FPS result is claimed by this update.


ymodlin (commented):
### rs-agentic automated review — PR #662 (`fix/realsense-vi-clock-floor` → `dev`)

Reviewed the single added file `nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch` against the NVIDIA R36.5 `nvidia-oot` sources, `CLAUDE.md`, and the `nvidia-oot-patch` skill.

**What checks out**
- `intel,d4xx` / `realsense,d5xx` both exist as DT `compatible` strings in `hardware/realsense/*.dts*`, and d4xx subdevs are `v4l2_i2c_subdev_init()`-based, so `subdev_on_csi->dev->of_node` carries them — the detection is live, not dead code.
- The recompute runs under `device_list_mutex` (held by `tegra_camera_update_clknbw()`), and `stream_on` is updated in the first list pass before the clock pass — so partial/last-close behaves as the description claims.
- `nvidia-oot/6.2.1` and `6.2.2` are directory symlinks to `6.2`, and `apply_patches.sh` globs `*` — the new patch is picked up without any series/list edit.
- Floor ordering in `vi5_set_rate()` is sound: `floor` is a CCF-supported rate, so `clk_set_rate()` rounding `max(rate, floor)` down still lands at or above the floor.
- The failed-start vote leak it fixes is real: `tegra_channel_set_stream()` votes at the top of the `on` branch, and the `is_streaming == on` guard makes the caller's cleanup `set_stream(false)` a no-op.

**Findings** — 1 bug, 3 guideline/enhancement items, inline.

**One question on scope:** the patch lands only in `nvidia-oot/6.2/`, but CI also builds JP6.0 and JP6.1 (`build-jp6.yml`, `build-jp6.1.yml`), which run the same throughput-only VI clock policy, and CLAUDE.md's layout rule is that `6.0/` holds the real file with `6.1`/`6.2` symlinking it. If 6.2-only is deliberate, worth stating why in the PR description; otherwise move the file to `6.0/` and symlink it from `6.1/` and `6.2/`.

Not approving — review only. The PR is still a draft and the HW validation it lists as outstanding (stream/restart/stop clock readback, Viewer FPS) is what would close out the risk here.

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
```diff
@@ -0,0 +1,125 @@
...
+index 2cf1a5a..af65ca3 100644
+--- a/drivers/media/platform/tegra/camera/vi/channel.c
++++ b/drivers/media/platform/tegra/camera/vi/channel.c
+@@ -972,6 +972,9 @@ int tegra_channel_set_stream(struct tegra_channel *chan, bool on)
+ 
+ 	if (ret == 0)
+ 		atomic_set(&chan->is_streaming, on);
++	else if (on)
++		/* A failed start must not leave the bandwidth/clock vote active. */
++		tegra_camera_update_clknbw(chan, false);
```
ymodlin: **BUG — this unwind can under-clock a partially started channel.**

In the `on == true` branch above, the subdev loop records the first failure in `ret` but keeps calling `s_stream(true)` on the remaining subdevs and never rolls back the ones that succeeded. Releasing the vote here drops `num_active_streams` / `active_pixel_rate` and the ISO-BW request while those subdevs are still streaming into the VI — so the VI can now be clocked *below* what live traffic needs (before the patch it only leaked a too-high vote).

`err_set_stream:` in `vi5_channel_start_streaming()` does not recover this: the `is_streaming == on` early return at the top of this function makes its `tegra_channel_set_stream(chan, false)` a no-op after a failed start.

Suggest stopping the subdevs that did start before releasing the vote.

_This comment was auto-generated by the **rs-agentic** bot._

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
```diff
@@ -0,0 +1,125 @@
+From a3063a66e9f6e0ab90158aae44050745d1d0ec42 Mon Sep 17 00:00:00 2001
+From: pinquanw <pinquan.wang@realsenseai.com>
+Date: Thu, 17 Sep 2026 14:16:16 +0800
+Subject: [PATCH] media: tegra: request a VI clock floor for active RealSense
+ streams
+
+---
```
ymodlin, line 32: **Guideline (CLAUDE.md).** Empty commit body and no `Signed-off-by:`. Sibling patches `0011`, `0012`, `0013` in this directory all carry both. Please regenerate with the rationale in the commit message — CLAUDE.md puts long rationale in the commit message / PR description, not in code comments.

_This comment was auto-generated by the **rs-agentic** bot._

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
```diff
@@ -0,0 +1,125 @@
...
+diff --git a/drivers/video/tegra/camera/tegra_camera_platform.c b/drivers/video/tegra/camera/tegra_camera_platform.c
+index b2b5510..892f265 100644
+--- a/drivers/video/tegra/camera/tegra_camera_platform.c
++++ b/drivers/video/tegra/camera/tegra_camera_platform.c
+@@ -23,6 +23,12 @@
+ #include <soc/tegra/fuse.h>
+ #define CAMDEV_NAME "tegra_camera_ctrl"
+ 
++/* Load-time setting; 0 restores the throughput-only VI clock policy. */
++static unsigned long realsense_vi_min_rate = 550400000;
```
ymodlin: **Enhancement.** `550400000` is an unexplained magic value. Give it a `#define` with a one-line *why* at the constant (CLAUDE.md in-code comment rule: state the why, at the site it explains) — e.g. the VI rate the D4xx/D5xx video+IMU load actually needs. The PR states this is not HW-measured yet, so that rationale is all a future reader will have.

_This comment was auto-generated by the **rs-agentic** bot._

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
```diff
@@ -0,0 +1,125 @@
...
+index 5b36749..cb66387 100644
+--- a/include/media/tegra_camera_platform.h
++++ b/include/media/tegra_camera_platform.h
+@@ -85,6 +85,9 @@ struct tegra_camera_dev_info {
+ 	u32 pixel_bit_depth;
+ 	u32 bpp;
+ 	bool stream_on;
++	bool is_realsense;
++	/* Optional VI clock floor, independent of the throughput vote. */
++	unsigned long min_clk_rate;
```
ymodlin: **Guideline (CLAUDE.md struct-ABI rule).** The two members are inserted mid-struct, shifting the offsets of `device_node` and `ops`. `tegra_camera_dev_info` is reached through the exported `tegra_camera_device_register()` / `tegra_camera_get_device_list_entry()`, so every module touching it must be rebuilt in lockstep. Appending after `ops` gets the same result with no offset churn.

Minor: `min_clk_rate` is VI-only state living on a struct shared by every registered device — worth saying in the comment that only the `HWTYPE_VI` entry uses it.

_This comment was auto-generated by the **rs-agentic** bot._

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
pinquanw: Agreed. In d768b163e I removed the newly added vote-only unwind. The original tegra_channel_set_stream() is now unchanged, so a partial start cannot lose its clock/bandwidth vote through this patch. A safe rollback would need to stop all started subdevices and handle stop failures before releasing the vote; that broader recovery change is outside this clock-floor patch.

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
pinquanw: Fixed in d768b163e. Regenerated the embedded patch with the rationale, ABI/lifetime constraints, and Signed-off-by. The amended repository commit also has a rationale and sign-off; in-code comments stay concise.

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
pinquanw: Fixed in d768b163e: REALSENSE_VI_DEFAULT_MIN_RATE_HZ names the 550400000UL default, with a one-line explanation that it comes from manual Orin Nano video/IMU receive-latency experiments. The commit/PR distinguish this tunable default from a universal hardware minimum or a frame-delivery guarantee. Hardware validation of the patch remains pending.

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
pinquanw: Agreed with the ABI concern. Appending would preserve offsets but still change sizeof and genksyms CRCs, so d768b163e instead keeps both fields in a private allocation and adds registration/callback-accessor functions. The public tegra_camera_dev_info definition is unchanged, and the private floor is documented as VI-only under device_list_mutex. ARM64 CONFIG_MODVERSIONS comparison against the same patch baseline confirms all 35 existing exports in the affected objects retain their CRCs; only two new exports are added. The updated NVIDIA modules still need deployment together for those new symbols.

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
ymodlin: change approved

_This comment was auto-generated by the **rs-agentic** bot._

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
ymodlin: change approved

_This comment was auto-generated by the **rs-agentic** bot._

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
ymodlin: change approved

_This comment was auto-generated by the **rs-agentic** bot._

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
ymodlin: change approved

_This comment was auto-generated by the **rs-agentic** bot._

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
```diff
@@ -0,0 +1,238 @@
...
++unsigned long tegra_camera_get_device_clock_floor(
++		const struct tegra_camera_dev_info *cdev_info)
++{
++	const struct tegra_camera_device *device = container_of(cdev_info,
++			struct tegra_camera_device, info);
++
++	/* set_rate callbacks run with device_list_mutex held. */
++	return device->min_clk_rate;
++}
++EXPORT_SYMBOL(tegra_camera_get_device_clock_floor);
```
ymodlin: **Enhancement.** This exported accessor does `container_of()` on a caller-supplied `tegra_camera_dev_info *`, but the already-exported `tegra_camera_get_device_list_entry()` hands out **by-value copies** of exactly that type. Any caller that pairs the two reads past the end of a stack copy. The header comment is the only guard.

Since `calculate_and_set_device_clock()` already holds `device_list_mutex` and has `device->min_clk_rate` in hand, applying the floor there (`clk_rate = max(clk_rate, device->min_clk_rate);` just before the `set_rate` call) drops this exported symbol, the `vi5.c` hunk, and the stored `min_clk_rate` field — `clk_set_rate()` rounds anyway, so nothing is lost but the explicit `clk_round_rate()` warning.

_This comment was auto-generated by the **rs-agentic** bot._

ymodlin:
Re-reviewed the amended commit. All four previous findings are addressed; verified against the R36.4.3 `nvidia-oot` sources.

**PR description is now stale** — two claims contradict the code:
- "Failed stream starts unwind their clock/bandwidth vote" — the `channel.c` teardown hunk was dropped; the commit message correctly says the opposite.
- "All NVIDIA modules using the changed internal `tegra_camera_dev_info` structure must be rebuilt" — that struct is no longer changed. The real constraint is the new exported symbols, so `tegra_camera_platform.ko` + the VI/`channel.c` modules must be installed together (ABI-safe per CLAUDE.md).

Checks that passed: all 12 hunks internally consistent (line counts + cumulative offsets); module name is `tegra_camera_platform.ko`, so the documented sysfs path is correct; `channel.c` is the only sensor registration site (vi5/isp5/nvcsi keep the old entry point); `calculate_and_set_device_clock()` is only reached from `tegra_camera_update_clknbw()` under `device_list_mutex` with real list entries, so the `container_of()` and the locking comment hold; `subdev_on_csi->dev` is already dereferenced by `tegra_channel_populate_dev_info()` on both paths, so `->of_node` adds no new NULL risk; `intel,d4xx` / `realsense,d5xx` match the DT nodes; `6.2.1`/`6.2.2` are directory symlinks to `6.2`, and `6.0`/`6.1` correctly do not pick up `0014`.

One non-blocking enhancement left inline. HW validation is still outstanding as the description states.

_This comment was auto-generated by the **rs-agentic** bot._

ghu3 (commented):
Review for RSDEV-13850 (cross-PR set: apps#339, io#623, device-mgr#531, prebuilts#95, HIF FW#3, host driver #662/#663, SDK #15680/#15692).

Two notes, no blockers:

- Only `nvidia-oot/6.2/` receives the VI clock-floor patch, while companion PRs report validation on JP7.2 as well. Please confirm the other supported JetPack patch sets don't need the equivalent.
- `container_of(cdev_info, struct tegra_camera_device, info)` in `tegra_camera_get_device_clock_floor()`/`calculate_and_set_device_clock()` assumes every `cdev_info` reaching `set_rate` is a list-allocated entry. True today; worth a comment/WARN_ON since the public `tegra_camera_dev_info` layout is preserved precisely so external callers can pass their own.

**Validation performed by reviewer:** the PR set's host unit tests (GmslImuBatchTest, test_gmsl_batch_xu.py, test_traffic_shaping/test_imu_pacing/test_imu_video_batch, test_hif_poll/test_hif_tx/test_hif_tx_publish) were built and run on host with `-Wall -Wextra -Werror -fsanitize=address,undefined` — all pass, no warnings. No full HKR image build, hardware run, or D555e/S585 USB regression was performed; findings are static unless stated.

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
```diff
@@ -0,0 +1,238 @@
...
++		of_device_is_compatible(sensor_node, "realsense,d5xx");
++
++	return camera_device_register(cdev_info, priv, is_realsense);
++}
++EXPORT_SYMBOL(tegra_camera_device_register_sensor);
++
++unsigned long tegra_camera_get_device_clock_floor(
++		const struct tegra_camera_dev_info *cdev_info)
++{
++	const struct tegra_camera_device *device = container_of(cdev_info,
```
ghu3: `container_of` assumes `cdev_info` is the embedded member of a `tegra_camera_device` allocated by `camera_device_register()`. Any caller passing a stack/own-allocated `tegra_camera_dev_info` (the public layout is intentionally preserved) would read garbage. Add a comment at the export and/or a sanity check.

test/standalone_tests/test_vi_clock_floor.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python3
+from pathlib import Path
+import argparse
+import subprocess
+import tempfile
+parser=argparse.ArgumentParser(description='Test applied production VI clock policy with mocked DT/clock providers')
+parser.add_argument('--source-root',type=Path,required=True)
+args=parser.parse_args()
```
ymodlin: **BUG — this aborts pytest collection for the whole `test/` tree.**

argparse (`required=True`), the `cc` compile and the test run all execute at *import* time. The file matches `test_*.py`, so `pytest -vs -m d457 test/` (the invocation in CLAUDE.md Testing) imports it during collection. Reproduced on this branch:

```
INTERNALERROR> File ".../test/standalone_tests/test_vi_clock_floor.py", line 8, in <module>
INTERNALERROR>   args=parser.parse_args()
INTERNALERROR> SystemExit: 2
mainloop: caught unexpected SystemExit!
```

It is an INTERNALERROR, so the entire run dies — not just this file, and `-m d457` does not help (marker filtering happens after import).

Wrap the body in `def main()` + `if __name__ == "__main__":` like `test_sync_mode_api.py` and `test_quad_depth_rgb_sync.py`, or rename it out of the `test_*.py` pattern.

_This comment was auto-generated by the **rs-agentic** bot._

test/standalone_tests/test_vi_clock_floor.py
```diff
@@ -0,0 +1,107 @@
...
+r=args.source_root
+def function(path,name):
+ s=(r/path).read_text();a=s.index(('unsigned long ' if name == 'tegra_camera_get_device_clock_floor' else 'int ' if name == 'tegra_camera_device_register_sensor' else 'static int ')+name+'(');b=s.index('\n}',a)+2;return s[a:b]
```
ymodlin: **Enhancement.** Two things vs. the sibling standalone tests: no module docstring stating what is verified and no `Usage:` block, and the 1-space indent / single-letter names / `;`-joined statements make this line hard to audit.

Also `function()` takes the **first** textual match of the signature. A forward declaration of `calculate_and_set_device_clock()` in the NVIDIA source would make it extract from the prototype to the next column-0 `}` — silently testing the wrong span rather than failing. Anchoring on the signature followed by `)\n{` would be safer.

_This comment was auto-generated by the **rs-agentic** bot._

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
```diff
@@ -0,0 +1,241 @@
...
+--- a/drivers/video/tegra/camera/tegra_camera_platform.c
++++ b/drivers/video/tegra/camera/tegra_camera_platform.c
+@@ -23,6 +23,21 @@
+ #include <soc/tegra/fuse.h>
+ #define CAMDEV_NAME "tegra_camera_ctrl"
+ 
++/* Selected from manual Orin Nano video/IMU receive-latency experiments. */
++#define REALSENSE_VI_DEFAULT_MIN_RATE_HZ 550400000UL
++
++static unsigned long realsense_vi_min_rate = REALSENSE_VI_DEFAULT_MIN_RATE_HZ;
```
ymodlin: **Enhancement.** Eligibility is now D5xx-only, but the identifiers still read `realsense_*` — `REALSENSE_VI_DEFAULT_MIN_RATE_HZ`, `realsense_vi_min_rate`, and the `/sys/module/tegra_camera_platform/parameters/realsense_vi_min_rate` name users are told to read back. The param name is user-visible once released, so `d5xx_vi_min_rate` is cheaper to pick now than later.

The PR description is also stale: it still says the patch identifies `intel,d4xx` / `realsense,d5xx` sensors.

_This comment was auto-generated by the **rs-agentic** bot._

CLAUDE.md
```diff
@@ -168,6 +168,8 @@ Each MIPI segment's lane count comes from a **different** DT property, so the ca
...
 - **max96717 and max96724 are single-source, unlike max96712.** `nvidia-oot/max96717.{c,h}` and `nvidia-oot/max96724.{c,h}` are the *only* copies that build: `apply_patches.sh` symlinks them into the JP6/JP7 `nvidia-oot/` tree **and** into the JP5 `kernel/nvidia/drivers/media/i2c/` tree ("MAX96717/MAX96724 use the same implementation on JP5 and JP6"). There is no per-generation patch carrier and no JP5 fork to port to — edit the `nvidia-oot/` file once and every JetPack gets it.
 
+- The JP6.2 VI clock floor is requested only by active `realsense,d5xx` DT sensor nodes. D4xx-only capture keeps the throughput policy; a mixed D4xx/D5xx session shares the VI clock until the last D5xx stream stops. Keep eligibility in private camera-device state and qualify any future device-family expansion separately.
```
ymodlin: **Guideline (CLAUDE.md).** Good that the post-patch hygiene rule was followed. But this bullet is device-eligibility / clock policy, not a CRC or module-compatibility note — `## NVCSI / VI capture notes` is where the other VI policy rules live. Also drop the preceding blank line so it reads as part of the list.

_This comment was auto-generated by the **rs-agentic** bot._

ymodlin (commented):
Reviewed the new commit b8b979b (D5xx narrowing + CLAUDE.md note + standalone test). The patch change itself is a clean mechanical narrowing and all pure-D5xx overlays (`...-d5xx.dts`, `fg12-4ch-d5xx`, `fg12-16ch-d5xx`, `seeed-d5xx`, `fg24-4ch-d5xx`) plus the mixed overlays do declare `compatible = "realsense,d5xx"`, so eligibility resolves. One blocking bug in the new test file, plus two minor items. The `container_of` thread from the previous round is still open.

_This comment was auto-generated by the **rs-agentic** bot._

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
ymodlin: Still open after b8b979b — the exported `tegra_camera_get_device_clock_floor()` / `container_of()` shape is unchanged, and @ghu3 raised the same hazard at line 141. Folding the floor into `calculate_and_set_device_clock()` (which already holds `device_list_mutex` and has `device->min_clk_rate`) still removes the export, the vi5.c hunk, and the stored field.

_This comment was auto-generated by the **rs-agentic** bot._

ymodlin: change approved

Verified against the R36.4.3 upstream sources: `calculate_and_set_device_clock()` is reached only from `tegra_camera_update_clknbw()`, which holds `device_list_mutex` across the whole clock loop, and `camera_device_register()` is the only `list_add()` site — so every entry the new loop walks is this module's wrapper. Export and `vi5.c` hunk are gone.

Placing the floor *after* the `info->pg_mode` override is also the correct order: `pg_mode` is a global flag that pins VI to `DEFAULT_PG_CLK_RATE`, so a floor applied earlier would be silently discarded.

_This comment was auto-generated by the **rs-agentic** bot._

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
ymodlin: Resolved by 873c4f2 — the exported accessor and the `container_of()` on a caller-supplied pointer are removed; the floor is now computed inside `calculate_and_set_device_clock()` under `device_list_mutex`, where every list entry is module-owned.

_This comment was auto-generated by the **rs-agentic** bot._

test/standalone_tests/test_vi_clock_floor.py
ymodlin: change approved

Verified on this branch: `pytest --collect-only -q -m d457 test/` now completes — `67/72 tests collected (5 deselected)`, no INTERNALERROR.

_This comment was auto-generated by the **rs-agentic** bot._

test/standalone_tests/test_vi_clock_floor.py
ymodlin: Still open. `function()` is still first-textual-match (`'static int ' + name + '('`), so a forward declaration of `calculate_and_set_device_clock()` upstream would extract from the prototype to the next column-0 `}` and silently test the wrong span. Anchor on the signature followed by `)\n{`. Module docstring / `Usage:` block are also still missing.

One more from the rewrite: the harness re-declares `struct tegra_camera_device` locally and mocks `list_for_each_entry` over a 3-element array. Both `container_of()` offsets are therefore 0 by construction, so the test cannot detect `info` ceasing to be the first member, nor a real list mis-iteration — the two things the production loop actually depends on. A static assert that `offsetof(struct tegra_camera_device, info) == 0` in the production source would be cheaper than mocking it.

_This comment was auto-generated by the **rs-agentic** bot._

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
ymodlin: PR description fixed — thanks. The naming half is still open: `REALSENSE_VI_DEFAULT_MIN_RATE_HZ`, `realsense_vi_min_rate`, and the user-visible `/sys/module/tegra_camera_platform/parameters/realsense_vi_min_rate` still read `realsense_*` while eligibility is `realsense,d5xx`-only. Cheaper to rename to `d5xx_vi_min_rate` before release than after.

(Module name in the docs checks out: `obj-m += tegra_camera_platform.o`, so the `modprobe options` line and the sysfs path are correct.)

_This comment was auto-generated by the **rs-agentic** bot._

CLAUDE.md
ymodlin: Still open. The bullet only gained a sentence; it is still under `## Kernel ABI / module compatibility notes` with a preceding blank line, and device eligibility / clock policy belongs under `## NVCSI / VI capture notes` with the other VI policy rules.

While editing: `0014` is a real file in `nvidia-oot/6.2/` with **no** symlink in `6.0/` or `6.1/`, unlike `0011`–`0013` which are real in `6.2` and symlinked from both. That 6.2-only pattern is deliberate per the PR, but CLAUDE.md's patch-layout paragraph still describes only the shared-symlink shape — worth one clause so the next patch author does not "fix" it.

_This comment was auto-generated by the **rs-agentic** bot._

ymodlin:
**Re-review of 873c4f2** (`b8b979b..873c4f2`) — the clock-floor mechanism was reworked, so I re-reviewed the production path against the real R36.4.3 `nvidia-oot` sources (patch applied cleanly, offset -69 in `channel.c`).

**No new bugs.** The refactor resolves both container_of hazards:
- Floor now computed in `calculate_and_set_device_clock()`, whose only caller `tegra_camera_update_clknbw()` holds `device_list_mutex` for the whole clock loop. `camera_device_register()` is the single `list_add()` site, so every entry walked is module-owned. Exported accessor and the `vi5.c` hunk are gone.
- Eligibility plumbing confirmed end to end: `realsense,d5xx` is a real compatible in the D5xx overlays and in `d4xx_of_match[]`; `v4l2_i2c_subdev_init()` gives the sensor subdev `dev->of_node`, and the nvcsi/TPG path gets `sd->dev = chan->csi->dev` (non-NULL), so `subdev_on_csi->dev->of_node` cannot fault on either register site.
- Exactly one `HWTYPE_VI` entry exists (`vi5.c`), registered through the legacy wrapper with `needs_d5xx_floor = false`; sensor entries are `HWTYPE_NONE` and early-return. Ordering after the `pg_mode` override is correct.
- Test no longer aborts collection: `pytest --collect-only -q -m d457 test/` → `67/72 collected`, no INTERNALERROR.

**Still open** (3 replies inline, all non-blocking): `realsense_*` → `d5xx_*` naming on the user-visible module param; the test harness `function()` first-textual-match extraction plus its offset-0-by-construction `container_of` mocks; CLAUDE.md bullet placement and the undocumented 6.2-only patch layout.

Could not verify the C-level behaviour claims directly — no compiler in this environment, so the ASan/UBSan harness result is taken as reported.

_This comment was auto-generated by the **rs-agentic** bot._

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
pinquanw: ✅ Fixed in 873c4f264 by removing the unsafe accessor entirely. The floor is applied directly inside the camera clock policy under `device_list_mutex`; private allocation access is confined to owned list entries. The public callback parameter may be a by-value copy without an out-of-bounds read. The ASan/UBSan harness covers that case, and all 35 existing export CRCs remain unchanged.

🤖 Generated by AI

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
pinquanw: ✅ Fixed in 873c4f264 using the suggested policy-level max: removed the exported floor accessor, stored min_clk_rate and vi5.c hunk. Existing `clk_set_rate()` delegates rate selection to the provider. Private list-entry access remains under device_list_mutex. A by-value public-info copy is exercised under ASan/UBSan, and the 35 existing export CRCs are unchanged.

🤖 Generated by AI

test/standalone_tests/test_vi_clock_floor.py
pinquanw: ✅ Fixed in 873c4f264: moved argument parsing, source reads, compilation and execution inside `main()` guarded by `if __name__ == "__main__"`. Importing the module with pytest-style `-m d457` arguments now completes without parsing them or running a compiler. Applied the same import-safety guard to the companion standalone XU test in #663.

🤖 Generated by AI

test/standalone_tests/test_vi_clock_floor.py
pinquanw: Addressed in 79f0e5acf. The test now has a docstring/usage and readable functions, rejects prototypes when extracting definitions, and tests an injected forward declaration. It extracts the production private wrapper and traverses intrusive lists in non-array order instead of mocking an array stride. A second run shifts info to a nonzero offset: container_of is valid for that layout, so an offset-zero production assertion would unnecessarily constrain the implementation. Both layouts pass ASan/UBSan, along with PG-floor ordering and public-copy checks.

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
pinquanw: Fixed in 79f0e5acf: D5XX_VI_DEFAULT_MIN_RATE_HZ and d5xx_vi_min_rate now name the D5xx-only policy. Updated the module parameter, test and PR instructions; the readback path is /sys/module/tegra_camera_platform/parameters/d5xx_vi_min_rate.

CLAUDE.md
pinquanw: Fixed in 79f0e5acf. Moved the eligibility/clock-floor rule into NVCSI / VI capture notes and removed the separating blank line in the ABI list. The patch-layout description now explicitly records that 0014 is 6.2-only, including directory aliases, with no 6.0/6.1 symlinks.

test/standalone_tests/test_vi_clock_floor.py
ymodlin: change approved

Verified locally on this branch: `definition()` now skips a forward declaration and extracts only the real body (tested against source with a prototype preceding the definition), and the struct pattern correctly picks `tegra_camera_device` over `tegra_camera_dev_info`. Docstring/`Usage:` added, and the `nonzero_offset` layout plus the real intrusive list (insert order 2,0,1) close the offset-0-by-construction gap.

_This comment was auto-generated by the **rs-agentic** bot._

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
ymodlin: change approved

`D5XX_VI_DEFAULT_MIN_RATE_HZ` / `d5xx_vi_min_rate` throughout, including the sysfs path.

One leftover: the PR description still documents the opt-out as `options tegra_camera_platform realsense_vi_min_rate=0` — that modprobe line no longer works.

_This comment was auto-generated by the **rs-agentic** bot._

CLAUDE.md
ymodlin: change approved

Bullet is under `## NVCSI / VI capture notes` with no stray blank line, and the 6.2-only clause is in the patch-layout paragraph. Verified on disk: `0014` is a real file only in `6.2/`, `6.2.1`/`6.2.2` are directory symlinks to `6.2`, and `6.0/`/`6.1/` have no `0014`.

_This comment was auto-generated by the **rs-agentic** bot._

ymodlin:
**Re-review of 79f0e5a** (`873c4f2..79f0e5a`) — all three remaining findings are resolved. **No new bugs.**

The only production-code change in this push is the `realsense_vi_min_rate` → `d5xx_vi_min_rate` rename; the clock-policy logic is byte-identical to the revision I reviewed at 873c4f2, so the earlier verification still stands.

Verified this round:
- **Patch integrity** — regenerated header is sound: all 11 hunks internally consistent (line counts and cumulative offsets: `tegra_camera_platform.c` 23→23, 485→498, 509→523, 532→548, 565→597, 672→704, 720→753), totalling +58/−7 and matching the diffstat. No hunk overlap.
- **Policy correctness against real upstream `nvidia-oot` sources** — `list_add()` in `tegra_camera_device_register()` is the only list-insertion site, so every entry the new loop walks is module-owned; `calculate_and_set_device_clock()` is reached only from `tegra_camera_update_clknbw()`, which holds `device_list_mutex` across the whole clock loop; `stream_on` is set on the sensor entry matched by `priv`, which is what the floor keys off. The register error path does **not** `kfree()`, so the new `container_of()` free in unregister has no mismatched counterpart.
- **Test harness** — `definition()` extractor confirmed locally: it skips a forward declaration and extracts only the real body, and picks `tegra_camera_device` over `tegra_camera_dev_info`. Upstream `clk_rate = (nr + dr) / dr` yields exactly the `100000001` baseline the test asserts. `pytest --collect-only -q -m d457 test/` → `67/72 collected (5 deselected)`, no INTERNALERROR.
- **Patch layout** — `0014` is a real file only in `6.2/`; `6.2.1`/`6.2.2` are directory symlinks; `6.0/`/`6.1/` correctly have none. Matches the new CLAUDE.md clause.

One non-blocking doc nit (inline): the PR description still documents the opt-out as `options tegra_camera_platform realsense_vi_min_rate=0`, which no longer matches the renamed parameter.

No compiler in this environment, so the ASan/UBSan and cross-compile/CRC results are taken as reported. Hardware qualification remains outstanding as the description states.

_This comment was auto-generated by the **rs-agentic** bot._

nvidia-oot/6.2/0014-RealSense-VI-stream-clock-floor.patch
ymodlin: change approved

The opt-out line now reads `options tegra_camera_platform d5xx_vi_min_rate=0`. This thread is fully resolved.

_This comment was auto-generated by the **rs-agentic** bot._

ymodlin:
**Push review — `770c461`**

This push squashed the series onto current `dev`. The tree is **byte-identical** to the previously reviewed `79f0e5acf` (verified: `CLAUDE.md`, `0014-...patch`, `test_vi_clock_floor.py` all diff clean), so the only change is the commit itself — which now carries a proper rationale body and `Signed-off-by`.

**No new findings.** All 10 review threads are resolved; the last leftover (PR description opt-out line) is fixed in the updated body.

Re-verified the final state against the R36.4.3 upstream sources rather than re-reading the diff:
- `chan->subdev_on_csi->dev->of_node` (channel.c) adds no NULL-deref risk — `tegra_channel_populate_dev_info()` already dereferences `->dev` on both call paths, including the `pg_mode`/`nvcsi` branch. `of_device_is_compatible()` is NULL-node safe, and an NVCSI/TPG `of_node` correctly fails the `realsense,d5xx` match.
- Floor release ordering is correct: `tegra_camera_update_clknbw()` sets `cdev->stream_on` in its first list pass and recalculates every device in the second, both under `device_list_mutex` — so the floor is dropped on the last D5xx stop.
- `kfree(container_of(cdev, struct tegra_camera_device, info))` is safe — `tegra_camera_device_unregister()` only frees entries it found on `device_list`, and `camera_device_register()` is the sole `list_add()` site.

Still open (not blocking, already stated in the PR): draft status, and hardware qualification of the 550.4 MHz value / provider rounding / stream-restart clock readback.

_This comment was auto-generated by the **rs-agentic** bot._

ymodlin (changes requested):
This is change was still not approved in the design meeting.

PR #663, opened 2026-09-17
[RSDEV-13850] Add XU control for GMSL IMU pair-batching negotiation

pinquanw:
Pair-batching negotiation currently uses an experimental HWMC command. For RSDEV-13850, expose a dedicated GMSL IMU control so the SDK can negotiate paired or single-sensor IMUB records through its XU API before STREAMON.

Add `D500_CAMERA_CID_IMU_BATCH` (`DS5_CAMERA_CID_BASE + 39`) on D58x IMU video nodes. SDK depth-XU selector `0x1b` maps to this control; the driver reads/writes the one-byte HKR register at `0x4590` without sending HWMC commands.

| Operation | Byte layout |
|---|---|
| SET | bits0/1 = requested accel/gyro; 0 disables, 1 accel, 2 gyro, 3 accel+gyro. All other bits must be zero. |
| GET | bits0/1 = configured mask; bit7 = supported; bit6 = pixel mode; bit5 = IMU source active. Bits2..4 reserved zero. |

Reject changes while the host or firmware reports IMU streaming, and reject enabling batches in tunnel mode. GET reports zero capability for unsupported legacy firmware so bulk V4L2 control reads can still retrieve unrelated controls. SET requires the capability signature; zero is never mistaken for support. Malformed supported responses and I2C errors still propagate. Firmware applies SET synchronously. Verify the accepted mask with exactly one GET after SET; an I2C ACK alone is not confirmation. Return `-EIO` for a rejected mask or loss of SUPPORTED after the write, and `-EBUSY` if the source became active. An absent capability at preflight still returns `-EOPNOTSUPP` before issuing SET. A failed SET after writing can leave a changed mask: readback I2C failure or ACTIVE does not prove that the write was rejected. Re-read status and perform cleanup when idle; never assume automatic rollback or overwrite an active format. The SDK handles asynchronous source-release waiting during close, before attempting SET=0. Register as volatile/execute-on-write, keeping negotiation explicit rather than applying it at probe.

This is a GMSL control-plane change. The RAW8 256x1 capture format, VI receiver, clocks, priorities and frame payload transport are unchanged. USB HID does not advertise this control.

Companion changes:
- SDK: https://github.com/realsenseai/librealsense/pull/15680 (XU mapping/negotiation and existing IMUB decoder).
- HKR: https://github.com/RealSenseAI-Internal/soc-rs-soc-io-drivers/pull/623 (register endpoint; retires experimental IMUB HWMC control). Existing Apps #339, Device Manager #531 and prebuilt #95 remain dependencies of the firmware feature.

Validation:
- Cross-compiled the updated `d4xx.o` for ARM64 on the configured JP6.2.1 and JP7.1 build trees with `-Werror`.
- `python3 test/standalone_tests/test_imu_batch_xu.py`: production register helpers tested with mocked I2C under ASan/UBSan. Covers old FW, reserved bits, all masks, active host/FW rejection, tunnel behavior, I2C failures, immediate accepted/rejected readback, post-write ACTIVE, capability loss and a latched mask despite readback failure. The production helpers compile with `-Werror`; reintroducing an undeclared sleep primitive fails compilation. The standalone test is import-safe for pytest collection.
- Companion SDK Release/V4L2 library build and production negotiation/rollback plus packet-decoder tests passed.
- `git diff --check` passed. Updated the existing CLAUDE.md synchronous-control invariant.

Not deployed or validated on target hardware yet. This PR establishes the control path, not new FPS measurements.

### RSDEV-13850 integration qualification — 2026-09-18

The complete current integration has **not** been hardware-qualified. Historical 180-second results from Apps `aac7cacf` / IO `c6843ad9` and manual VI 550.4 MHz do not establish a pass for the current XU/merged revisions. Individual tests above retain their stated scope; this qualification requirement concerns the combined D585 milestone, not whether this PR is independently usable.

Firmware serialization/shaping needs Apps #339 + IO #623 + Device Manager #531 + HIF FW #3 + matching prebuilt #95. Paired delivery additionally needs Host #663 **and SDK #15680**. JP6.2 automatic clock mitigation is Host #662; the Viewer scenario additionally needs SDK #15692. Installing only the seven firmware/host PRs does not enable or decode pairs.

[IO #623](https://github.com/RealSenseAI-Internal/soc-rs-soc-io-drivers/pull/623) records the exact nine-PR candidate commits, deployment identity requirements, XU/start/stop/restart checks and acceptance contract. For four 960p60 streams: 180-second run averages of >=198 Hz per sensor when paired, >=100 Hz when legacy, with Depth/IR retaining 60 FPS and the previously accepted RGB-specific tolerance of >=59.35 FPS. Report all 5-second windows and source-counter gaps separately. The 5 ms check is at HKR pair formation; it is not a proven SDK delivery-latency bound or per-window rate guarantee. No new integrated FPS result is claimed by this update.


kernel/realsense/d4xx.c
```diff
@@ -3852,6 +3862,66 @@ static int d500_set_ae_policy(struct ds5 *state, u32 policy)
...
+	}
+	if (mask && !(status & D500_IMU_BATCH_PIXEL)) {
+		ret = -EOPNOTSUPP;
+		goto out;
+	}
+	ret = ds5_raw_write(state, D500_IMU_BATCH_XU_BASE, &value, sizeof(value));
+	if (ret)
+		goto out;
+	/* An I2C ACK alone does not confirm that FW accepted the new mask. */
+	for (tries = 0; tries < 20; ++tries) {
```
ymodlin: 🤖 auto-generated by the rs-agentic bot

`ds5_raw_read()` retries 5× with 5 ms delays, so on a flaky bus each of the 20 readbacks can cost ~20 ms → up to ~0.4 s holding `ds5_dev->lock`, which STREAMON/STREAMOFF and the link-flush arming also take (plus retry log spam).

CLAUDE.md: polling loops should use a single-shot read (`ds5_read_poll()` style), not the retrying helper. Also suggest named constants for the `20` tries / total timeout.

kernel/realsense/d4xx.c
```diff
@@ -3852,6 +3862,66 @@ static int d500_set_ae_policy(struct ds5 *state, u32 policy)
...
+		ret = d500_get_imu_batch(state, &status);
+		if (ret)
+			goto out;
+		if (status & D500_IMU_BATCH_ACTIVE) {
```
ymodlin: 🤖 auto-generated by the rs-agentic bot

Possible bug: this `-EBUSY` (and the `-ETIMEDOUT` exit below) happens *after* the register write, so SET is reported as failed while the new mask may already be latched in FW — the opposite of the "reject changes while streaming" contract in the PR description. Either roll back to the mask read before the write, or document that a failed SET can leave a changed configuration.

kernel/realsense/d4xx.c
```diff
@@ -5383,6 +5458,16 @@ static int ds5_g_volatile_ctrl(struct v4l2_ctrl *ctrl)
 				*ctrl->p_new.p_s32 = policy;
 		}
 		break;
+	case D500_CAMERA_CID_IMU_BATCH:
```
ymodlin, line 5464: 🤖 auto-generated by the rs-agentic bot

On legacy FW `d500_get_imu_batch()` returns `-EOPNOTSUPP`, and `VIDIOC_G_EXT_CTRLS` fails the whole batch at `error_idx` — so a single ioctl reading other IMU-node controls (fw_version, gyro sensitivity) fails too, and `v4l2-ctl --all` on that node errors out.

Precedent in this file: hide the control instead — `ds5_v4l2_ctrl_set_disabled()` as `d500_refresh_device_mode()` does for the 2C-only AE policy — or report status 0 here.

kernel/realsense/d4xx.c
```diff
@@ -3852,6 +3862,66 @@ static int d500_set_ae_policy(struct ds5 *state, u32 policy)
...
+{
+	int ret = ds5_raw_read(state, D500_IMU_BATCH_XU_BASE,
+			      status, sizeof(*status));
+
+	if (ret)
+		return ret;
+	/* An old FW may return zero for an unknown window. Never infer support. */
+	if (!(*status & D500_IMU_BATCH_SUPPORTED))
+		return -EOPNOTSUPP;
+	if (*status & 0x1c)
```
ymodlin: 🤖 auto-generated by the rs-agentic bot

Nit: name the reserved-bit mask (e.g. `D500_IMU_BATCH_RESERVED GENMASK(4, 2)`) instead of the bare `0x1c`. The new `#define`s (L3332-3336) also use single spaces where the neighbouring `*_XU_BASE` defines are tab-aligned.

ymodlin (commented):
🤖 Auto-generated review by the **rs-agentic bot** (no approval implied).

Control shape follows the CLAUDE.md XU convention (one CID, `VOLATILE | EXECUTE_ON_WRITE`, no probe-time apply); CID `+39`, register `0x4590` and the 1-byte read (short read → safe on legacy FW) all check out. 4 inline notes: 1 likely bug, 2 robustness, 1 nit.

Follow-ups — model/effort per CLAUDE.md:
- Rework the readback loop / lock hold — Opus, effort high (touches `ds5_dev->lock` on the streaming path).
- Hide-on-unsupported + naming nits — Sonnet, effort low.

kernel/realsense/d4xx.c
ghu3: HKR applies the SET synchronously (atomic CAS in `imu_gmsl_batch_configure()`), so the value is visible on the very next GET. A single readback is sufficient; the 20×1–2 ms poll loop adds up to 40 ms of latency to every failed SET (e.g. when the source is ACTIVE) without benefit.

ghu3 (commented):
Review for RSDEV-13850 (cross-PR set: apps#339, io#623, device-mgr#531, prebuilts#95, HIF FW#3, host driver #662/#663, SDK #15680/#15692).

XU selector 0x1B → I2C 0x4590, CID `BASE+39`, status bits (0x80/0x40/0x20, reserved 0x1c) are consistent with HKR io#623 and SDK#15680. `ds5_raw_read/write` do not take `ds5_dev->lock`, so holding it in `d500_set_imu_batch()` is safe. One nit inline. No blockers.

**Validation performed by reviewer:** the PR set's host unit tests (GmslImuBatchTest, test_gmsl_batch_xu.py, test_traffic_shaping/test_imu_pacing/test_imu_video_batch, test_hif_poll/test_hif_tx/test_hif_tx_publish) were built and run on host with `-Wall -Wextra -Werror -fsanitize=address,undefined` — all pass, no warnings. No full HKR image build, hardware run, or D555e/S585 USB regression was performed; findings are static unless stated.

kernel/realsense/d4xx.c
ymodlin: 🤖 auto-generated by the rs-agentic bot

change approved

ymodlin: 🤖 auto-generated by the rs-agentic bot

change approved — single GET, and the lock is now held for 3 I2C transactions instead of up to ~0.4 s.

kernel/realsense/d4xx.c
ymodlin: 🤖 auto-generated by the rs-agentic bot

Partly addressed: `-ETIMEDOUT` is gone. Still open — the `-EBUSY`/`-EIO` exits at L3910-3912 are after the write, so a failed SET can still leave a changed mask latched, and a post-write I2C error returns `-EIO` (indistinguishable from FW rejection) after FW applied it. The PR body documents the error codes but not that a failed SET may have changed the configuration; please add that to the CLAUDE.md bullet too, since that is the contract the SDK has to code against.

kernel/realsense/d4xx.c
ymodlin: 🤖 auto-generated by the rs-agentic bot

Still open in `f46cc20`. `d500_get_imu_batch()` returning `-EOPNOTSUPP` on legacy FW still propagates out of `ds5_g_volatile_ctrl` (L5460), so `VIDIOC_G_EXT_CTRLS` fails the whole batch and `v4l2-ctl --all` on the IMU node errors. Hiding the control via `ds5_v4l2_ctrl_set_disabled()` (as `d500_refresh_device_mode()` does) or reporting status 0 keeps unrelated IMU controls readable.

kernel/realsense/d4xx.c
ymodlin: 🤖 auto-generated by the rs-agentic bot

Still open in `f46cc20`: bare `0x1c` at L3875, and L3332-3336 use single spaces while the neighbouring `*_XU_BASE` defines at L3325-3327 are tab-aligned.

test/standalone_tests/test_imu_batch_xu.py
```diff
@@ -0,0 +1,83 @@
...
+        assert(d500_set_imu_batch(&state,3)==-EBUSY && !writes && !shared.lock);
+        reset(0xe3);assert(d500_set_imu_batch(&state,0)==-EBUSY && !writes);
+        reset(0x80);assert(d500_set_imu_batch(&state,3)==-EOPNOTSUPP && !writes);
+        reset(0x80);assert(d500_set_imu_batch(&state,0)==0);
+        reset(0);assert(d500_set_imu_batch(&state,3)==-EOPNOTSUPP && !writes);
+        reset(0xc0);fail_write=1;assert(d500_set_imu_batch(&state,3)==-EIO);
+        reset(0xc0);fail_read=2;assert(d500_set_imu_batch(&state,3)==-EIO);
+        reset(0xc0);delay_reads=3;assert(d500_set_imu_batch(&state,3)==-EIO && reads==2);
+        reset(0xc0);delay_reads=-1;
+        assert(d500_set_imu_batch(&state,3)==-EIO && sleeps==0 && reads==2 && !shared.lock);
```
ymodlin: 🤖 auto-generated by the rs-agentic bot

Test gap on the branch this commit rewrote: the post-write `ACTIVE` → `-EBUSY` path (d4xx.c L3910) is never exercised. `reset(0xe3)` only covers the *pre*-write check, and the `delay_reads` mock can flip the mask but never sets bit5, so the mock can't reach it. Suggest a `pending_active` hook that ORs `D500_IMU_BATCH_ACTIVE` into `wire` on the readback.

Also `sleeps` is now never incremented (no `usleep_range` stub remains), so `sleeps==0` here asserts nothing — either drop the counter or note that the real guard is the `-Werror` implicit-declaration failure if `usleep_range()` comes back.

CLAUDE.md
```diff
@@ -170,6 +170,9 @@ Each MIPI segment's lane count comes from a **different** DT property, so the ca
 
 ## Concurrency notes
 
+- The GMSL IMU batching register at `0x4590` applies SET synchronously: verify it with one GET under the device lock. Handle the asynchronous source-release wait in SDK close cleanup before SET, rather than polling for a rejected mask in the kernel.
```
ymodlin: 🤖 auto-generated by the rs-agentic bot

Nit: two blank lines after this bullet break the list run — every other bullet in the section is contiguous. Also this reads as a V4L2 control/register contract rather than a locking rule; "V4L2 control conventions" (L137) is the closer home.

ymodlin (commented):
🤖 auto-generated by the rs-agentic bot — review of `f46cc20` (synchronize).

The single-GET rewrite is correct: `ds5_dev->lock` is held across 3 I2C transactions instead of up to ~0.4 s, and nothing can change `0x4590` between the write and the readback (both `imu_streaming` and the register are serialized by that same lock). No new driver bug found in this commit.

Three earlier threads are still open — replied inline. Two notes on the new commit below.

kernel/realsense/d4xx.c
pinquanw: ✅ Fixed in f46cc202e: SET now performs one GET readback, returning `-EIO` on mask mismatch or `-EBUSY` for ACTIVE. Removed the 20× sleep/poll loop. ASan/UBSan production-helper tests verify immediate acceptance/rejection with no retry sleeps; ARM64 `d4xx.o` compilation passed. SDK close separately waits for asynchronous source release before issuing SET=0.

🤖 Generated by AI

pinquanw: Addressed by the single-readback implementation retained in 6272fa3c9. The retry loop and sleeps are gone; there is one preflight GET, SET, and verification GET under the device lock. Individual register transactions retain their existing I2C error handling. Updated tests reject any undeclared sleep primitive at compilation rather than relying on a never-incremented counter.

kernel/realsense/d4xx.c
pinquanw: Addressed in 6272fa3c9 by documenting the actual contract in both the PR description and V4L2 control conventions: after the write, failed SET does not imply unchanged configuration. Readback failure or ACTIVE can follow a latched mask. Tests now exercise a latched mask with readback failure and post-write ACTIVE. No rollback is attempted while the source could own the format; SDK #15680 records the endpoint before SET and performs bounded idle cleanup on failure.

kernel/realsense/d4xx.c
pinquanw: Fixed in 6272fa3c9. GET normalizes an unsupported legacy response to status 0 and success, preserving unrelated controls in G_EXT_CTRLS. SET still requires SUPPORTED and rejects unsupported firmware; I2C errors and malformed supported responses still propagate. Tests cover zero/unsupported responses, unsupported SET and capability loss after writing. ARM64 JP6.2.1/JP7.1 object compilation passed.

kernel/realsense/d4xx.c
pinquanw: Fixed in 6272fa3c9: D500_IMU_BATCH_RESERVED names GENMASK(4, 2), and the new register/status constants use tab-aligned definitions.

test/standalone_tests/test_imu_batch_xu.py
pinquanw: Fixed in 6272fa3c9. Added pending_active to set ACTIVE on the post-write readback and assert -EBUSY with one write, two reads, the latched mask, and an unlocked mutex. Also tests readback failure after the mask latched and capability loss. Removed the meaningless sleeps counter; the extracted production helpers build with -Werror and no sleep declaration. ASan/UBSan tests pass.

CLAUDE.md
pinquanw: Fixed in 6272fa3c9. Moved the register/control contract into V4L2 control conventions and removed the extra blank lines. The rule now also documents zero-capability legacy GET and the potentially changed configuration after a failed post-write verification.

kernel/realsense/d4xx.c
ymodlin: 🤖 auto-generated by the rs-agentic bot

change approved

kernel/realsense/d4xx.c
ymodlin: 🤖 auto-generated by the rs-agentic bot

change approved — the CLAUDE.md bullet now states a failed SET can leave a changed mask.

kernel/realsense/d4xx.c
ymodlin: 🤖 auto-generated by the rs-agentic bot

change approved — legacy FW now normalizes to status 0 / success, so bulk `G_EXT_CTRLS` on the IMU node survives.

kernel/realsense/d4xx.c
ymodlin: 🤖 auto-generated by the rs-agentic bot

change approved. Tiny leftover: `D500_IMU_BATCH_SUPPORTED` is 24 chars, so its two tabs land the value at col 48 while the other five sit at col 40 — one tab is enough there.

test/standalone_tests/test_imu_batch_xu.py
ymodlin: 🤖 auto-generated by the rs-agentic bot

change approved — `pending_active` reaches the post-write `-EBUSY` path and `sleeps` is gone.

CLAUDE.md
ymodlin: 🤖 auto-generated by the rs-agentic bot

change approved

kernel/realsense/d4xx.c
```diff
@@ -3852,6 +3863,68 @@ static int d500_set_ae_policy(struct ds5 *state, u32 policy)
...
+		goto out;
+	}
+	ret = ds5_raw_write(state, D500_IMU_BATCH_XU_BASE, &value, sizeof(value));
+	if (ret)
+		goto out;
+	/* FW applies SET synchronously; an ACK still requires one verified GET. */
+	ret = d500_get_imu_batch(state, &status);
+	if (ret)
+		goto out;
+	if (!(status & D500_IMU_BATCH_SUPPORTED))
```
ymodlin, line 3917: 🤖 auto-generated by the rs-agentic bot

New in `6272fa3`: this post-write branch returns `-EOPNOTSUPP`, but the write already landed. The SET contract in the PR body only documents `-EIO` (rejected mask) and `-EBUSY` (source active) as post-write outcomes, and `-EOPNOTSUPP` reads as "feature absent, nothing changed" — the SDK would reasonably skip the idle cleanup that CLAUDE.md now says callers must perform, leaving the mask latched. Suggest `-EIO` here (capability loss after a successful preflight is a FW fault, not an absent feature), or add `-EOPNOTSUPP` to the documented post-write set.

pinquanw: Agreed on the error semantics; fixed in 8c28d9bf2. Missing SUPPORTED at preflight still returns -EOPNOTSUPP before writing. Losing SUPPORTED after SET now returns -EIO because the mask may already be latched. Tests cover all four masks, one write/two reads, retained mask and mutex release; ASan/UBSan and ARM64 JP6.2.1/JP7.1 object builds passed. Updated CLAUDE.md and the PR contract. The current SDK already attempts idle cleanup after SET errors, so this fixes the interface ambiguity rather than claiming a reproduced SDK leak. Squashed the PR into one commit and force-pushed with a lease.

ymodlin: 🤖 auto-generated by the rs-agentic bot

change approved — post-write capability loss is now `-EIO`, and CLAUDE.md documents `-EOPNOTSUPP` as pre-write only.

kernel/realsense/d4xx.c
```diff
@@ -3852,6 +3863,68 @@ static int d500_set_ae_policy(struct ds5 *state, u32 policy)
...
+			      status, sizeof(*status));
+
+	if (ret)
+		return ret;
+	/* Zero capability keeps bulk control reads usable with legacy firmware. */
+	if (!(*status & D500_IMU_BATCH_SUPPORTED)) {
+		*status = 0;
+		return 0;
+	}
+	if (*status & D500_IMU_BATCH_RESERVED)
```
ymodlin: 🤖 auto-generated by the rs-agentic bot

Possible bug: an all-ones read is treated as SUPPORTED. A register the connected FW does not implement (or leaves uninitialised) commonly answers `0xFF`/`0xFFFF` — this file already documents that for `DS5_READOUT_SHAPING` (L5491). `0xFF` passes the bit7 gate, then trips `D500_IMU_BATCH_RESERVED` → `-EBADMSG`, which propagates out of `ds5_g_volatile_ctrl` (L5469) and fails the whole `VIDIOC_G_EXT_CTRLS` batch on the IMU node — the exact regression the zero-capability normalisation was added to prevent.

Bit7 alone is not a capability signature. Suggest normalising an all-ones byte to status 0 as well.

Same path on SET: preflight then returns `-EBADMSG` instead of `-EOPNOTSUPP`, and the post-write verification GET can also return `-EBADMSG`, which is outside the `-EIO`/`-EBUSY` post-write set CLAUDE.md now documents.

ymodlin (commented):
🤖 auto-generated by the rs-agentic bot

Re-review of `8c28d9b` (PR → `dev`). All eight earlier threads look addressed. One remaining possible bug in the GET capability check, plus a test gap and two nits.

test/standalone_tests/test_imu_batch_xu.py
```diff
@@ -0,0 +1,99 @@
...
+        pending_active=drop_capability=false;
+        writes=reads=delay_reads=fail_read=fail_write=0;
+    }
+    '''
+    tests = r'''
+    int main(void) {
+        u8 status;
+        reset(0);assert(d500_get_imu_batch(&state,&status)==0 && status==0);
+        reset(0x7f);assert(d500_get_imu_batch(&state,&status)==0 && status==0);
+        reset(0xc4);assert(d500_get_imu_batch(&state,&status)==-EBADMSG);
```
ymodlin, line 61: 🤖 auto-generated by the rs-agentic bot

Test gap for the above: no case for an all-ones (`0xFF`) answer from FW that doesn't implement `0x4590`. Add `reset(0xff)` for GET and SET and pin the intended behaviour.

kernel/realsense/d4xx.c
```diff
@@ -4880,6 +4953,11 @@ static int ds5_s_ctrl(struct v4l2_ctrl *ctrl)
 			    DS5_DEVICE_TYPE_D58X)
 			ret = d500_set_ae_policy(state, ctrl->val);
 		break;
+	case D500_CAMERA_CID_IMU_BATCH:
+		if (state->is_imu &&
+		    READ_ONCE(state->ds5_dev->cached_device_type) == DS5_DEVICE_TYPE_D58X)
```
ymodlin, line 4962: 🤖 auto-generated by the rs-agentic bot

Nit: every other D58x case in this switch wraps after `cached_device_type) ==` (see the `GYRO_SENSITIVITY` case right below). Same at L5466.

kernel/realsense/d4xx.c
ymodlin: 🤖 auto-generated by the rs-agentic bot

Still open at `8c28d9b`: `D500_IMU_BATCH_SUPPORTED` is 24 chars, so its two tabs put `BIT(7)` at col 48 while the other five land at col 40. One tab there.

kernel/realsense/d4xx.c
ymodlin: 🤖 auto-generated by the rs-agentic bot

change approved

kernel/realsense/d4xx.c
ymodlin: 🤖 auto-generated by the rs-agentic bot

Still open at `39e3cb1`. `0xff` passes the bit7 gate, then trips `D500_IMU_BATCH_RESERVED` → `-EBADMSG` out of `ds5_g_volatile_ctrl` (L5474), failing the whole `VIDIOC_G_EXT_CTRLS` batch on the IMU node. Same on the SET preflight (`-EBADMSG` instead of `-EOPNOTSUPP`) and on the post-write GET, which is outside the `-EIO`/`-EBUSY` post-write set CLAUDE.md documents. Normalising an all-ones byte to status 0 still looks like the fix.

test/standalone_tests/test_imu_batch_xu.py
ymodlin: 🤖 auto-generated by the rs-agentic bot

Still open at `39e3cb1`: `reset(0xc4)`/`reset(0xc3)` cover reserved and out-of-range bits, but there is still no `reset(0xff)` case for FW that does not implement `0x4590`.

kernel/realsense/d4xx.c
ymodlin: 🤖 auto-generated by the rs-agentic bot

Still open at `39e3cb1`: L4962 and L5470 keep `== DS5_DEVICE_TYPE_D58X` on the `READ_ONCE` line, while the `GYRO_SENSITIVITY` case directly below each one wraps after `cached_device_type) ==`.

kernel/realsense/d4xx.c
ymodlin: 🤖 auto-generated by the rs-agentic bot

Still open at `39e3cb1` (L3340): `D500_IMU_BATCH_SUPPORTED` is 24 chars, so its two tabs put `BIT(7)` at col 48 while the other seven land at col 40. One tab there.

kernel/realsense/d4xx.c
```diff
@@ -3852,6 +3866,69 @@ static int d500_set_ae_policy(struct ds5 *state, u32 policy)
...
+		goto out;
+	/* FW applies SET synchronously; an ACK still requires one verified GET. */
+	ret = d500_get_imu_batch(state, &status);
+	if (ret)
+		goto out;
+	if (!(status & D500_IMU_BATCH_SUPPORTED))
+		ret = -EIO;
+	else if (status & D500_IMU_BATCH_ACTIVE)
+		ret = -EBUSY;
+	else if ((status & D500_IMU_BATCH_MASK) != mode)
```
ymodlin, line 3925: 🤖 auto-generated by the rs-agentic bot

Possible bug, new in `39e3cb1`: this verifies by exact equality, but the new encoding makes **0 and 1 both mean "no batching"** (CLAUDE.md L139: `0/1 disable batching (default 1)`). If FW canonicalises a disable to its default encoding `1`, a `SET 0` reads back as `1` → `-EIO` for a write that did exactly what was asked. The PR body says the SDK issues `SET=0` on close, so every teardown would report failure and then run the idle cleanup CLAUDE.md requires after a failed SET.

The tests cannot catch this: the mock echoes the written byte verbatim (`wire=(wire & 0xfc)|pending`), so no case has FW answer with a different-but-equivalent mode.

Suggest comparing the effective state rather than the raw value (`((status & D500_IMU_BATCH_MASK) == D500_IMU_BATCH_PAIR) != (mode == D500_IMU_BATCH_PAIR)`), or normalising `mode` to one canonical disable encoding before the write.

ymodlin (commented):
🤖 auto-generated by the rs-agentic bot

Re-reviewed after the force-push to `39e3cb1`, which reworks the register semantics from an accel/gyro bitmask to a mode enum (`0/1` = no batching, `2` = pairs). That is a contract change, so re-reviewed the helpers rather than just the delta.

- 1 new possible BUG inline (post-write verification vs. the two disable encodings).
- The 4 comments open at `8c28d9b` are all still unaddressed at `39e3cb1`; replied in-thread rather than reopening them here.
- Doc mismatch: the **PR description** still documents the old bitmask contract (`bits0/1 = requested accel/gyro; ... 3 accel+gyro`, `GET bits0/1 = configured mask`), which now contradicts the commit message and CLAUDE.md L139 (`0/1 disable batching (default 1)`, `2` pairs, `3..255` invalid). Since the SDK codes against this, please update the body.

ymodlin (changes requested):
IMU-Batching is still not approved during the design review.

PR #664, opened 2026-09-17
[D585][GMSL] Expose multiplex profiles and direct-XU ODPD control

polejoe:
Depends-On: https://github.com/RealSenseAI-Internal/soc-rs-soc-io-drivers/pull/645
Depends-On: https://github.com/RealSenseAI-Internal/soc-rs-soc-device-mgr/pull/564
Depends-On: https://github.com/RealSenseAI-Internal/soc-rs-hkr-apps/pull/368
Depends-On: https://github.com/RealSenseAI-Internal/soc-rs-hkr-test/pull/450

## Purpose and current implementation

- Preserve camera-described RSVL capture and export the validated logical RSDS snapshot through MUX_CAPS (0x009A4034).
- Expose the shared 64-byte read/write MUX_CONTROL (0x009A4035), mapped to direct-XU window 0x4590.
- Use IR's physical VC2 capture resources without IR-bank activation. The standard sensor s_stream hook drives capture start/stop.
- Validate explicit member/profile and configured receive geometry; invalidate cached configuration across camera reset.
- The existing mux_control_lock serializes transactions; it does not provide per-file-descriptor queue ownership. Runtime member SET is rejected while streaming.
- No Host code or NVIDIA VI patch is changed by this review follow-up; this body update clarifies the current boundary.

Current head: `381e338`. Branch: `zhous/GMSL_PD`. Base: `dev`.

## Current follow-up validation (2026-09-20)

- D5x freertos incremental build: PASS, with no compiler/linker errors.
- IO host-unit suite: PASS, including 10 camera descriptor checks and legacy-table compatibility.
- 32 Python catalog/control/preview tests: PASS.
- Camera-produced 410-byte RSDS blobs cross-parsed by the App parser: PASS; Pixel/Tunnel logical catalogs match; only ODPD is transport-available.
- Extracted production prepareOdPayload method tested with ASan/UBSan: fixed 3072-byte extent, full padding, bounds rejection, AutoSrc preservation and non-OD bypass PASS.
- These are local validation results. The fixed-carrier follow-up has NOT been redeployed or hardware-requalified.

## Earlier hardware evidence, not qualification of the follow-up

The preceding combined stack was tested on 2026-09-20 with D5x 3C, Jetson67, MAX96712 Pixel and JP6.2.2:
- Real OD 222/222 and RGB 447/447; two repeats each OD 117/117 and RGB 237/237; valid video/metadata pairing.
- Negative member/profile controls rejected without camera-state mutation.
- Ordinary-stream smoke: Depth 228/228 + RGB 227/227; separately Depth 211/211 + IR 212/212 + IMU 1487/1487.
- No HIF timeout/panic or NVCSI/VI capture error in the OD captures. Calibration/NVM notices, Depth APM/Laser errors and recovered I2C retries prevent an all-system zero-error claim.

## Review disposition and extension boundary

- STANDARD: V4L2 format/queue/STREAMON/OFF and extended-control ioctl framework.
- REALSENSE EXTENSIONS: RSVL FourCC, logical RSDS TOCs, custom MUX CIDs and direct-XU payload. None is presented as an upstream standard perception ABI.
- Existing USB parameter mappings remain separate from member lifecycle; this control does not implement a generic SET_PARAMETER writer or automatic SDK XU mapping.
- SDK resolver/shared-node integration, authoritative VI allocation admission, per-fd ownership, drain/recovery and multi-member hot-add remain open. The single-ODPD implementation does not close those gates.
- Occupancy is the next runtime feature after ODPD; PCL follows Occupancy. LPCL requires its own registered metadata identity before publication.
- Current OD request/catalog rate 30 is not a promise of 30 detection results per second: prior RGB30 produced OD15 with entry cadence 2.
- Starting OD rebuilds the shared RGB graph; seamless hot-add is not implemented. GUI overlay, 2C/new-control coverage, USB/D555 regression, distance with Depth and fresh Tunnel validation remain open.
- Retain draft status. New runtime types or larger receive envelopes must not be enabled solely from descriptor presence.


kernel/realsense/d4xx.c
```diff
@@ -1187,6 +1189,7 @@ static const struct {
 	/* D401 CSI passthrough: 10bit data riding an opeque 8-bit carrier. */
 	{ RS_PIXFMT_SBGGR10P,   MEDIA_BUS_FMT_RS_SBGGR10P_1X8, GMSL_CSI_DT_RAW_8 },
 	{ RS_PIXFMT_IMU,        MEDIA_BUS_FMT_Y8_1X8 },
+	{ RS_PIXFMT_RSVL,       MEDIA_BUS_FMT_RS_VARLEN_1X8 },
```
ymodlin, line 1192: 🤖 _auto-generated by the rs-agentic bot_

**BUG — breaks the JP5 builds.** `MEDIA_BUS_FMT_RS_VARLEN_1X8` (0x500a) is defined only by the jammy/noble patches (`kernel/kernel-jammy-src/*/0006-realsense-variable-length-data-format.patch`, `kernel-noble-src/7.0/0007`). The JP5 carriers (`kernel/kernel-5.10/{5.0.2,5.1.2}/000*-Add-Realsense-NV12-User-Defined-media-bus-format.patch`) define only 0x5008/0x5009, and there is no `#ifndef` fallback in `d4xx.c`.

`apply_patches.sh:92` symlinks this single `d4xx.c` into **every** JetPack tree, and CI runs `build-jp502` / `build-jp512` / `build-jp5.1.6` — those will fail to compile. Validation in the PR body only covers 6.2.2.

Fix: add the define to the two JP5 patches, or guard the table entry.

ymodlin, line 1192: 🤖 _auto-generated by the rs-agentic bot_

**Possible BUG — `wire_dt` left at 0.** Per the table's own contract, `wire_dt == 0` means "no override, the wire DT is whatever FW declares" (`fmt->data_type = rec.src_data_type`). But the VI row for `RS_VARLEN_1X8` matches **RAW8**, and in pixel mode the MAX96717 pipe silently drops user-defined DTs.

The other opaque 8-bit carriers pin it explicitly (`RS_PIXFMT_NV12_FLAT`, `RS_PIXFMT_SBGGR10P` → `GMSL_CSI_DT_RAW_8`). Unless the descriptor is guaranteed to declare 0x2A for this pixfmt, pin `GMSL_CSI_DT_RAW_8` here too — otherwise a non-RAW8 descriptor yields a silent VI timeout rather than an error.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
```diff
@@ -980,35 +980,49 @@ index 8e237ee..c1e57e2 100644
...
-+	return err;
++	/*
++	 * The configure-once table provides the default Depth/RGB/IR/IMU data
++	 * types. A camera-provided profile can select another wire DT on any VC
++	 * (for example RAW8 Perception on VC2 instead of regular IR YUV422), so
++	 * refresh the active VC's primary mapping on every stream setup.
++	 */
++	src_vc = vc_id << 6;
++	dst_vc = (priv->vc_remap[link][vc_id] & 0x3) << 6;
++	map_pipe_control[0].addr = primary_src_addr[vc_id] + map_pipe_offs;
```
ymodlin, line 1016: 🤖 _auto-generated by the rs-agentic bot_

**Possible BUG — duplicate mapping on VC2.** VC2 is the only VC with a *second* payload map in the configure-once table: map 12 (`TX37/TX38`) is already `vc2 | GMSL_CSI_DT_RAW_16`.

Streaming the D58x IR RAW16 calibration format (`ds5_y_formats_d58x[2]`, `.data_type = GMSL_CSI_DT_RAW_16`) now writes `vc2 | RAW_16` into map 8 as well, so two enabled mappings match the same (VC, DT) and the deserializer forwards each packet twice. The old `vc_id == 1` gate made this unreachable, so this is a regression on an existing HW-validated path.

Suggest skipping the refresh when `data_type1` already matches a static map for that VC (or clearing/reprogramming map 12 together with map 8). Same code in the two JP5 carriers (`kernel/nvidia/5.1.2/0007`, `5.0.2/0017`).

ymodlin (commented):
Automated review by the **rs-agentic bot**. External contribution — scoped to potential bugs only. 3 inline comments; no approval implied.

kernel/realsense/d4xx.c
ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `cb2c054`: this push is **message-only** (tree identical to `b7fdea2`, only a `Signed-off-by` line added), so this is still open.

Still confirmed: `0x500a` is defined only in `kernel-jammy-src/{6.0,6.2,6.2.1,6.2.2}/0006` and `kernel-noble-src/7.0/0007`; `kernel/kernel-5.10/{5.0.2,5.1.2,5.1.6}` define only `0x5008`/`0x5009`, and `apply_patches.sh:92` still symlinks this `d4xx.c` into those trees. `build-jp502` / `build-jp512` / `build-jp5.1.6` will not compile.

ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `cb2c054`: message-only push (tree unchanged), so this is still open. The `RS_PIXFMT_RSVL` row still omits `wire_dt`, while the other opaque 8-bit carriers (`NV12_FLAT`, `SBGGR10P`) still pin `GMSL_CSI_DT_RAW_8`.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `cb2c054`: message-only push (tree unchanged), so this is still open. Map 12 is still statically `src_vc2 | GMSL_CSI_DT_RAW_16` (patch line ~931) while `__max96712_update_multi_vc_pipe()` now writes the profile DT into map 8 for `vc_id == 2`, so streaming `ds5_y_formats_d58x[2]` (`GMSL_CSI_DT_RAW_16`) still yields two enabled mappings on the same (VC, DT).

kernel/realsense/d4xx.c
ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `3a3bb3f6`: this push is a **rebase onto `dev`** (picks up `ace5639c` + `3e20c289`); the PR's own diff is byte-identical to `cb2c054`, so this is still open.

Now confirmed by CI on the previous head — `Build JP 5.1.2` / `5.0.2` / `5.1.6` all failed with:

```
sources_5.1.2/kernel/nvidia/drivers/media/i2c/d4xx.c:1192:26: error: 'MEDIA_BUS_FMT_RS_VARLEN_1X8' undeclared here (not in a function); did you mean 'MEDIA_BUS_FMT_RS_SBGGR10P_1X8'?
```

Fix: add `#define MEDIA_BUS_FMT_RS_VARLEN_1X8 0x500a` to `kernel/kernel-5.10/{5.0.2,5.1.2,5.1.6}/000*-Add-Realsense-NV12-User-Defined-media-bus-format.patch`, or guard the table entry.

ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `3a3bb3f6`: rebase-only push, PR diff unchanged — still open. `RS_PIXFMT_RSVL` still omits `wire_dt` while `NV12_FLAT` / `SBGGR10P` pin `GMSL_CSI_DT_RAW_8`.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `3a3bb3f6`: rebase-only push, PR diff unchanged — still open. Map 12 (`TX37`/`TX38`, patch lines 931-932) is still statically `src_vc2 | GMSL_CSI_DT_RAW_16` while `__max96712_update_multi_vc_pipe()` writes the profile DT into map 8 (`TX29`/`TX30`) for `vc_id == 2`.

kernel/realsense/d4xx.c
ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `e9f0f27`: **rebase-only** push (new base `a57b229c`, the `1.0.7.1` version bump); all four PR files are byte-identical to `3a3bb3f6`. Still open — `0x500a` is still absent from `kernel/kernel-5.10/{5.0.2,5.1.2,5.1.6}/000*-Add-Realsense-NV12-User-Defined-media-bus-format.patch`, so `build-jp502` / `build-jp512` / `build-jp5.1.6` will fail again.

ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `e9f0f27`: rebase-only push, PR diff byte-identical — still open. `RS_PIXFMT_RSVL` still omits `wire_dt` (→ `override_data_type = 0`, `d4xx.c:1510`) while `NV12_FLAT` / `SBGGR10P` pin `GMSL_CSI_DT_RAW_8`.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `e9f0f27`: rebase-only push, PR diff byte-identical — still open. Map 12 (`TX37`/`TX38`) is still statically `src_vc2 | GMSL_CSI_DT_RAW_16` while `__max96712_update_multi_vc_pipe()` writes the profile DT into map 8 (`TX29`/`TX30`) for `vc_id == 2`; `ds5_y_formats_d58x[2]` (`d4xx.c:2406`, `GMSL_CSI_DT_RAW_16`) still hits it.

kernel/realsense/d4xx.c
```diff
@@ -4421,6 +4559,185 @@ static int ds5_hw_reset_with_recovery(struct ds5 *state)
...
+		if (reply.transaction_id == request->transaction_id) {
+			if (reply.operation != request->operation ||
+			    reply.stream_id != request->stream_id)
+				return -EPROTO;
+			state->mux_status = reply;
+			if (reply.state != D500_MUX_PENDING) {
+				u8 expected = operation == D500_MUX_CAPTURE_START ? D500_MUX_ACTIVE :
+					operation == D500_MUX_ENABLE ? D500_MUX_ARMED : D500_MUX_CONFIGURED;
+
+				ret = (s32)le32_to_cpu(reply.result);
```
ymodlin, line 4632: 🤖 _auto-generated by the rs-agentic bot_

**BUG — device-supplied `result` returned as an errno.** `reply.result` is FW data; a non-zero *positive* status is returned verbatim up through `d500_mux_set_control()` → `ds5_s_ctrl()` → `VIDIOC_S_EXT_CTRLS`. The v4l2 ioctl layer only diverts on `err < 0`, so a positive FW failure code reaches userspace as the ioctl return value and `ioctl() != -1` reads as **success** — a rejected CONFIGURE/ENABLE looks like it worked.

Map it before returning, e.g. `if (ret > 0) ret = -EREMOTEIO;` (or a FW-status → errno table).

kernel/realsense/d4xx.c
```diff
@@ -7450,6 +7817,100 @@ static void ds5_flush_idle_link(struct ds5 *state)
...
+#endif
+		/* Keep teardown possible after an uncertain device-side start. */
+		sensor->streaming = true;
+		ret = d500_mux_exchange(state, &request, D500_MUX_CAPTURE_START);
+		if (ret) {
+			/* STREAMON failure need not be followed by STREAMOFF. Stop our request. */
+			stop_ret = d500_mux_exchange(state, &request, D500_MUX_CAPTURE_STOP);
+			if (stop_ret) {
+				dev_err(&state->client->dev,
+					"MUX start failed %d; cleanup unresolved %d\n", ret, stop_ret);
```
ymodlin, line 7943: 🤖 _auto-generated by the rs-agentic bot_

**BUG — failed start latches `streaming = true`.** When the compensating `CAPTURE_STOP` also fails, this `goto out` leaves `sensor->streaming = true` and `ds5_dev->ir_streaming = true` (pipe never released). STREAMON has already failed, so V4L2 will not call STREAMOFF; the next STREAMON then hits the `sensor->streaming == on` short-circuit above and **returns 0 without starting the camera** — the app blocks on DQBUF forever.

Reporting the unresolved cleanup is right, but the node must not look "already streaming" afterwards: clear `sensor->streaming`/`ir_streaming`, or latch an explicit fault that the next start rejects, so a retry cannot silently succeed.

kernel/realsense/d4xx.c
```diff
@@ -4421,6 +4559,185 @@ static int ds5_hw_reset_with_recovery(struct ds5 *state)
...
+	if (state->ir.sensor.streaming) {
+		ret = -EBUSY;
+		goto out;
+	}
+	ret = d500_mux_validate_profile(state, &request);
+	if (ret)
+		goto out;
+	if (operation != D500_MUX_CONFIGURE &&
+	    (!state->mux_configured ||
+	     memcmp(&request.payload_version, &state->mux_profile.payload_version, 20))) {
```
ymodlin: 🤖 _auto-generated by the rs-agentic bot_

**Enhancement — bare `20` couples this compare to the struct layout.** It currently spans `payload_version`..`max_payload_bytes` (offsets 16-35), but any inserted field silently changes what ENABLE/DISABLE verifies. Prefer an explicit bound, e.g. `offsetofend(struct d500_mux_control, max_payload_bytes) - offsetof(struct d500_mux_control, payload_version)`, or group those fields in a named sub-struct and `memcmp` that.

It also excludes `stream_id` — harmless while `d500_mux_validate_profile()` pins `stream_id == 5`, but that hardcoded id deserves a named constant too.

ymodlin (commented):
🤖 _auto-generated by the rs-agentic bot_

Reviewed the new commit `381e338` (*expose multiplex capabilities and direct XU lifecycle*, +472 in `d4xx.c`). Note the PR body still advertises `e9f0f27` as the current head.

The descriptor-side work holds up: the new logical TOCs are bounds-checked by the existing generic `off + n * stride > total` gate, `total_size` is capped at `RS_DESC_MAX_SIZE` so the `MUX_CAPS` copy into the 2048-byte control cannot overflow, the published-descriptor lifetime makes the post-unlock `blob` use safe, and the two-byte version probe before the 64-byte XU read follows the short-read rule in `CLAUDE.md`. `CLAUDE.md` was updated per the post-patch hygiene rule.

Two new bugs below in the XU lifecycle, plus one maintainability note. **The three earlier findings are still open** — this push touches only `CLAUDE.md` and `d4xx.c`, so the JP5 build break in particular is unchanged.

kernel/realsense/d4xx.c
ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `381e338`: this push changes only `CLAUDE.md` and `kernel/realsense/d4xx.c`, so this is **still open**. `grep -r MEDIA_BUS_FMT_RS_VARLEN_1X8 kernel/kernel-5.10/` returns nothing — `0x500a` is still undefined in `5.0.2/0005`, `5.1.2/0003` and `5.1.6/0003`, while `d4xx.c:1251` still references it. `build-jp502` / `build-jp512` / `build-jp5.1.6` will fail again (CI has not run yet on this head).

ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `381e338`: `d4xx.c` changed, but not this table — **still open**. `d4xx.c:1251` is still `{ RS_PIXFMT_RSVL, MEDIA_BUS_FMT_RS_VARLEN_1X8 }` with no `wire_dt`, while `SBGGR10P` on line 1249 pins `GMSL_CSI_DT_RAW_8`.

Worth noting this head raises the stakes: `ds5_configure()` now returns early for the MUX path (`d4xx.c:3184`), so the override-register write that would have carried the wire DT is skipped entirely on this stream.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `381e338`: this push does not touch `nvidia-oot/`, so this is **still open**. Map 12 (`TX37`/`TX38`) remains statically `src_vc2 | GMSL_CSI_DT_RAW_16` while `__max96712_update_multi_vc_pipe()` writes the profile DT into map 8 for `vc_id == 2`; `ds5_y_formats_d58x[2]` (`GMSL_CSI_DT_RAW_16`) still collides.

kernel/realsense/d4xx.c
```diff
@@ -7450,6 +7870,116 @@ static void ds5_flush_idle_link(struct ds5 *state)
...
+	if (on) {
+		if (!state->mux_configured || !state->mux_armed) {
+			ret = -EINVAL;
+			goto out;
+		}
+		if ((state->mux_selected_mask & state->mux_configured_mask) !=
+		    state->mux_selected_mask) {
+			ret = -EINVAL;
+			goto out;
+		}
```
ymodlin, line 7920: 🤖 _auto-generated by the rs-agentic bot_

**BUG (new in `cd58dee`) — STREAMON no longer re-validates the cached profile.** The `d500_mux_validate_profile()` call on the start path was replaced by a mask-subset check, so the carrier-geometry compare is gone.

`__ds5_sensor_set_fmt()` (`d4xx.c:2919`) takes `state->lock`, not `mux_control_lock`, and freely rewrites `ir.sensor.config.resolution` while not streaming. CONFIGURE → S_FMT to another RSVL resolution → STREAMON now passes, and the camera emits payloads sized for the old carrier into the new VI geometry — a silent short/long-line capture, not an error.

The PR body still claims "validate … configured receive geometry". Keep the `d500_mux_validate_profile()` call on the `on` path, or re-compare `carrier_width`/`carrier_height` against `ir.sensor.config.resolution` here.

kernel/realsense/d4xx.c
```diff
@@ -4421,6 +4571,226 @@ static int ds5_hw_reset_with_recovery(struct ds5 *state)
...
+		if (operation == D500_MUX_CONFIGURE) {
+			state->mux_profile[profile_index] = request;
+			state->mux_configured_mask |=
+				d500_mux_member_bit(request.stream_id);
+			state->mux_configured = true;
+			state->mux_reset_ref = atomic_read(ds5_get_reset_gen(state));
+		}
+		if (operation == D500_MUX_ENABLE || operation == D500_MUX_DISABLE)
+			state->mux_selected_mask = member_mask;
+		state->mux_armed = state->mux_selected_mask != 0;
```
ymodlin, line 4787: 🤖 _auto-generated by the rs-agentic bot_

**BUG (new in `cd58dee`) — `mux_armed`/`mux_selected_mask` can diverge from the device.**

1. **CONFIGURE no longer disarms.** Old code was `mux_armed = operation == D500_MUX_ENABLE`; it is now recomputed from `mux_selected_mask`, which CONFIGURE never touches. `ENABLE(ODPD)` → `CONFIGURE(ODPD)` leaves `mux_armed == true` while the device replied `D500_MUX_CONFIGURED` (per the `expected` ternary in `d500_mux_exchange`). The next STREAMON then issues CAPTURE_START against a member the FW has not armed.
2. **DISABLE assigns `member_mask` verbatim.** `DISABLE(stream_id=ODPD, member_mask=OCCUPANCY)` passes the `mux_configured_mask` subset check and *arms* occupancy driver-side with no ENABLE.

Related ABI trap: `member_mask` means "target member" for CONFIGURE/ENABLE but "residual armed set" for DISABLE, so the intuitive `DISABLE(member_mask=ODPD)` keeps ODPD armed. Worth spelling out next to `D500_MUX_DISABLE`.

Suggest clearing the reconfigured member's bit on CONFIGURE, and computing DISABLE's new mask from the previous one (`selected &= ~bit`) rather than trusting the request.

kernel/realsense/d4xx.c
```diff
@@ -691,7 +717,52 @@ enum {
...
+#define D500_MUX_ENABLE 2
+#define D500_MUX_DISABLE 3
+#define D500_MUX_CAPTURE_START 4
+#define D500_MUX_CAPTURE_STOP 5
+#define D500_MUX_CONFIGURED 1
+#define D500_MUX_ARMED 2
+#define D500_MUX_PENDING 3
+#define D500_MUX_ACTIVE 4
+#define D500_MUX_FAULT 5
+#define D500_MUX_STREAM_ODPD 5
```
ymodlin, line 748: 🤖 _auto-generated by the rs-agentic bot_

Enhancement — `D500_MUX_STREAM_ODPD` is `5`, the same value as `D500_MUX_FAULT` one line above, in a single undifferentiated `#define` block that also holds operations (`D500_MUX_CAPTURE_STOP 5`). Nothing stops a `state == D500_MUX_STREAM_ODPD` typo from compiling. Separate the three groups with `D500_MUX_OP_*` / `D500_MUX_STATE_*` / `D500_MUX_STREAM_*` prefixes, or at least a blank line each.

kernel/realsense/d4xx.c
```diff
@@ -4421,6 +4571,226 @@ static int ds5_hw_reset_with_recovery(struct ds5 *state)
...
+		     s.stream_id != D500_MUX_STREAM_OCCUPANCY) ||
+		    s.carrier_stream_id != DS5_STREAM_IR)
+			return -EOPNOTSUPP;
+		for (j = s.fmt_first; j < s.fmt_first + s.fmt_count; j++) {
+			struct d500_desc_logical_format f;
+
+			ds5_desc_rec(blob, tabs[RS_DESC_T_LOGICAL_FORMAT], j, &f, sizeof(f));
+			if (f.payload_magic !=
+			    (s.stream_id == D500_MUX_STREAM_ODPD ?
+			     0x5445444f : 0x3150414d) ||
```
ymodlin, line 4704: 🤖 _auto-generated by the rs-agentic bot_

Enhancement — bare payload magics inside a ternary inside a comparison. `0x5445444f` is `'ODET'`, `0x3150414d` is `'MAP1'`; neither is readable here and a third member makes it worse. Define them next to the stream ids and pick via a small `stream_id → magic` helper.

kernel/realsense/d4xx.c
```diff
@@ -691,7 +717,52 @@ enum {
...
+#define D500_MUX_ARMED 2
+#define D500_MUX_PENDING 3
+#define D500_MUX_ACTIVE 4
+#define D500_MUX_FAULT 5
+#define D500_MUX_STREAM_ODPD 5
+#define D500_MUX_STREAM_OCCUPANCY 7
+#define D500_MUX_MEMBER_ODPD BIT(D500_MUX_STREAM_ODPD)
+#define D500_MUX_MEMBER_OCCUPANCY BIT(D500_MUX_STREAM_OCCUPANCY)
+#define D500_MUX_SUPPORTED_MEMBERS \
+	(D500_MUX_MEMBER_ODPD | D500_MUX_MEMBER_OCCUPANCY)
```
ymodlin, line 753: 🤖 _auto-generated by the rs-agentic bot_

Guideline (CLAUDE.md) — this makes OCCUPANCY a fully startable member, but `CLAUDE.md` § *GMSL logical-member lifecycle* (added by this PR) states "the current direct-XU adapter starts **ODPD only**" and "Runtime multi-member updates … remain unimplemented in the single-ODPD stage"; the PR body says "only ODPD is transport-available". Either keep occupancy out of `D500_MUX_SUPPORTED_MEMBERS` until it is qualified, or update `CLAUDE.md` in this commit — *Post-patch instruction hygiene* requires that review after every patch, and `cd58dee` touches only `d4xx.c`.

Same gap for `D500_MUX_VERSION` 1→2: `d500_mux_read()` now hard-rejects v1 FW with `-EOPNOTSUPP`, which is the FW the body's 2026-09-20 hardware evidence was collected on.

ymodlin (commented):
🤖 _auto-generated by the rs-agentic bot_

Review of the new commit `cd58dee` ("control multiple perception members", 88+/19- in `d4xx.c`). Only the MUX control/capture code changed.

**New in this push:** 2 bugs (STREAMON profile re-validation dropped; `mux_armed`/`mux_selected_mask` can diverge from the device), 2 nits, 1 CLAUDE.md gap.
**Carried over:** 5 open threads — none addressed by this push; replies in-thread. The JP5 build break (`MEDIA_BUS_FMT_RS_VARLEN_1X8` undefined in `kernel/kernel-5.10/`) is the blocker; CI is still `in_progress` on this head.

Also note the PR body is stale: it states head `381e338`, and its 2026-09-20 hardware evidence predates the `D500_MUX_VERSION` 1→2 ABI bump in this commit.

kernel/realsense/d4xx.c
ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `cd58dee`: this push touches only the MUX code in `d4xx.c`, so **still open**. `grep -rn MEDIA_BUS_FMT_RS_VARLEN_1X8 kernel/kernel-5.10/` is still empty while `d4xx.c:1263` references it — `build-jp502` / `build-jp512` / `build-jp5.1.6` will fail again (CI still `in_progress` on this head).

ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `cd58dee`: the format table is untouched — **still open**. `d4xx.c:1263` is still `{ RS_PIXFMT_RSVL, MEDIA_BUS_FMT_RS_VARLEN_1X8 }` with no `wire_dt`, while `SBGGR10P` on the line above pins `GMSL_CSI_DT_RAW_8`.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `cd58dee`: `nvidia-oot/` is untouched by this push — **still open**. Map 12 (`TX37`/`TX38`) remains statically `src_vc2 | GMSL_CSI_DT_RAW_16` while `__max96712_update_multi_vc_pipe()` writes the profile DT into map 8 for `vc_id == 2`.

kernel/realsense/d4xx.c
ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `cd58dee`: `d4xx.c:4632` is unchanged — **still open**. `ret = (s32)le32_to_cpu(reply.result)` still returns a positive FW status verbatim up to `VIDIOC_S_EXT_CTRLS`, where `ioctl() != -1` reads as success. The new DISABLE branch added to the `expected` ternary just above does not affect this.

kernel/realsense/d4xx.c
ymodlin: 🤖 _auto-generated by the rs-agentic bot_

Re-checked at `cd58dee`: **still open**. `d4xx.c:7936` still sets `sensor->streaming = true` before CAPTURE_START, and the `goto out` at 7944 (both exchanges failed) leaves it — and `ds5_dev->ir_streaming` — set. The next STREAMON short-circuits at 7898 and returns 0 without starting the camera.

kernel/realsense/d4xx.c
ymodlin: 🤖 _auto-generated by the rs-agentic bot_

`change approved` — the bare `memcmp(..., 20)` is gone and the hardcoded `stream_id == 5` is now `D500_MUX_STREAM_ODPD`.

One pointer: the replacement dropped profile-identity validation rather than bounding it; covered in a new comment on `d500_mux_capture()`.

kernel/realsense/d4xx.c
```diff
@@ -7450,6 +7870,116 @@ static void ds5_flush_idle_link(struct ds5 *state)
...
+		}
+#ifdef CONFIG_VIDEO_D4XX_SERDES
+		if (state->dser_ops->retrigger_datapath)
+			state->dser_ops->retrigger_datapath(state->dser_dev);
+#endif
+		goto out;
+	}
+	ret = d500_mux_exchange(state, &request, D500_MUX_CAPTURE_STOP);
+	if (ret)
+		goto out;
```
ymodlin, line 7956: 🤖 _auto-generated by the rs-agentic bot_

**BUG — failed CAPTURE_STOP leaks the transport and latches `streaming = true`.** This `goto out` skips `release_transport:` entirely, so `sensor->streaming`, `ds5_dev->ir_streaming` and the SerDes pipe all stay held.

`vb2` ignores the `stop_streaming` return value, so userspace sees STREAMOFF complete and frees its buffers. The next STREAMON then hits the `sensor->streaming == on` short-circuit at `d4xx.c:7898` and **returns 0 without starting the camera** — the app blocks on DQBUF forever, and `pipe_id` is never released for other sensors.

Same class as the start-path finding on line 7943, but a distinct path. Release the transport regardless (fall through to `release_transport:` and keep `ret`), or latch a fault the next start rejects.

ymodlin (commented):
🤖 _auto-generated by the rs-agentic bot_

**Re-review requested — head is unchanged (`cd58dee`), so no fixes have landed since the 2026-09-21 pass.** All 11 threads remain open. One new finding below; everything else is unchanged.

CI at this head now confirms the oldest finding:

| job | result |
|---|---|
| Build JP 5.0.2 / 5.1.2 / 5.1.6 | ❌ failure |
| Build JP 6.0 / 6.1 / 6.2 / 6.2.2 / 7.0 / 7.1 / 7.2 | ✅ success |

`grep -rn MEDIA_BUS_FMT_RS_VARLEN_1X8 kernel/kernel-5.10/` is still empty while `d4xx.c:1263` references it.

Blocking bugs still open: JP5 build break (`#4034678891`), `result`>0 returned as ioctl success (`#4055322651`), failed start latches `streaming=true` (`#4055322653`), STREAMON drops `d500_mux_validate_profile()` (`#4060599994`), `mux_armed`/`mux_selected_mask` divergence on CONFIGURE/DISABLE (`#4060599999`).

PR #665, opened 2026-09-17
[RSDEV-14847] MIPI desc refresh

ejgoldik:
- Refresh format tables upon HW reset
- The refresh is done by adding a "dirty" flag to the tegra channels using patching


kernel/nvidia/5.0.2/0026-Export-tegra_channel_refresh_fmts-for-runtime-format-list-changes.patch
```diff
@@ -0,0 +1,134 @@
...
++	bool fmts_dirty;		/* fmts_bitmap needs a runtime rebuild */
+ 	atomic_t power_on_refcnt;
+ 	struct v4l2_fh *fh;
+ 	bool bypass;
+@@ -365,6 +366,7 @@
+ int tegra_channel_cleanup(struct tegra_channel *chan);
+ int tegra_vi_channels_cleanup(struct tegra_mc_vi *vi);
+ int tegra_channel_init_subdevices(struct tegra_channel *chan);
++void tegra_channel_mark_fmts_dirty(struct tegra_channel *chan);
+ int tegra_channel_refresh_fmts(struct tegra_channel *chan);
```
ymodlin: **[BUG] rs-agentic (auto-generated)** — context line `int tegra_channel_refresh_fmts(struct tegra_channel *chan);` doesn't exist in `mc_common.h`. That symbol appears nowhere in this repo or in any NVIDIA tree (0 hits in a global code search), and the 5.1.2 / 6.2 versions of this same patch don't carry it.

`apply_patches.sh` runs a single `git apply <dir>/*`, so this hunk fails the whole JP5.0.2 patch set → CI break. Drop the line.

kernel/nvidia/5.0.2/0026-Export-tegra_channel_refresh_fmts-for-runtime-format-list-changes.patch
```diff
@@ -0,0 +1,134 @@
...
++ * ioctl.
++ */
++void tegra_channel_mark_fmts_dirty(struct tegra_channel *chan)
++{
++	if (chan)
++		WRITE_ONCE(chan->fmts_dirty, true);
++}
++EXPORT_SYMBOL(tegra_channel_mark_fmts_dirty);
++
+ /* The bitmap is otherwise built once, at subdev bind, so a sensor whose format
```
ymodlin: **[BUG] rs-agentic (auto-generated)** — same problem in `channel.c`: this trailing context comment ("D58x device mode" / "Caller must not hold this channel's video_lock") isn't in the baseline either — it looks like leftover context from the earlier, abandoned `tegra_channel_refresh_fmts` version. 5.1.2 and 6.2 use the real `/* Tegra channel frame setup ... */` block here.

Also note it contradicts the new comment above (`...which tegra_channel_fmts_bitmap_init() needs held`).

kernel/realsense/d4xx.c
```diff
@@ -1706,14 +1751,49 @@ static void ds5_desc_recheck(struct ds5 *state)
...
+	if (!changed) {
+		mutex_unlock(&dd->lock);
+		return;			/* common case: FW unchanged across reset */
+	}
+	/* Force a fresh fetch+parse. Publish CRC-dedups on the module-lifetime
+	 * list, so an unchanged blob costs nothing and a changed one adds one
+	 * table (the old is never freed - a sibling subdev may still use it).
+	 */
+	dd->desc_state = DS5_DESC_UNKNOWN;
+	dd->cached_device_type = DS5_DEVICE_TYPE_UNKNOWN;
```
ymodlin: **[BUG] rs-agentic (auto-generated)** — `cached_device_type` is cleared here and nothing re-populates it: `ds5_desc_load_locked()` never writes it, and `ds5_wait_device_type()` (recovery step 7) already ran *before* this.

So after any descriptor-changing reset it stays `UNKNOWN` for the module's life → `ds5_is_d58x()` returns false, and `ds5_rgb_ctrl_offset()` then picks the **D4xx** register offsets on a D585 (plus the D58x gates at ~4917/5403/6628/6850/8662).

Drop the line, or restore `dev_type` into it after the reload.

kernel/realsense/d4xx.c
```diff
@@ -9083,6 +9170,22 @@ static int ds5_probe(struct i2c_client *c
...
+	 * every node's VI channel to refresh its format list. ds5_dev and the
+	 * role flags are both set by now (serdes/board setup and cam-type parse
+	 * above).
+	 */
+	if (state->ds5_dev) {
+		int role_pad = ds5_state_to_pad(state);
+
+		if (role_pad >= DS5_MUX_PAD_DEPTH && role_pad < DS5_MUX_PAD_COUNT) {
+			mutex_lock(&state->ds5_dev->lock);
+			state->ds5_dev->role_inst[role_pad] = state;
```
ymodlin: **[BUG] rs-agentic (auto-generated)** — registering here is too early. `ds5_chrdev_init()`, `ds5_wait_device_type()` and `ds5_v4l_init()` all still follow, and their failure paths `return ret` without `ds5_remove()` — so the slot keeps a pointer to a torn-down instance. `ds5_init_ds5_dev()` doesn't clear `role_inst[]` either, so it survives a slot re-init. A later HW reset on a surviving sibling then walks it → use-after-free.

The DFU-recovery early `return 0` above also leaves an instance registered whose sensors/`variant` were never initialised — `ds5_apply_builtin_tables()` on it can deref `state->variant` (NULL) in the `default:` IR branch.

Suggest registering after `ds5_v4l_init()` succeeds, and clearing on the error paths.

kernel/realsense/d4xx.c
```diff
@@ -1706,14 +1751,49 @@ static void ds5_desc_recheck(struct ds5 *state)
...
+	 */
+	for (pad = DS5_MUX_PAD_DEPTH; pad < DS5_MUX_PAD_COUNT; pad++) {
+		struct ds5 *inst = insts[pad];
+
+		if (!inst)
+			continue;
+		if (have_desc)
+			ds5_desc_apply(inst);
+		else
+			ds5_apply_builtin_tables(inst, dev_type);
```
ymodlin, line 1806: **[BUG] rs-agentic (auto-generated)** — `ds5_apply_builtin_tables()` reads `inst->fw_version` and `inst->d58x_pixel_mode`, but recovery step 8 refreshes `fw_version` only on the *resetting* instance. Siblings still hold their probe-time values.

That's exactly the case this branch exists for: a FW downgrade across 5.16 leaves the IMU node on the pre-reset extended/legacy table choice. Pass the freshly read values in, or propagate them to the role instances before applying.

kernel/realsense/d4xx.c
```diff
@@ -1706,14 +1751,49 @@ static void ds5_desc_recheck(struct ds5 *state)
...
+	 * tables probe uses. Marking is a lock-free flag; the VI rebuilds the
+	 * bitmap on the next format ioctl, so the first query sees the new list.
+	 */
+	for (pad = DS5_MUX_PAD_DEPTH; pad < DS5_MUX_PAD_COUNT; pad++) {
+		struct ds5 *inst = insts[pad];
+
+		if (!inst)
+			continue;
+		if (have_desc)
+			ds5_desc_apply(inst);
```
ymodlin: **[BUG] rs-agentic (auto-generated)** — the swap isn't idle-gated. Recovery step 1 only writes `DS5_START_STOP_STREAM` (stops the FW) and clears the driver flags; VI/vb2 on a sibling node can still be in STREAMON. The kernel side correctly skips a busy channel, but `ds5_desc_apply()` / `ds5_apply_builtin_tables()` replace `sensor->formats`/`n_formats` unconditionally, so a streaming client's negotiated format can start resolving out of a table it never saw.

Suggest gating the per-instance apply on the `ds5_dev` streaming flags (same invariant the flush sites use).

Also: `ds5_desc_apply()`'s `false` return is ignored — a descriptor that covers only some streams leaves the rest silently on the old table.

nvidia-oot/6.2/0014-Export-tegra_channel_refresh_fmts-for-runtime-format-list-changes.patch
```diff
@@ -0,0 +1,135 @@
...
++ * is never changed underneath it; the channel stays dirty and retries later.
++ */
++static void tegra_channel_refresh_fmts_locked(struct tegra_channel *chan)
++{
++	if (!chan || !chan->subdev_on_csi || !READ_ONCE(chan->fmts_dirty))
++		return;
++	if (atomic_read(&chan->is_streaming) || vb2_is_busy(&chan->queue))
++		return;
++	tegra_channel_fmts_bitmap_init(chan);
++	chan->fmts_dirty = false;
```
ymodlin: **[BUG, minor] rs-agentic (auto-generated)** — plain write racing the `WRITE_ONCE()` mark: a mark landing between `tegra_channel_fmts_bitmap_init()` and this clear is lost, and the node keeps the stale list with no further trigger. Clear before the rebuild, or use `xchg(&chan->fmts_dirty, false)`. (Same in all four carriers.)

kernel/realsense/d4xx.c
```diff
@@ -9188,6 +9291,20 @@ static void ds5_remove(struct i2c_client *c)
 		state = i2c_get_clientdata(c);
 	}
 
+	/* Drop this instance from the reload registry, then flush its deferred
```
ymodlin: **[nit] rs-agentic (auto-generated)** — stale comment: there's no flush and no deferred work in this design (the rebuild is lazy, on the next ioctl). Reads as a leftover from the workqueue variant.

kernel/realsense/d4xx.c
```diff
@@ -804,6 +807,14 @@ struct ds5_dev {
...
 	bool rgb_streaming;
 	bool imu_streaming;
+
+	/* The camera's per-role probe instances, indexed by DS5_MUX_PAD_*
+	 * (see ds5_state_to_pad). Each owns its own mux subdev and VI channel;
+	 * the post-HW-reset reload uses them to refresh every node's published
+	 * format list. Written under lock at probe, snapshotted under lock on
+	 * reload.
+	 */
+	struct ds5 *role_inst[DS5_MUX_PAD_COUNT];
```
ymodlin, line 815: **[guideline] rs-agentic (auto-generated)** — CLAUDE.md *Comment style*: in-code comments are capped at 2–3 lines. This block and the `ds5_desc_reload_after_reset()` / `ds5_probe()` / `ds5_remove()` headers run 4–7. The rationale (deadlock reasoning, lazy-rebuild design) is already in the commit message and `docs/camera-format-descriptor.md` — trim the in-code copies.

docs/camera-format-descriptor.md
```diff
@@ -659,25 +659,38 @@ makes.
...
+
+A HW reset — the case a DFU/FW update ends in — can leave the camera serving a *different*
+descriptor. `ds5_desc_reload_after_reset()` re-reads it in the recovery path (after
+`ds5_hw_init()` has re-asserted the link mode) and refreshes what each node advertises **in
+place**, so a FW update's new formats appear on `/dev/videoN` without an `rmmod`/`modprobe`.
+
+The formats are already *published*, which is the part that needed kernel help: the Tegra VI
+channel builds its format bitmap from the subdev's mbus codes once, when the video node is
+registered, so a refreshed `sensor->formats` is otherwise invisible until the node is
+re-created. The `0014-Export-tegra_channel_refresh_fmts` patch (every JetPack carrier, since
```
ymodlin: **[nit] rs-agentic (auto-generated)** — name mismatch: the patch exports `tegra_channel_mark_fmts_dirty()`, not `tegra_channel_refresh_fmts`, and it's `0014` only on 6.x/7.x (JP5 ships it as `5.0.2/0026`, `5.1.2/0015`, `5.1.6/0017`). Worth renaming the files/subject so the symbol is greppable.

docs/camera-format-descriptor.md
```diff
@@ -659,25 +659,38 @@ makes.
...
+one `d4xx.c` builds against all of them) exports a re-enumeration entry point that rebuilds
+that bitmap; it refuses with `-EBUSY` while the node streams or holds buffers.
+
+The reload only fires when the header CRC actually changed across the reset (an unchanged FW
+is a no-op). It is idle-gated by construction — the recovery path stops every stream before
```
ymodlin: **[nit] rs-agentic (auto-generated)** — "idle-gated by construction — the recovery path stops every stream before it runs" isn't accurate: the recovery path stops the **FW** streams and clears the driver flags, it doesn't stop VI/vb2. See the related comment on `d4xx.c`.

ymodlin (commented):
Auto-generated review by the **rs-agentic** bot.

Nice mechanism — the lock-free dirty flag + lazy rebuild under `video_lock` is the right shape, and the carrier fan-out (JP5.0.2/5.1.2/5.1.6 + 6.x/7.x) is complete.

4 likely bugs + a few nits inline. Highest priority: the JP5.0.2 patch context can't apply, and `cached_device_type` is zeroed and never restored.

kernel/realsense/d4xx.c
```diff
@@ -9083,6 +9170,22 @@ static int ds5_probe(struct i2c_client *c
 		state->control_base = DS5_DEPTH_CONTROL_BASE;
 		state->control_status_reg = DS5_DEPTH_CONTROL_STATUS;
 	}
+
+	/* Register this per-role instance so the post-HW-reset reload can reach
```
ymodlin: **BUG** — the slot is published before every remaining probe failure path (`ds5_chrdev_init` → `goto e_chardev`/`e_regulator`, the device-type wait, `ds5_fixed_configuration()`, sensor/mux/v4l2 registration). `state` is `devm_kzalloc`'d (line 9041) and `ds5_remove()` is **not** called for a failed probe, so a failure after this point leaves `role_inst[pad]` pointing at freed memory. A later HW reset on a surviving sibling then dereferences it in `ds5_desc_reload_after_reset()`.

Register on the success path (just before `return 0`), or clear the slot at `e_chardev`/`e_regulator`.

kernel/nvidia/5.0.2/0026-Export-tegra_channel_refresh_fmts-for-runtime-format-list-changes.patch
ymodlin: **BUG** — `tegra_channel_refresh_fmts()` is declared nowhere in this repo (only inside this patch), and the JP5.1.2/6.2 twins of this same hunk don't have it. As a *context* line it makes the hunk unmatchable, and `apply_patches.sh` runs `git apply <dir>/*` — so JP5.0.2 fails to patch and `build-jp502` breaks.

Regenerate against a synced `sources_5.0.2` and verify with `git apply --check`.

kernel/nvidia/5.0.2/0026-Export-tegra_channel_refresh_fmts-for-runtime-format-list-changes.patch
```diff
@@ -0,0 +1,134 @@
...
+  * list varies at runtime (D58x device mode) keeps the stale set until reprobe.
```
ymodlin: **BUG** — same problem as the `mc_common.h` hunk: this trailing context comment is added by no patch in the tree (grep finds it only here), and the 5.1.2/6.2 variants use the stock `/* ---- Tegra channel frame setup ---- */` context at this spot. The hunk will be rejected.

kernel/realsense/d4xx.c
ymodlin: **BUG** — this clears the `cached_device_type` that `ds5_wait_device_type()` re-validated a few steps earlier in the same recovery flow, and nothing restores it. Every later `READ_ONCE(ds5_dev->cached_device_type) == DS5_DEVICE_TYPE_D58X` gate (lines 4917-4929, 5403-5429, 6628, 6850, 8662) then sees `UNKNOWN` and silently takes the non-D58x path for the rest of the module's life — on exactly the D585 FW-update flow this PR targets.

If the clear is only there to force a fresh parse, restore it (`dd->cached_device_type = dev_type;`) after `ds5_desc_load_locked()`.

kernel/realsense/d4xx.c
ymodlin, line 1806: **BUG** — `ds5_apply_builtin_tables()` picks the IMU table from `inst->fw_version` (`>= 0x510`) and `inst->d58x_pixel_mode`, but only the *reset* instance re-reads `DS5_FW_VERSION` (line 4396). Siblings still hold the pre-reset value, so a downgrade below 5.16 — the case that drops the descriptor and lands here — leaves the IMU node on the 32-bit extended table against a 16-bit FW.

Refresh each instance's `fw_version` (or pass the post-reset one in, like `dev_type`).

nvidia-oot/6.2/0014-Tegra_channel_runtime_format_changes.patch
```diff
@@ -0,0 +1,135 @@
...
++	tegra_channel_refresh_fmts_locked(chan);
++
+ 	if (f->index >= bitmap_weight(fmts_bitmap, MAX_FORMAT_NUM))
+ 		return -EINVAL;
+ 
+@@ -2094,6 +2131,8 @@
+ 	struct tegra_channel *chan = video_drvdata(file);
+ 	struct v4l2_pix_format *pix = &format->fmt.pix;
+ 
++	tegra_channel_refresh_fmts_locked(chan);
```
ymodlin, line 98: **Potential BUG** — `tegra_channel_fmts_bitmap_init()` does more than rebuild the bitmap: it re-selects `chan->fmtinfo` from the first supported hi-res code and re-runs `tegra_channel_update_format()`. Hooking it into `g_fmt` means a client that did `S_FMT` but hasn't called `REQBUFS` yet (`vb2_is_busy()` still false) has its negotiated format silently replaced on the next `G_FMT` — which the helper's own comment says never happens.

Either restore `chan->format`/`fmtinfo` around the rebuild, or drop the `g_fmt`/`try_fmt` hooks and keep only the enumeration ones.

nvidia-oot/6.2/0014-Export-tegra_channel_refresh_fmts-for-runtime-format-list-changes.patch
ymodlin: **BUG (race)** — plain store, while the setter uses `WRITE_ONCE`. A `tegra_channel_mark_fmts_dirty()` landing between `tegra_channel_fmts_bitmap_init()` and this clear is lost, and the node keeps the stale list until some later reset happens to re-mark it. Clear the flag *before* the rebuild (a spurious extra rebuild is harmless), or use `xchg`/`test_and_clear_bit`.

kernel/realsense/d4xx.c
ymodlin: Nit — there is no flush and no pending work in this driver; the comment describes machinery that doesn't exist. CLAUDE.md *Comment style*: 2–3 lines max, and no comments describing an absence. One line ("drop the slot before the instance is freed") is enough.

ymodlin (commented):
### rs-agentic bot review — auto-generated

Solid design (lock-free dirty flag + lazy rebuild under `video_lock` is the right call over a workqueue). Findings below, ordered by severity.

**Blocking**
1. `role_inst[]` is registered before every remaining probe failure path → dangling pointer to a `devm`-freed `state` → UAF on the next HW reset.
2. `kernel/nvidia/5.0.2/0026-…patch` carries context that exists nowhere in the tree → `git apply` rejects → JP5.0.2 CI break.

**Also**
3. `cached_device_type` cleared and never restored (D58x gates silently fall through).
4. Siblings' stale `fw_version` on the descriptor-gone path.
5. `g_fmt` rebuild can replace an already-negotiated format.
6. Lost-mark race on `fmts_dirty` clear.

**Housekeeping**
- Patch files are named `Export-tegra_channel_refresh_fmts…` but the exported symbol is `tegra_channel_mark_fmts_dirty()`; `tegra_channel_refresh_fmts()` does not exist. Rename to match, or the next reader greps for a symbol that isn't there.
- `d4xx.c:50` comment says "nvidia-oot 0014" — the JP5 carriers are `5.0.2/0026` and `5.1.2/0015`. Worth naming all three.
- **CLAUDE.md "Post-patch instruction hygiene"**: `CLAUDE.md` is not in this diff, but line 184 still documents `ds5_desc_recheck()` — the function this PR removed — including the whole "comparing earlier would report a false change" rationale. It needs updating to the new reload semantics in the same PR.
- **CLAUDE.md "Comment style" (2–3 lines max, no absence-describing comments)**: several new blocks run 6–10 lines (`ds5_desc_reload_after_reset()` header, `role_inst[]` field, both patch helpers). The rationale is already in `docs/camera-format-descriptor.md` §12 — point there instead of duplicating it in-code.

**Model / effort for the follow-ups** (per CLAUDE.md):
- Fix (1) probe-registration lifetime — Opus, effort high: UAF on a shared multi-instance registry, and the correct insertion point depends on which failure paths can still run.
- Regenerate (2) the JP5.0.2 patch against the real tree — Sonnet, effort medium: mechanical, but must be verified with `git apply --check` on a synced `sources_5.0.2`.
- Fix (3)/(4) post-reset state refresh — Opus, effort medium: small edits, but they decide which format table a node publishes.
- Fix (5)/(6) VI-side rebuild semantics — Opus, effort high: touches the ioctl path of every camera on every JetPack carrier.
- HW validation (FW upgrade + downgrade across a reset, plus a stream-during-reset case) — any tier / inherit session for the run; Opus, effort high for the dmesg/`v4l2-ctl --list-formats-ext` triage.

_This review was generated automatically by the rs-agentic bot; no approval and no commits were made._

kernel/nvidia/5.0.2/0026-Export-tegra_channel_refresh_fmts-for-runtime-format-list-changes.patch
ymodlin: change approved

kernel/nvidia/5.0.2/0026-Export-tegra_channel_refresh_fmts-for-runtime-format-list-changes.patch
ymodlin: change approved

kernel/realsense/d4xx.c

kernel/realsense/d4xx.c

kernel/realsense/d4xx.c
ymodlin: change approved

nvidia-oot/6.2/0014-Export-tegra_channel_refresh_fmts-for-runtime-format-list-changes.patch
ymodlin: change approved

kernel/realsense/d4xx.c

docs/camera-format-descriptor.md
ymodlin: change approved

kernel/realsense/d4xx.c

kernel/nvidia/5.0.2/0026-Export-tegra_channel_refresh_fmts-for-runtime-format-list-changes.patch
ymodlin: change approved

kernel/nvidia/5.0.2/0026-Export-tegra_channel_refresh_fmts-for-runtime-format-list-changes.patch
ymodlin: change approved

kernel/realsense/d4xx.c

kernel/realsense/d4xx.c
ymodlin: change approved

nvidia-oot/6.2/0014-Export-tegra_channel_refresh_fmts-for-runtime-format-list-changes.patch
ymodlin: change approved

kernel/realsense/d4xx.c

kernel/realsense/d4xx.c
ymodlin: Still open. `ds5_desc_apply(inst)` / `ds5_apply_builtin_tables(inst, ...)` still swap `sensor->formats`/`n_formats` unconditionally — recovery step 1 stops the **FW** streams and clears the driver flags, it does not stop VI/vb2. The new VI-side `vb2_is_busy()` gate only protects the published bitmap, not the subdev table a streaming client is resolving against.

`ds5_desc_apply()`'s `false` return is also still ignored (line 1773).

Not blocking if you consider the window unreachable in practice — but worth a line in the commit message saying why.

kernel/realsense/d4xx.c
ymodlin: Mostly addressed — thanks. Two blocks still over the 2–3 line cap: the loop header in `ds5_desc_reload_after_reset()` (5 lines, ~1767) and the `tegra_channel_refresh_fmts_locked()` / `tegra_channel_mark_fmts_dirty()` headers in the VI patches (6 lines each). The deadlock rationale is already in the patch commit message and `docs/camera-format-descriptor.md`.

docs/camera-format-descriptor.md
ymodlin: Still open — the four patch files are still named `...Export-tegra_channel_refresh_fmts...` while the exported symbol is `tegra_channel_mark_fmts_dirty()`. The patch `Subject:` line was fixed; the filenames were not, so the symbol is not greppable from the tree.

nvidia-oot/6.2/0014-Tegra_channel_runtime_format_changes.patch
ymodlin: Still open. The added gate is `atomic_read(&chan->is_streaming) || vb2_is_busy(&chan->queue)` — `vb2_is_busy()` is only true once buffers are allocated, so the `S_FMT` → `G_FMT` window (before `REQBUFS`) is unguarded, and `tegra_channel_fmts_bitmap_init()` re-picks `chan->fmtinfo` from the first enumerated code and re-runs `tegra_channel_update_format()`. A client that negotiated a format and then calls `G_FMT` gets a different one back.

Simplest fix: drop the `g_fmt`/`try_fmt`/`s_fmt` hooks and keep only the enumeration ones (`enum_fmt`, `enum_framesizes`, `enum_frameintervals`) — enumeration is what actually needs the fresh list.

kernel/realsense/d4xx.c
```diff
@@ -46,6 +46,9 @@
 #include <media/max96712.h>
 #include <media/max96717.h>
 #include <media/max96724.h>
+#ifdef CONFIG_TEGRA_CAMERA_PLATFORM
+#include <media/mc_common.h>	/* tegra_channel_mark_fmts_dirty() - nvidia-oot 0014 */
```
ymodlin: **BUG (build)** — this include lands *inside* the `#ifdef CONFIG_VIDEO_D4XX_SERDES` opened at line 43 (it closes at 123/129), but its only consumer, `ds5_mark_channel_fmts_dirty()` (line 1698), is guarded by `CONFIG_TEGRA_CAMERA_PLATFORM` alone.

On JP5 `VIDEO_D4XX_SERDES` is a user-selectable `bool` (`kernel/nvidia/5.0.2/0010-Add-Kconfig-VIDEO_D4XX_SERDES.patch`, `default y` — not forced), so `CONFIG_TEGRA_CAMERA_PLATFORM=y` + `CONFIG_VIDEO_D4XX_SERDES=n` fails to compile: `struct tegra_channel` incomplete, `tegra_channel_mark_fmts_dirty` undeclared.

Move it to its own top-level block, like the `camera_common.h` include at line 664.

kernel/realsense/d4xx.c
```diff
@@ -1706,14 +1740,53 @@ static void ds5_desc_recheck(struct ds5 *state)
...
+	if (!changed) {
+		mutex_unlock(&dd->lock);
+		return;			/* common case: FW unchanged across reset */
+	}
+	/* Force a fresh fetch+parse. Publish CRC-dedups on the module-lifetime
+	 * list, so an unchanged blob costs nothing and a changed one adds one
+	 * table (the old is never freed - a sibling subdev may still use it).
+	 */
+	dd->desc_state = DS5_DESC_UNKNOWN;
+	ds5_desc_load_locked(state);
```
ymodlin: **BUG** — the return value is dropped, and a *transient* failure here is indistinguishable from "descriptor gone".

`ds5_desc_check_magic()` already succeeded above (a real I2C error returns early), so we know the camera serves a descriptor. If `ds5_desc_fetch()`/`ds5_desc_build()` then fails for any non-`-ENODEV` reason — short read, CRC mismatch on a partial blob — `ds5_desc_load_locked()` sets `desc_state = DS5_DESC_ABSENT`, `have_desc` goes false, and **all four nodes are reverted to built-in per-SKU tables** and logged as "descriptor gone after HW reset". It stays that way until the next reset that happens to change the CRC.

Suggest restoring the previous `dd->desc`/`desc_state` and returning when `ds5_desc_load_locked()` fails with anything other than `-ENODEV`.

kernel/realsense/d4xx.c
ejgoldik: Fixed — the `mc_common.h` include moved to its own top-level `#ifdef CONFIG_TEGRA_CAMERA_PLATFORM` block, above (and independent of) the SERDES ifdef, matching its consumer's guard.

kernel/realsense/d4xx.c
ejgoldik: Fixed — the reload now snapshots `dd->desc`/`desc_state` before forcing the re-parse; on a non-`-ENODEV` failure from `ds5_desc_load_locked()` it restores both and returns, keeping the current tables. Only a genuine magic-mismatch (`-ENODEV`) path reverts to built-ins. Doc §12 updated with the same rule.

nvidia-oot/6.2/0014-Tegra_channel_runtime_format_changes.patch
ejgoldik: Fixed, via the other option: the rebuild is now bitmap-only — `chan->format`/`chan->fmtinfo` are saved and restored across `tegra_channel_fmts_bitmap_init()`, so the S_FMT→G_FMT window can't have the negotiated format re-picked (first-match by mbus code is lossy for formats sharing a code, e.g. the calib fourccs). Kept all six hooks rather than dropping s_fmt/try_fmt: those need the fresh bitmap so a client can negotiate a new-table format immediately post-reset. All three carriers regenerated; channel.o compiles rc=0.

kernel/realsense/d4xx.c
ejgoldik: Partially fixed: the `ds5_desc_apply()` return is now handled — on failure the instance falls back to `ds5_apply_builtin_tables()` (and `fw_version` is propagated before either branch). On the idle-gate: the swap stays ungated by design. It is memory-safe (tables are module-lifetime, both old and new stay live), and a d4xx-side gate on the streaming flags would be vacuous here — recovery clears them before the reload runs. The VI-side negotiated-format hazard is closed separately by making the bitmap rebuild restore `chan->format`/`fmtinfo`. Rationale added to doc §12 and will be in the commit message.

kernel/realsense/d4xx.c
ejgoldik: Trimmed — the reload loop header is 3 lines, and both VI-patch function headers are ≤3 lines (all three carriers regenerated).

docs/camera-format-descriptor.md
ejgoldik: Renamed — all nine carriers are now `…-Export-tegra_channel_mark_fmts_dirty-for-runtime-format-list-changes.patch` (real: `nvidia-oot/6.2/0014`, `kernel/nvidia/5.1.2/0015`, `5.0.2/0026`; symlinks re-pointed: 6.0/6.1/7.0/7.1/7.2 and 5.1.6/0017). Doc reference updated.

ejgoldik: Shortened per offline feedback: the files are now `0014-Tegra_channel_runtime_format_changes.patch` (JP6/7 carriers + symlinks) and `5.0.2/0026` / `5.1.2/0015` / `5.1.6/0017` of the same name.

ejgoldik:
## HW validation — post-reset table reload without driver reload

Validated on a rig with four DS5 FW builds, switched via RealSense Viewer FW update (DFU → HW reset), checking `v4l2-ctl --list-formats-ext` after each update completed — **no `rmmod`/`modprobe` at any point**:

| FW build | Descriptor served | Expected post-update result |
|---|---|---|
| Single depth resolution | yes | depth node lists only that resolution |
| Single color resolution (regular depth) | yes | color node shrinks, depth unaffected |
| 5.17.3.12 (pre-descriptor) | **no** | all nodes revert to built-in per-SKU tables |
| Default (full tables) | yes | full format set restored |

Every transition — descriptor→descriptor (changed CRC), descriptor→absent (downgrade), and absent→descriptor (upgrade) — updated the enumerated formats in place upon update completion. This also exercises the deadlock fix: the reload runs in the reset node's `s_ctrl` and the first post-reset `--list-formats-ext` triggers the lazy VI bitmap rebuild.

kernel/realsense/d4xx.c
```diff
@@ -1706,14 +1743,57 @@ static void ds5_desc_recheck(struct ds5 *state)
...
-			 "format descriptor changed across reset (crc 0x%08x -> 0x%08x); reload d4xx to apply\n",
-			 dd->desc->crc, crc);
-	else if (dd->desc_state == DS5_DESC_ABSENT && present)
-		dev_warn(&state->client->dev,
-			 "camera now serves a format descriptor; reload d4xx to apply\n");
+	if (dd->desc_state == DS5_DESC_READY)
+		changed = !present || crc != dd->desc->crc;
+	else
+		changed = present;	/* on static tables; a descriptor appeared */
+	if (!changed) {
```
ymodlin: **[BUG] rs-agentic (auto-generated)** — the whole refresh is gated on a descriptor change, but one of the tables it refreshes is gated on `fw_version` instead.

On a camera that serves no descriptor (D457 and every legacy SKU) `desc_state` is `ABSENT` and `present` is false, so `changed` is false and we return here. A DFU that crosses the 5.16 boundary — exactly a FW update ending in this HW reset — therefore leaves the IMU node on `ds5_imu_formats` (16-bit) against a 32-bit FW, and leaves every sibling's `fw_version` at its probe-time value (only `state` re-read it at step 8).

Suggest also treating a changed `fw_version` as a trigger, e.g. snapshot it before step 8 and `changed |= (new != old)`.

kernel/realsense/d4xx.c
```diff
@@ -1706,14 +1743,57 @@ static void ds5_desc_recheck(struct ds5 *state)
...
+		/* Transient fetch/parse failure, not "no descriptor" (the magic
+		 * matched above): keep serving the previous tables.
+		 */
+		dd->desc = prev_desc;
+		dd->desc_state = prev_state;
+		mutex_unlock(&dd->lock);
+		return;
+	}
+	have_desc = dd->desc_state == DS5_DESC_READY;
+	memcpy(insts, dd->role_inst, sizeof(insts));
```
ymodlin: **[BUG, minor] rs-agentic (auto-generated)** — the snapshot is taken under `dd->lock` but used after it is dropped. A concurrent `ds5_remove()` on a sibling clears its `role_inst[]` slot and then lets devm free the `state`, so the loop below can write `inst->fw_version` / `inst->*.sensor.formats` into freed memory.

Narrow (needs a sysfs unbind racing a reset ioctl), but the snapshot is what re-opens the window the `role_inst[]` locking otherwise closes.

docs/camera-format-descriptor.md
```diff
@@ -659,25 +659,46 @@ makes.
...
-`sensor->formats` is not seen by user space until the node is re-created. A V4L2 patch that
-allowed the published format list to be refreshed in place was written for this driver in
-the past and abandoned; this follow-up is what would justify reviving it.
-
-- Driver side: reload the blob in the recovery path (after `ds5_hw_init()` has re-written
-  the link mode), and swap `sensor->formats` only while no stream is active — the
-  module-lifetime table list already makes the old tables safe to keep.
-- Kernel side: revive the VI/V4L2 refresh patch across every JetPack carrier, since one
-  `d4xx.c` builds against all of them.
+## 12 · Reloading the descriptor after a HW reset
```
ymodlin, line 664: **[doc] rs-agentic (auto-generated)** — §12 is rewritten but the invariant bullet at line 589 still states the opposite: "**After a HW reset the descriptor is re-probed, not re-applied.** `ds5_desc_recheck()` … A changed or newly present descriptor is logged with "reload d4xx to apply"; formats are never swapped underneath registered V4L2 nodes (see section 12)." That function no longer exists and formats *are* now swapped. Please update it to match.

CLAUDE.md
```diff
@@ -183,6 +183,7 @@ Each MIPI segment's lane count comes from a **different** DT property, so the ca
...
...e. **under that node's VI `video_lock`** — it must never rebuild a VI format bitmap directly (self-deadlock; HW-proven hang). It only sets the lock-free per-channel `fmts_dirty` mark via `tegra_channel_mark_fmts_dirty()` (VI patch `0014-Tegra_channel_runtime_format_changes.patch` on 6.x/7.x; JP5: `5.0.2/0026`, `5.1.2/0015`, `5.1.6/0017` symlink); the VI rebuilds lazily on the next format ioctl, so the first post-reset query sees the fresh list. The rebuild is **bitmap-only**: `chan->format`/`fmtinfo` are saved and restored across `tegra_channel_fmts_bitmap_init()`, which otherwise re-picks `fmtinfo` first-match by mbus code — lossy for fourccs sharing a code. In the reload, a **non-`-ENODEV`** `ds5_desc_load_locked()` failure restores the previous `desc`/`desc_state` (the magic already matched, so "descriptor gone" would be false); only a magic mismatch reverts nodes to built-in tables (`ds5_apply_builtin_tables()`, with the reset instance's `fw_version` propagated to siblings first). `ds5_dev->role_inst[]` slots register at the **end** of successful probe and clear in `ds5_remove()`, both under `ds5_dev->lock` — registering earlier leaves a dangling pointer on probe-failure paths.
```
ymodlin: **[guideline] rs-agentic (auto-generated)** — the bullet directly above (line 184) still names `ds5_desc_recheck()` twice and describes the old "report a spurious 'descriptor changed' telling the operator to reload the driver" behaviour. Rename to `ds5_desc_reload_after_reset()` and reword, otherwise the file carries two contradicting rules for the same call site.

ymodlin (commented):
Re-review of the new commit. The five fixes from the last round check out: the `mc_common.h` include is now under its own `CONFIG_TEGRA_CAMERA_PLATFORM` block, the non-`-ENODEV` reload failure restores `desc`/`desc_state`, the VI rebuild is bitmap-only (clear-before-rebuild + `format`/`fmtinfo` save/restore), all nine patch carriers are renamed/re-symlinked, and the in-code comment blocks are within the 2-3 line cap. `tegra_channel_fmts_bitmap_init()` is defined above the insertion point in both the JP5 and JP6/7 trees, so the new static helper compiles without a forward declaration.

Four items below.

kernel/realsense/d4xx.c

kernel/realsense/d4xx.c
ymodlin: change approved — verified: `prev_desc`/`prev_state` are restored on a non-`-ENODEV` failure, and `ds5_desc_load_locked()` leaves `dd->desc` untouched on that path, so the restore is consistent.

nvidia-oot/6.2/0014-Tegra_channel_runtime_format_changes.patch
ymodlin: change approved — the save/restore around `tegra_channel_fmts_bitmap_init()` closes the S_FMT→G_FMT window, and keeping all six hooks is the right call. (Minor, non-blocking: `tegra_channel_fmts_bitmap_init()` also calls `update_gang_mode()` from the *new* format before the restore, so the gang params can be left derived from a format that is then rolled back — irrelevant at `total_ports == 1`, which is every D4xx config.)

kernel/realsense/d4xx.c
ymodlin: change approved — `ds5_desc_apply()`'s return is handled now (falls back to `ds5_apply_builtin_tables()`), and I agree the d4xx-side gate would be vacuous: recovery clears the streaming flags before the reload. The swap is memory-safe given module-lifetime tables, and the VI-side negotiated-format hazard is closed by the `chan->format`/`fmtinfo` restore. Rationale in doc §12 is enough.

kernel/realsense/d4xx.c
ymodlin: change approved — all the new blocks in `d4xx.c` and the three VI patches are within the 2–3 line cap.

docs/camera-format-descriptor.md
ymodlin: change approved — verified all nine carriers resolve: real files at `nvidia-oot/6.2/0014`, `kernel/nvidia/5.0.2/0026`, `5.1.2/0015`; symlinks for 6.0/6.1/7.0/7.1/7.2 and 5.1.6/0017; 6.2.1 and 6.2.2 are directory symlinks to 6.2. No leftovers under the old name, and the numbering does not collide with the existing 00xx patches in any carrier.

kernel/realsense/d4xx.c
ymodlin: Re-opening this partially. The propagation itself is correct (`inst->fw_version = state->fw_version;` before the apply), but it only runs when the *descriptor* changed — see the new comment at the `if (!changed)` early return. On a camera with no descriptor the refresh never runs, so a DFU across 5.16 still leaves the IMU node on the 16-bit table.

kernel/realsense/d4xx.c
ejgoldik: Fixed — the reload now also triggers on `fw_changed`: under `dd->lock` it compares every `role_inst[]` sibling's `fw_version` against the freshly read `state->fw_version` (siblings hold the pre-reset value, so any mismatch means the DFU changed it). On a no-descriptor camera this takes the built-in path and re-applies `ds5_apply_builtin_tables()` with the new `fw_version`, covering the 5.16 IMU boundary. A fw-only change on a descriptor camera skips the re-fetch (descriptor CRC unchanged) and just re-applies + propagates.

kernel/realsense/d4xx.c
ejgoldik: Closed by the `fw_changed` trigger — see the reply on the `if (!changed)` thread: the early return now also fires the refresh when any sibling's `fw_version` differs from the post-reset read.

kernel/realsense/d4xx.c
ejgoldik: Fixed — the snapshot is gone; the apply loop now runs entirely under `dd->lock`, reading `dd->role_inst[pad]` directly. That's safe because everything in the loop is lock-free or a pointer swap (no I2C): `ds5_desc_apply()` was split into a lock-held core (`__ds5_desc_apply()`) used by the loop and a locking wrapper for the probe path. `ds5_remove()` clears the slot under the same lock before devm frees the instance, so the window is closed.

docs/camera-format-descriptor.md
ejgoldik: Fixed — the line-589 invariant bullet now describes the reload-in-place behavior (`ds5_desc_reload_after_reset()`, tables swapped or reverted to built-ins, safe under concurrent readers via module-lifetime tables) and points to §12; "Never on the stream-start path" now reads "Probe and post-reset recovery only".

CLAUDE.md
ejgoldik: Fixed — the link-context bullet now names `ds5_desc_reload_after_reset()` and describes the current failure mode for the non-prototype-D585 path (wrong-mode tables swapped in, rather than the old "spurious descriptor-changed warning").

kernel/realsense/d4xx.c
ymodlin: change approved — verified `fw_changed` closes the no-descriptor DFU gap: the `!changed` early return now also fires when any `role_inst[]` sibling's `fw_version` differs from the post-reset read, and the `!have_desc` path re-runs `ds5_apply_builtin_tables()` after `inst->fw_version` is propagated, so the IMU table follows across the 5.16 boundary. The re-fetch stays gated on `changed`, so a fw-only change costs no extra I2C.

Two non-blocking notes:
- The scan compares siblings against `state->fw_version`, which was already overwritten at step 8 — so a change is invisible if the resetting instance is the *only* registered slot. Not reachable on a real camera (probe registers every role), but snapshotting the pre-read value would remove the blind spot.
- The `!have_desc` log lost the "descriptor gone" wording; that branch now reads the same for a genuine descriptor loss and for a plain fw-only refresh on a no-descriptor SKU.

kernel/realsense/d4xx.c
ymodlin: change approved — the `insts[]` snapshot is gone and the apply loop reads `dd->role_inst[pad]` under `dd->lock`, so the `ds5_remove()` race is closed. Verified the loop body is lock-safe: `__ds5_desc_apply()` takes no lock and can't reach `ds5_desc_load_locked()` (it only runs when `desc_state == READY`), `ds5_apply_builtin_tables()` is pure pointer assignment, and `ds5_mark_channel_fmts_dirty()` ends in a `WRITE_ONCE()`. No I2C and no nested lock under `dd->lock`.

docs/camera-format-descriptor.md
ymodlin: change approved — the line-589 bullet now describes the reload-in-place behavior and points to §12, and "Never on the stream-start path" reads "Probe and post-reset recovery only". No `ds5_desc_recheck()` references left in the doc.

CLAUDE.md
ymodlin: change approved — the link-context bullet names `ds5_desc_reload_after_reset()` and the failure mode now reads "swap in the tunnel-mode tables under a pixel-mode link", which matches the new semantics. The §12 bullet below it also documents the `fw_version` trigger and the lock-held apply loop.

kernel/realsense/d4xx.c
ymodlin: change approved — closed by the `fw_changed` trigger. `inst->fw_version = state->fw_version` now runs on both branches, and the no-descriptor case reaches `ds5_apply_builtin_tables()` with the fresh value, so the IMU table is re-picked across 5.16.

ymodlin:
### rs-agentic bot — incremental review of `38ffe9f` (auto-generated)

Reviewed only the delta since the last reviewed head (`1a14a61` → `38ffe9f`): `CLAUDE.md`, `docs/camera-format-descriptor.md`, `kernel/realsense/d4xx.c`. The VI patches are unchanged from the round already approved.

**No new bugs found.** All four open threads are resolved and replied to inline:

- `fw_changed` trigger — the reload now also fires on a sibling `fw_version` mismatch, closing the no-descriptor DFU-across-5.16 gap for the built-in IMU table. Re-fetch stays gated on `changed`, so a fw-only change costs no extra I2C.
- Apply loop moved under `dd->lock` (snapshot removed, `__ds5_desc_apply()` split out as the lock-held core). Verified nothing in the loop takes a lock or does I2C: `__ds5_desc_apply()` can't reach `ds5_desc_load_locked()` on this path, `ds5_apply_builtin_tables()` is pointer assignment only, `ds5_mark_channel_fmts_dirty()` ends in `WRITE_ONCE()`.
- Doc §589 invariant and the `CLAUDE.md` link-context bullet both updated to the reload-in-place semantics; no `ds5_desc_recheck()` references left in the tree.

Two non-blocking nits noted inline (fw compare happens after `state->fw_version` is already overwritten; the `!have_desc` log no longer distinguishes "descriptor gone" from a fw-only refresh). Neither blocks.

Still outstanding from earlier rounds: HW validation of the FW upgrade **and** downgrade across a reset on both a descriptor SKU and a no-descriptor SKU (`v4l2-ctl --list-formats-ext` before/after), plus `git apply --check` of the four JP5 patch carriers.

_No approval and no commits were made by this bot._

kernel/realsense/d4xx.c
```diff
@@ -1706,14 +1753,69 @@ static void ds5_desc_recheck(struct ds5 *state)
...
+	 * same lock before its instance is freed, so no slot can go stale here.
+	 */
+	for (pad = DS5_MUX_PAD_DEPTH; pad < DS5_MUX_PAD_COUNT; pad++) {
+		struct ds5 *inst = dd->role_inst[pad];
+
+		if (!inst)
+			continue;
+		/* Share the post-reset FW version; only @state re-read it. */
+		inst->fw_version = state->fw_version;
+		if (!have_desc || !__ds5_desc_apply(inst))
```
ymodlin, line 1805: **[BUG, minor] rs-agentic (auto-generated)** — `formats`/`n_formats` are swapped as an unsynchronized pair, and readers on *sibling* nodes are not excluded here.

`ds5_desc_apply_sensor()` writes `sensor->formats` then `sensor->n_formats`. `ds5_sensor_enum_mbus_code()` reads them in the opposite order (`index >= n_formats` guard, then `formats[index]`), and `ds5_sensor_enum_frame_size()` walks `formats` bounded by `n_formats`. Neither takes `dd->lock` or the reset node's `video_lock`, and neither is gated on streaming — so this isn't covered by the "recovery clears the streaming flags" argument that closed the earlier idle-gating thread.

Failure: `VIDIOC_ENUM_FMT` on `/dev/video-rs-ir-0` (or the VI's own `tegra_channel_fmts_bitmap_init()` rebuild, which loops `enum_mbus_code`) runs while the depth node's `s_ctrl` is in this loop. The reader loads the old `n_formats` (say 6, built-in Y8 table), then the new `formats` from a 2-entry descriptor table → `formats[5]` reads past the kmalloc'd `ds5_desc` table. Garbage `mbus_code`/`resolutions` pointer; KASAN slab-out-of-bounds.

Shrink-safe swap in `ds5_desc_apply_sensor()` (and the built-in path) costs nothing:
```c
sensor->n_formats = 0;
sensor->formats = st->formats;
sensor->n_formats = st->n_formats;
```
A concurrent reader then sees either the old pair or an empty list, never a mismatched one.

ymodlin (commented):
**rs-agentic re-review** (head `38ffe9f`) — _auto-generated by the rs-agentic bot._

All 14 earlier threads are addressed and were approved on this head; CI is green so far (7 build jobs pass, 3 pending). Re-read the full diff for new issues — one left, inline.

Verified this pass, no action needed:
- Patch carriers: `0014` (6.0/6.1/6.2 real + 7.0/7.1/7.2 symlinks, 6.2.1/6.2.2 are dir symlinks to 6.2), `5.0.2/0026`, `5.1.2/0015`, `5.1.6/0017` — numbering is free in every dir, `apply_patches.sh` globs so no list to update. Hunk line counts/offsets are internally consistent.
- `ds5_mark_channel_fmts_dirty()`: `tegra_channel_init_subdevices()` sets `v4l2_set_subdev_hostdata(sd, chan)` on every subdev in the channel's graph walk, and each d4xx role instance maps to exactly one VI channel (the `-md` node is a second `video_device` inside the same `tegra_channel`, not a second channel), so the mark reaches the right node.
- Apply loop under `dd->lock` really is I2C-free: `__ds5_desc_apply()` can only re-enter `ds5_desc_load_locked()` when `desc_state == UNKNOWN`, and `have_desc` short-circuits that branch.

Not blocking, no inline comment: `fmts_dirty` is inserted mid-`struct tegra_channel` rather than appended. Safe here (no prebuilt OOT module references `struct tegra_channel`; the whole camera module set rebuilds together), but `0013` sets the trailing-field precedent for this repo.

kernel/realsense/d4xx.c
ejgoldik: Fixed, but with reader-side locking instead of the shrink-safe swap: the suggested store ordering still leaves a window on ARM64's weak memory model (a reader whose stale `n_formats` load pairs with a fresh `formats` load — writer-side ordering alone can't exclude it without reader barriers). Since the reload's apply loop already publishes the pair under `ds5_dev->lock`, the readers now snapshot under the same lock: `ds5_sensor_enum_mbus_code/enum_frame_size/enum_frame_interval`, `ds5_sensor_find_format` (set_fmt path), and the mux EXTERNAL-pad count snapshot. Lock order is the established `video_lock → state->lock → ds5_dev->lock` (s_ctrl already nests state→dd; an awk scan confirms no dd→state or dd→enum nesting anywhere). Table *contents* are immutable module-lifetime, so the lock covers only the pair read — a returned `fmt` pointer stays valid after unlock. Also fixed a pre-existing past-the-end deref in `ds5_sensor_find_format()`: the dev_dbg dereferenced `fmt` one past the table when the mbus code wasn't found. Compiles rc=0, checkpatch clean.

ymodlin: Reader-side locking is the right call and it does close the torn pair. Verified independently: `ds5_dev->lock` stays innermost (`video_lock -> state->lock -> dd->lock`) - no site takes `state->lock` or a `video_lock` under it, the only I2C under `dd->lock` is short single-register writes plus the reload's own `ds5_desc_load_locked()`, and `ds5_mux_enum_mbus_code()` drops the lock before forwarding, so the VI bitmap rebuild (`enum_mbus_code` in a loop under `video_lock`) cannot recurse into it.

Not approving yet - the `ds5_sensor_find_format()` leg derives `state` from `v4l2_get_subdevdata()`, which is still NULL on the probe path. New inline comment at line 2920.

kernel/realsense/d4xx.c
```diff
@@ -2688,30 +2911,32 @@ static int ds5_sensor_get_fmt(struct v4l2_subdev *sd,
...
 }
 
-/* Called with lock held */
+/* Called with state->lock held */
 static const struct ds5_format *ds5_sensor_find_format(
 		struct ds5_sensor *sensor,
 		struct v4l2_mbus_framefmt *ffmt,
 		const struct ds5_resolution **best)
 {
+	struct ds5 *state = v4l2_get_subdevdata(&sensor->sd);
```
ymodlin, line 2920: **[BUG] rs-agentic (auto-generated)** — `v4l2_get_subdevdata(&sensor->sd)` is NULL on the probe path, so the `mutex_lock(&state->ds5_dev->lock)` two lines down oopses.

`ds5_v4l_init()` → `ds5_parse_cam()` → `ds5_fixed_configuration()` calls `__ds5_sensor_set_fmt()` (8232 / 8242) **before** `ds5_depth_init()` / `ds5_ir_init()` reach `ds5_sensor_init()`, which is the only place `v4l2_set_subdevdata(sd, state)` runs (line 7023). `state` is `devm_kzalloc`'d, so `sensor->sd.dev_priv` is still NULL there. The old code survived it because the only use was `dev_dbg(sensor->sd.dev, ...)`.

Gated on `(cfg0 & 0xff00) == 0x1800` (YUV420 depth DT at probe), so not every boot — but it is a NULL deref when it fires, and it is the one caller of this helper.

`__ds5_sensor_set_fmt()` already holds `state`; pass it in instead of re-deriving it.

kernel/realsense/d4xx.c
```diff
@@ -6912,19 +7138,25 @@ static int ds5_mux_enum_mbus_code(struct v4l2_subdev *sd,
 		remote_sd = &state->imu.sensor.sd;
 		break;
 	case DS5_MUX_PAD_EXTERNAL:
-		if (mce->index >= state->ir.sensor.n_formats +
-				state->depth.sensor.n_formats)
+		/* Snapshot the counts; the forwarded call re-locks and
+		 * re-checks its own bounds.
+		 */
+		mutex_lock(&state->ds5_dev->lock);
```
ymodlin, line 7144: **[nit] rs-agentic (auto-generated)** — the `/* No locking needed */` header on this function (line 7111) is now wrong — it takes `ds5_dev->lock`. CLAUDE.md *Post-patch instruction hygiene*: update a helper's locking contract in the same change.

ymodlin (commented):
Re-review of the reader-side locking push. One blocker, one nit; the rest of the locking change checks out (details in the thread reply on the torn-pair comment).

PR #668, opened 2026-09-18
Add FG12-12CH D457 suppor

mengyui:


nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
```diff
@@ -1340,13 +1344,13 @@ index 8e237ee..c1e57e2 100644
 +	}
 +
 +	if (priv->fangzhu_fg12_reverse_lane_polarity ||
-+	    priv->fangzhu_fg12_poc_enable) {
+	    priv->fangzhu_fg12_4ch_poc_enable || priv->fangzhu_fg12_12ch_poc_enable) {
```
ymodlin: **BUG — this patch will no longer apply.** The line lost its leading `+`, so inside the patch it is now a *context* line instead of an added one.

The enclosing hunk header (line 542) declares `@@ -284,13 +631,846 @@`, but the body now contains only **845** added lines, and the context line does not exist in the original `drivers/media/i2c/max96712.c`. `./apply_patches.sh apply 6.0` will fail on this hunk.

```diff
-	    priv->fangzhu_fg12_4ch_poc_enable || priv->fangzhu_fg12_12ch_poc_enable) {
++	    priv->fangzhu_fg12_4ch_poc_enable || priv->fangzhu_fg12_12ch_poc_enable) {
```

The equivalent 5.0.2 / 5.1.2 patches are fine — only this copy is affected.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
```diff
@@ -574,19 +578,19 @@ index 8e237ee..c1e57e2 100644
 +
 +static int max96712_enable_poc(struct max96712 *priv)
 +{
-+	struct reg_pair poc_enable[] = {
-+		{MAX96712_FANGZHU_POC_IO_ADDR, 0xe0},
-+		{MAX96712_FANGZHU_FG12_POC_ADDR, 0x80},
-+	};
 +	int err;
++	u16 fg12_poc_addr = MAX96712_FANGZHU_FG12_POC_ADDR;
```
ymodlin: **BUG — undefined macro, build break.** `MAX96712_FANGZHU_FG12_POC_ADDR` was renamed to `MAX96712_FANGZHU_FG12_4CH_POC_ADDR` (line 134 of this patch), but this use still references the old name.

```c
u16 fg12_poc_addr = MAX96712_FANGZHU_FG12_4CH_POC_ADDR;
```

The 5.0.2 (line 526) and 5.1.2 (line 609) copies already use the new name — only the 6.0 copy was missed. Net effect on JP6.0: 4-channel POC is broken as well, not just 12-channel.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
```diff
@@ -574,19 +578,19 @@ index 8e237ee..c1e57e2 100644
...
++	if (priv->fangzhu_fg12_12ch_poc_enable)
++		fg12_poc_addr = MAX96712_FANGZHU_FG12_12CH_POC_ADDR;
++
++	err = max96712_write_reg(priv, MAX96712_FANGZHU_POC_IO_ADDR, 0xe0);
++	err |= max96712_write_reg(priv, fg12_poc_addr, 0x80);
```
ymodlin: **BUG — `|=` accumulated onto an errno.** OR-ing negative error codes produces a meaningless value (`-EIO | -EINVAL` == `-5 | -22` == `-1`/`-EPERM`), and `if (err) return err;` then hands that to `max96712_power_on()`. It also masks *which* write failed.

Check each write separately:

```c
err = max96712_write_reg(priv, MAX96712_FANGZHU_POC_IO_ADDR, 0xe0);
if (err)
	return err;
err = max96712_write_reg(priv, fg12_poc_addr, 0x80);
if (err)
	return err;
```

Same at line 593 (`err |= ... 0x90`) and in the 5.0.2 / 5.1.2 copies of this function.

ymodlin (commented):
**Automated review by the rs-agentic bot.** External contribution — scope limited to potential bugs.

Two blockers, both isolated to `nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch`, which diverges from the (correct) `kernel/nvidia/5.0.2` and `kernel/nvidia/5.1.2` copies of the same change: the JP6.0 patch will neither apply nor compile. A third, lower-severity error-handling bug is present in all three copies. Details inline.

The rest looks consistent: all three patch copies that actually carry the FangZhu POC feature were updated for the `fangzhu,fg12-poc-enable` → `fangzhu,fg12-4ch-poc-enable` rename (6.1/6.2/7.x do not contain it), the three existing 4-channel overlays were renamed to match, the new `.dtbo` is registered in the JP6 overlay Makefile, and `link-mask = <0x1>` + `maxim,link0-vc-remap` in the new 12-channel overlay matches the single-serializer / 4-VC topology already used by `tegra234-camera-d4xx-overlay-fg12-16ch-cams-0-4.dts`.

Also note the PR is `behind` `dev`; a rebase before merge is advisable.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
```diff
@@ -574,19 +578,20 @@ index 8e237ee..c1e57e2 100644
...
++		fg12_poc_addr = MAX96712_FANGZHU_FG12_12CH_POC_ADDR;
 +
-+	err = max96712_set_registers(priv, poc_enable,
-+				     ARRAY_SIZE(poc_enable));
++	err = max96712_write_reg(priv, MAX96712_FANGZHU_POC_IO_ADDR, 0xe0);
 +	if (err)
 +		return err;
++	err = max96712_write_reg(priv, fg12_poc_addr, 0x80);
++	if (err)
++		return err
```
ymodlin, line 591: **BUG — missing semicolon, build break.**

```c
		return err
```

This patch is symlinked by JP6.1/6.2/6.2.1/6.2.2 and by `nvidia-oot/7.0/0006` + `7.1/0006`, so `max96712.c` fails to compile on every JP6.x and JP7.x build.

`kernel/nvidia/5.0.2/0017` (line 535) has it right; only this copy and `5.1.2/0007` are wrong.

kernel/nvidia/5.1.2/0007-Adding-max96712-support-for-D4xx.patch
```diff
@@ -601,19 +605,20 @@ index 329a10d..c8bde22 100644
...
++		fg12_poc_addr = MAX96712_FANGZHU_FG12_12CH_POC_ADDR;
 +
-+	err = max96712_set_registers(priv, poc_enable,
-+				     ARRAY_SIZE(poc_enable));
++	err = max96712_write_reg(priv, MAX96712_FANGZHU_POC_IO_ADDR, 0xe0);
 +	if (err)
 +		return err;
++	err = max96712_write_reg(priv, fg12_poc_addr, 0x80);
++	if (err)
++		return err
```
ymodlin, line 618: **BUG — same missing semicolon as the 6.0 copy.** `return err` → `return err;`. Breaks the JP5.1.2 (and 5.1.6 symlink) build.

ymodlin (commented):
Re-review of `3c6c771`. Previous three findings are addressed — but the fix introduced a new build break in 2 of the 3 patch carriers. See inline.

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
ymodlin: change approved

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
ymodlin: change approved

nvidia-oot/6.0/0003-Adding-max96712-support-for-D4xx.patch
ymodlin: Error handling is fixed, but the new `return err` on the line above is missing its `;` — build break on 6.0 (and the 5.1.2 copy). See the new inline comment.

`err |= ... 0x90` is now harmless (`err` is provably 0 there), but plain `err =` would be clearer.

PR #670, opened 2026-09-18
[RSDEV-14893] Allow MinZ runtime updates and valid defaults

lemon6630:
## Summary

- expose the canonical per-slot MinZ ranges, steps, and defaults through the shared composite-control type operations
- allow enable, shift, and threshold fields to update while Depth or IR is streaming
- keep filter type and downscale ratio pre-stream-only by comparing them with the live firmware payload and returning EBUSY on a streaming change
- serialize the streaming-state check and MinZ wire transaction against stream state changes
- preserve MinZ out-of-range rejection instead of adopting the shared clamp behavior
- preserve the existing seven-element XU wire ABI; decimation and temporal are unchanged

## Validation

- checkpatch at 80 columns: 0 errors, 0 warnings
- JP 6.2 full kernel/module build: PASS
- JP 6.2 and JP 7/Linux 6.8 D4xx target compilation with Werror: PASS
- JP 7 full clean environment: stopped before OOT modules because the local NVIDIA source bundle does not contain sources_7.0/hwpm
- Jetson JP 6.2 / Linux 5.15.148-tegra, D585 Prototype FW 7.58.46301.15388: PASS
  - V4L2 WHICH_DEF_VAL: [0, 0, 1, 0, 126, 1, 0]
  - idle out-of-range downscale: rejected with EINVAL and remained unchanged
  - Depth streaming: enable and threshold SETs succeeded; filter type and downscale changes returned EBUSY
  - IR-only streaming state: enable and threshold SETs succeeded; filter type and downscale changes returned EBUSY
  - depth capture: 300/300 frames at 30.00 FPS, stream exit 0

Jira: https://rsjira.realsenseai.com/browse/RSDEV-14893


kernel/realsense/d4xx.c
```diff
@@ -4494,6 +4475,21 @@ static int ds5_s_ctrl(struct v4l2_ctrl *ctrl)
...
+				goto unlock;
+			if (ctrl->p_new.p_u32[D500_MINZ_FILTER_TYPE_PARAM] !=
+			    le32_to_cpu(current_control.params[D500_MINZ_FILTER_TYPE_PARAM]) ||
+			    ctrl->p_new.p_u32[D500_MINZ_DOWNSCALE_PARAM] !=
+			    le32_to_cpu(current_control.params[D500_MINZ_DOWNSCALE_PARAM])) {
+				ret = -EBUSY;
+				goto unlock;
+			}
+		}
 		if (!d500_dpp_params_valid(dpp_desc, ctrl->p_new.p_u32)) {
```
ymodlin: **[rs-agentic bot — auto-generated] Potential BUG: MinZ out-of-range writes are now silently clamped instead of rejected.**

`type_ops->validate` runs in `validate_new()` before `s_ctrl`, and `d500_dpp_ctrl_validate_elem()` *clamps* to `[min,max]` and rounds to `step` rather than returning an error. With MinZ now supplying `.min/.max/.step`, this call is effectively unreachable — `p_new` is already in range by the time we get here.

So removing `d500_minz_params_valid()` changes the ABI: `threshold_mm = 70000` used to return `-EINVAL`; it now writes `65535` to FW and returns `0`. Same for `downscale = 0` (was `-EINVAL`, now `1`).

That may be the intent (it matches decimation/temporal), but it is not mentioned in the PR description or covered by the validation list. Please confirm, and if intended, drop the dead `d500_dpp_params_valid()` call or state the clamping contract.

kernel/realsense/d4xx.c
```diff
@@ -4494,6 +4475,21 @@ static int ds5_s_ctrl(struct v4l2_ctrl *ctrl)
...
 			goto unlock;
 		}
+		if (ctrl->id == D500_CAMERA_CID_MINZ && sensor &&
+		    sensor->streaming) {
+			struct d500_dpp_xu_control current_control;
+
+			ret = d500_dpp_xu_read(state, dpp_desc, &current_control);
+			if (ret)
+				goto unlock;
+			if (ctrl->p_new.p_u32[D500_MINZ_FILTER_TYPE_PARAM] !=
```
ymodlin: **[rs-agentic bot — auto-generated] Potential BUG: the idle-only guard can be bypassed by the clamp.**

`ctrl->p_new` has already been clamped by `d500_dpp_ctrl_validate_elem()` before `s_ctrl` runs, so the comparison is against the *clamped* value, not what userspace asked for.

Streaming, FW `downscale == 2`, userspace writes `7` → validate clamps `7 → 2` → compare is equal → no `EBUSY` → the write proceeds and `S_EXT_CTRLS` returns `0`. The requested idle-only change is neither applied nor reported. Same for `filter_type` at either range end.

The HW test in the description only covers an in-range downscale change, which does hit the `EBUSY` path. Making `validate` return `-ERANGE` for out-of-range slots instead of clamping would close both this and the issue flagged on the `d500_dpp_params_valid()` call.

kernel/realsense/d4xx.c
```diff
@@ -3350,6 +3350,8 @@ enum ds5_ae_type {
 #define D500_DPP_XU_DECIMATION_PARAM_COUNT	2
 #define D500_DPP_XU_TEMPORAL_PARAM_COUNT	4
 #define D500_DPP_XU_MINZ_PARAM_COUNT	7
+#define D500_MINZ_FILTER_TYPE_PARAM	1
```
ymodlin: **[rs-agentic bot — auto-generated] Enhancement:** these duplicate `enum d500_minz_param_index` — `D500_MINZ_FILTER_TYPE` (1) and `D500_MINZ_DOWNSCALE_RATIO` (2) — declared ~20 lines below and currently unused. Two index namespaces over the same wire layout will drift.

Reuse the enum and drop these defines. It also brings the comparison lines back under 80 cols (they are currently 95 and 93).

kernel/realsense/d4xx.c
```diff
@@ -3461,7 +3471,10 @@ static const struct d500_dpp_ctrl_desc d500_minz_desc = {
 	.control_id = D500_DPP_XU_MINZ_CONTROL_ID,
 	.param_count = D500_DPP_XU_MINZ_PARAM_COUNT,
 	.param_type = D500_DPP_XU_INTEGER_PARAMS,
-	.idle_only = true,
+	.min = d500_minz_min,
+	.max = d500_minz_max,
+	.step = d500_minz_step,
```
ymodlin, line 3474: **[rs-agentic bot — auto-generated]** Verified the new `.def` `{0,0,1,0,126,1,0}` satisfies `d500_minz_min`/`_max` on every slot (notably `downscale` min is `1`, so the old implicit all-zero default was invalid) — this is the actual fix. No issue here.

ymodlin (commented):
**Auto-generated review by the rs-agentic bot.** Not an approval.

The core idea is right: routing MinZ through `d500_dpp_ctrl_type_ops` is exactly what `CLAUDE.md` § "V4L2 control conventions" prescribes for a heterogeneous `V4L2_CTRL_TYPE_U32` array, and the reported `WHICH_DEF_VAL` `[0,0,1,0,126,1,0]` confirms `.init` is wired correctly. Two correctness issues fall out of the switch from hard rejection to `.validate` clamping — see inline.

Guideline notes (not blocking):
- `.claude/skills/add-xu-command` step 6 makes tests mandatory, and there is no `TestMinZ` class in `test/v4l2_test/tests/test_controls.py` today. The two behaviours this PR adds are exactly the ones that table covers: `test_minz_default` (per-slot defaults via `V4L2_CTRL_WHICH_DEF_VAL`) and a streaming case asserting `EBUSY` on filter-type/downscale vs. success on the runtime-updatable slots.
- `CLAUDE.md` § "Post-patch instruction hygiene": "idle-only enforced per-slot by diffing the new value against the live FW payload" is a new, reusable convention (the first `d500_dpp_ctrl_desc` that is neither fully `idle_only` nor fully runtime). Worth one line under § "V4L2 control conventions" so the next composite control does not re-derive it.
- PR is still in draft.

kernel/realsense/d4xx.c
lemon6630: Fixed in 358609e. MinZ now rejects an out-of-range element in type_ops validation before the shared clamp/round assignment, preserving the previous EINVAL ABI. Verified on D585: threshold_mm=70000 and downscale=7 both return EINVAL and leave the payload unchanged.

kernel/realsense/d4xx.c
lemon6630: Fixed in 358609e by the same MinZ-specific pre-clamp range check. An invalid request cannot become equal to the current value through clamping. Hardware validation covered out-of-range downscale=7 (EINVAL) and in-range streaming downscale=2 (EBUSY).

kernel/realsense/d4xx.c
lemon6630: Fixed in 358609e. The duplicate defines were removed and the existing d500_minz_param_index enum is used for both comparisons.

kernel/realsense/d4xx.c
lemon6630: Confirmed. The per-slot default remains [0, 0, 1, 0, 126, 1, 0] and was revalidated through V4L2_CTRL_WHICH_DEF_VAL on JP6.2 hardware.

kernel/realsense/d4xx.c
ymodlin: **[rs-agentic bot — auto-generated]** change approved

kernel/realsense/d4xx.c
ymodlin: **[rs-agentic bot — auto-generated]** change approved

kernel/realsense/d4xx.c
ymodlin: **[rs-agentic bot — auto-generated]** change approved

kernel/realsense/d4xx.c
ymodlin: **[rs-agentic bot — auto-generated]** Still holds on the new commit — no issue here.

kernel/realsense/d4xx.c
```diff
@@ -4494,6 +4476,35 @@ static int ds5_s_ctrl(struct v4l2_ctrl *ctrl)
...
+			bool depth_or_ir_streaming;
+			u32 current_filter_type;
+			u32 current_downscale_ratio;
+
+			mutex_lock(&state->ds5_dev->lock);
+			depth_or_ir_streaming =
+				state->ds5_dev->depth_streaming ||
+				state->ds5_dev->ir_streaming;
+			mutex_unlock(&state->ds5_dev->lock);
+			if (depth_or_ir_streaming) {
```
ymodlin: **[rs-agentic bot — auto-generated] Potential BUG: the streaming snapshot is released before it is used.**

`ds5_mux_s_stream()` never takes `state->lock`, so the flags can flip right after `mutex_unlock(&state->ds5_dev->lock)`. If depth starts in that window the idle-only `downscale`/`filter_type` write still lands mid-stream — exactly what the guard exists to prevent (reverse direction gives a spurious `EBUSY`).

Narrow, but the lock/unlock reads as if the guard were atomic and it is not. Either re-check the flags after `d500_dpp_xu_write()`, or note in a comment that the window is accepted.

Nit, same block: lines 4496/4498/4501 are 87/91/83 cols (checkpatch 80).

kernel/realsense/d4xx.c
```diff
@@ -3560,6 +3562,9 @@ static int d500_dpp_ctrl_validate_elem(const struct v4l2_ctrl *ctrl, u32 idx,
 		return -EINVAL;
 
 	value = clamp(ptr.p_u32[idx], desc->min[idx], desc->max[idx]);
+	if (ctrl->id == D500_CAMERA_CID_MINZ && value != ptr.p_u32[idx])
```
ymodlin, line 3565: **[rs-agentic bot — auto-generated] Enhancement:** this CID check (and the second one in `ds5_s_ctrl()`) special-cases a shared type-ops helper. A `bool strict_range` in `struct d500_dpp_ctrl_desc`, next to the existing `idle_only`, keeps the behaviour table-driven and lets the next control opt in without touching this function.

Also: V4L2 documents `-ERANGE` for an out-of-bounds control value; `-EINVAL` is accepted by the framework but `-ERANGE` is the spec errno for this case.

kernel/realsense/d4xx.c
lemon6630: Fixed in 3c0d93e. The MinZ path now holds ds5_dev->lock continuously from the Depth/IR state check through the live FW read, validation, and d500_dpp_xu_write(). ds5_mux_s_stream() must acquire the same lock before changing either shared streaming flag, so stream start and the control transaction now have a defined order. The flagged lines were also wrapped; the complete PR passes checkpatch with an 80-column limit. JP6.2/JP7 strict compiles and a 300-frame JP6.2 hardware regression passed.

kernel/realsense/d4xx.c
lemon6630: Keeping this MinZ-specific intentionally. MinZ is the only current descriptor that must reject rather than clamp, so strict_range would be a one-user field added for a hypothetical future control and would not remove the separate MinZ field-level streaming policy in ds5_s_ctrl(). Also keeping EINVAL: that is the pre-PR MinZ ABI from d500_minz_params_valid(), and the earlier correctness review explicitly required preserving it. Changing to ERANGE would be an unrelated ABI change. No code change for this enhancement.

kernel/realsense/d4xx.c
ymodlin: **[rs-agentic bot — auto-generated]** change approved

kernel/realsense/d4xx.c
```diff
@@ -4488,19 +4470,55 @@ static int ds5_s_ctrl(struct v4l2_ctrl *ctrl)
...
 			goto unlock;
 		}
-		if (!d500_dpp_params_valid(dpp_desc, ctrl->p_new.p_u32)) {
+		if (ctrl->id == D500_CAMERA_CID_MINZ) {
+			struct d500_dpp_xu_control current_control;
+			bool geometry_changed;
+			u32 current_filter_type;
+			u32 current_downscale_ratio;
+
+			mutex_lock(&state->ds5_dev->lock);
```
ymodlin, line 4490: **[rs-agentic bot — auto-generated] Guideline:** the race fix is correct, but `ds5_dev->lock` is now held across two retrying I2C transactions (`d500_dpp_xu_read()` + `d500_dpp_xu_write()`), which stalls any concurrent sibling `ds5_mux_s_stream()` for the duration.

That runs against `CLAUDE.md` *Concurrency notes*: "snapshot pointers/flags under lock and perform I2C probing after unlocking", and the same file's rule against holding the lock across a slow op so it cannot "needlessly stall a concurrent sibling" (cf. `ds5_flush_idle_link()`, which deliberately does its op unlocked).

Atomicity genuinely requires the lock here, so this reads as a deliberate deviation rather than a mistake. Please make it explicit: a 2–3 line comment on the `mutex_lock()` stating why the transaction must stay locked, plus a line in the `CLAUDE.md` Concurrency notes recording the new contract (`ds5_s_ctrl()` holds `ds5_dev->lock` across MinZ XU I2C) — `CLAUDE.md` asks that a changed locking contract be documented in the same change.

lemon6630: Addressed in d5d12bf: the lock site now explicitly documents that MinZ XU I/O is serialized against sibling stream changes. I kept CLAUDE.md unchanged because this is a narrow MinZ atomicity exception documented at its enforcement site; adding a repo-wide rule would expand this one-file functional fix without changing the contract elsewhere. JP6.2 and JP7 D4xx compilation with -Werror pass, and checkpatch reports 0 errors/0 warnings.

ymodlin: **[rs-agentic bot — auto-generated]** Comment part: change approved — the 3-line `why` at the `mutex_lock()` is concise and correct, and the new lines stay under 80 cols.

Still open, though: the `CLAUDE.md` half of the request was not done — this PR touches only `kernel/realsense/d4xx.c`. `ds5_s_ctrl()` now holds `ds5_dev->lock` across two retrying I2C transactions, which is a new locking contract and a deliberate deviation from the *Concurrency notes* rule "snapshot pointers/flags under lock and perform I2C probing after unlocking". `CLAUDE.md` *Post-patch instruction hygiene* requires the doc update in the same change. One line in *Concurrency notes* closes it, e.g.:

> `ds5_s_ctrl()` holds `ds5_dev->lock` across the MinZ XU read+write — the streaming-flag check and the wire transaction must be atomic against `ds5_mux_s_stream()`. This is the sanctioned exception to the snapshot-then-unlock rule.

lemon6630: Added the requested one-line MinZ exception to CLAUDE.md Concurrency notes in 56688b9. The PR remains one commit; no functional code changed after the already-passing JP6.2/JP7 compile checks.

ymodlin: **[rs-agentic bot — auto-generated]** change approved

kernel/realsense/d4xx.c
ymodlin: **[rs-agentic bot — auto-generated]** Acknowledged, agreed on both counts. `EINVAL` is the pre-PR MinZ ABI so keeping it is correct, and `strict_range` would be a single-user field today. Not blocking, no change needed.

ymodlin (approved):
LGTM

kernel/realsense/d4xx.c
```diff
@@ -4488,19 +4470,58 @@ static int ds5_s_ctrl(struct v4l2_ctrl *ctrl)
...
+			u32 current_filter_type;
+			u32 current_downscale_ratio;
+
+			/*
+			 * Serialize MinZ XU I/O against sibling stream changes.
+			 */
+			mutex_lock(&state->ds5_dev->lock);
+			minz_locked = true;
+			if (state->ds5_dev->depth_streaming ||
+			    state->ds5_dev->ir_streaming) {
```
ymodlin, line 4493: **[rs-agentic bot — auto-generated] Note (narrow, non-blocking): the guard can still let a geometry change through during stream *stop*.**

`ds5_mux_s_stream()` sets `*streaming_flag = on` under `ds5_dev->lock` *before* it issues the FW stop command and polls for completion (up to `DS5_START_MAX_TIME`). So for a stop, the flag reads `false` while the FW is still streaming. A MinZ `S_CTRL` landing in that window skips this check and writes `filter_type`/`downscale_ratio` to a non-idle FW.

Holding the lock does not close this — the flag itself is already optimistically false; the lock only orders the flag update, not the FW transition. (The start direction is the safe one: flag goes `true` early, so you get a conservative `EBUSY`.)

Same property affects the existing `d500_camera_is_idle_locked()` callers, so it is pre-existing and I would not block on it. Suggest a one-line comment here noting the flags track *intent*, not FW completion, so the next reader does not assume this check is exact.

ymodlin (commented):
**[rs-agentic bot — auto-generated] Re-review on request — no new commits since the last round.**

Head is still `56688b9`, the same commit I signed off on. All 7 earlier threads are resolved and the fixes hold up on a fresh read:

- `.min/.max/.step/.def` on `d500_minz_desc` + `.type_ops` on `d500_ctrl_minz` — `WHICH_DEF_VAL` now yields `[0,0,1,0,126,1,0]`, every slot inside `d500_minz_min/_max` (notably `downscale` min `1`). This is the actual fix.
- The MinZ pre-clamp reject in `d500_dpp_ctrl_validate_elem()` preserves the `-EINVAL` ABI and keeps the clamp from laundering an out-of-range value into a false "unchanged" compare in `ds5_s_ctrl()`.
- `ds5_dev->lock` held continuously across the flag check, `d500_dpp_xu_read()`, validation and `d500_dpp_xu_write()`. No lock-ordering inversion: no site in the file takes `ds5_dev->lock` before `state->lock`, and the I2C helpers take no locks. The pattern also matches `d500_set_device_mode()` / `d500_set_ae_policy()` / `d500_set_gyro_sensitivity()`, which already hold this lock across `ds5_raw_write()`.
- CLAUDE.md *Concurrency notes* records the exception.
- New/changed lines are all ≤80 cols; wire ABI stays 7 elements; decimation and temporal untouched.

One new **non-blocking** note inline about a stop-in-progress window, plus two observations that are pre-existing and out of scope here:

1. `ds5_mux_s_stream()` has an early `return ret;` on `-ENOSR` from `ds5_configure()` (~line 7632) that skips the `*streaming_flag = restore_val` restore, leaving `depth_streaming`/`ir_streaming` stuck `true`. It self-heals on the next stop, but this PR makes those flags load-bearing for MinZ, so a failed start now also means spurious `EBUSY` on geometry writes until then. Pre-existing — worth a separate ticket, not this PR.
2. Minor reuse: the `depth_streaming || ir_streaming` test is open-coded next to the existing `d500_camera_is_idle_locked()` helper. A sibling `_locked` helper would keep the convention. Cosmetic.

Not approving, per instruction.

